PR #6, opened 2018-07-04
Fix Paddle on centos running

sabreshao:


adityaatluri:
As user did not respond to code requests, pushed a commit myself.
commit: 
https://github.com/ROCmSoftwarePlatform/rccl/commit/624d84da2d2501a686708e00c5f10ef14ab4c05a

PR #21, opened 2018-09-25
Fixing the README.md and the compilation of the performance tests

abhinavvishnu:
Aditya/Jeff -- please review.


adityaatluri:
We don't want to remove usage of rccl op from README.md

abhinavvishnu:
Aditya: the usage is shown incorrectly. Either we fix it or remove it.

PR #38, opened 2018-10-10
Use one workgroup and improve performance by using 64b copying if

wenkaidu:
possible.

Data item size must be 64b aligned for this optimzation to work
in case of all gather. For example if data size is 20 bytes, copy
will be from offset 0 to offset 20. It is not possible to align
src and dst buffer to 64b.

### DIFF
--- a/src/rcclScalarAllGatherKernels.h
+++ b/src/rcclScalarAllGatherKernels.h
@@ -12,45 +12,111 @@ All rights reserved.
  * This file contains implementation of kernels used by rcclAllGather
  *
  */
+template<class T>
+__device__ bool is_aligned(const void * ptr) noexcept {
+    auto iptr = reinterpret_cast<std::uintptr_t>(ptr);
+    return !(iptr % alignof(T));
+}
 
 //! @brief Definition of RcclKernelScalarAllGather
 //! Gather data from all gpus and store to current gpu destination buffer
 template <typename DataType_t>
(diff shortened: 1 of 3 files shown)


src/rcclScalarAllGatherKernels.h
```diff
@@ -12,45 +12,111 @@ All rights reserved.
  * This file contains implementation of kernels used by rcclAllGather
  *
  */
+template<class T>
+__device__ bool is_aligned(const void * ptr) noexcept {
+    auto iptr = reinterpret_cast<std::uintptr_t>(ptr);
+    return !(iptr % alignof(T));
+}
```
adityaatluri: Add comments about functionality of the code.

src/rcclScalarAllGatherKernels.h
```diff
@@ -12,45 +12,111 @@ All rights reserved.
...
-                                          int count) {
-    int tx = threadIdx.x;
-    int bx = blockIdx.x;
-    int tid = tx + bx * knum_vectors_per_workgroup;
-
+        int count) {
+    int tid = threadIdx.x;
+    RingNode_t* pnext_track = pcurr_track->next_gpu;
+    bool use_64b_copy = 0;
+    int packed_count = count * sizeof(DataType_t) / sizeof(unsigned long long);
```
adityaatluri: You can add `VectorType_t` to templates and find the packed count as `count / (sizeof(VectorType_t) / sizeof(DataType_t))`

PR #39, opened 2018-10-15
Fixed arguments for reduce kernels to match with launch parameters

adityaatluri:


adityaatluri:
The diff is just cosmetic. It will get fixed after next commits clang-format run.

PR #40, opened 2018-10-17
Don't require doxygen if not building docs

acowley:


adityaatluri:
LGTM. Thank you for the contribution

PR #41, opened 2018-10-17
Make unit tests optional and runnable from top build directory

acowley:
- Lets one use `make test` in the top build directory to run the
unit tests.

- Makes it possible to use a system installation of googletest.


adityaatluri:
Hi,
I am planning to decouple tests cmake from build cmake. Let me know if you can update PR with that change, else I'll make the changes.

acowley:
I'm just looking at adapting this to your changes, and I have a question: would it be okay to rename the `inc` directory to `include` in this repository? The reason is that I'd like to run the tests before installing the packaging, so I'd like to set `RCCL_DIR` to the source directory so that the header is found. I can always symlink `include` to `inc` in my build script, but I'm wondering if you'd be open to just renaming the directory.

acowley:
I've updated the PR to work with current master.

acowley:
Hi @adityaatluri, do things look okay with the PR now?

adityaatluri:
> I'm just looking at adapting this to your changes, and I have a question: would it be okay to rename the `inc` directory to `include` in this repository? The reason is that I'd like to run the tests before installing the packaging, so I'd like to set `RCCL_DIR` to the source directory so that the header is found. I can always symlink `include` to `inc` in my build script, but I'm wondering if you'd be open to just renaming the directory.

Sure. Renaming works better. But, some of the code in `src` directory should be changed too.

PR #48, opened 2018-11-09
version information moved to rccl-version.h

jeffdaily:


jeffdaily:
This PR is no longer required; I was able to find an alternative for integration with tensorflow.

PR #51, opened 2018-11-14
add reduce scatter support

wenkaidu:
```bash
$ build/reduce_scatter_perf -b 32 -e 128M -f 2 -g 4
nThread 1 nGpus 4 minBytes 32 maxBytes 134217728 step: 2(factor) warmup iters: 5 iters: 20 validation: 1 
# RCCL Tests compiled with RCCL 1.0
# Using devices
#   Rank  0 on prj47-rack-05 device  0 [0x23] Device 6860
#   Rank  1 on prj47-rack-05 device  1 [0x26] Device 6860
#   Rank  2 on prj47-rack-05 device  2 [0x63] Device 6860
#   Rank  3 on prj47-rack-05 device  3 [0x66] Device 6860

#                                                 out-of-place                    in-place
#      bytes             N    type      op    time  algbw  busbw      res     time  algbw  busbw      res
          32             2   float     sum    0.079   0.00   0.00    1e-07    0.082   0.00   0.00    2e-07
          64             4   float     sum    0.152   0.00   0.00    2e-07    0.152   0.00   0.00    2e-07
         128             8   float     sum    0.151   0.00   0.00    2e-07    0.152   0.00   0.00    2e-07
         256            16   float     sum    0.153   0.00   0.00    2e-07    0.152   0.00   0.00    2e-07
         512            32   float     sum    0.168   0.00   0.00    5e-07    0.150   0.00   0.00    2e-07
        1024            64   float     sum    0.148   0.01   0.01    5e-07    0.150   0.01   0.01    2e-07
        2048           128   float     sum    0.150   0.01   0.01    5e-07    0.150   0.01   0.01    5e-07
        4096           256   float     sum    0.150   0.02   0.02    5e-07    0.150   0.02   0.02    5e-07
        8192           512   float     sum    0.149   0.04   0.04    5e-07    0.150   0.04   0.04    5e-07
       16384          1024   float     sum    0.149   0.08   0.08    5e-07    0.149   0.08   0.08    5e-07
       32768          2048   float     sum    0.151   0.16   0.16    5e-07    0.151   0.16   0.16    5e-07
       65536          4096   float     sum    0.150   0.33   0.33    5e-07    0.151   0.33   0.33    5e-07
      131072          8192   float     sum    0.150   0.65   0.65    5e-07    0.152   0.65   0.65    5e-07
      262144         16384   float     sum    0.152   1.29   1.29    5e-07    0.165   1.19   1.19    5e-07
      524288         32768   float     sum    0.196   2.00   2.00    5e-07    0.195   2.02   2.02    5e-07
     1048576         65536   float     sum    0.314   2.50   2.50    5e-07    0.315   2.49   2.49    5e-07
     2097152        131072   float     sum    0.538   2.92   2.92    5e-07    0.548   2.87   2.87    5e-07
     4194304        262144   float     sum    0.982   3.20   3.20    5e-07    0.981   3.21   3.21    5e-07
     8388608        524288   float     sum    1.929   3.26   3.26    5e-07    1.865   3.37   3.37    5e-07
    16777216       1048576   float     sum    3.623   3.47   3.47    5e-07    3.639   3.46   3.46    5e-07
    33554432       2097152   float     sum    8.897   2.83   2.83    5e-07    7.148   3.52   3.52    5e-07
    67108864       4194304   float     sum   14.176   3.55   3.55    5e-07   14.653   3.43   3.43    5e-07
   134217728       8388608   float     sum   29.826   3.38   3.38    5e-07   28.667   3.51   3.51    5e-07
 Out of bounds values : 0 OK
 Avg bus bandwidth    : 1.30644
```

### DIFF
--- a/inc/rccl.h
+++ b/inc/rccl.h
@@ -190,16 +190,16 @@ rcclResult_t rcclReduce(const void* sendbuff, void* recvbuff, int count,
 rcclResult_t rcclAllGather(const void* sendbuff, int count, rcclDataType_t datatype,
                             void* recvbuff, rcclComm_t comm, hipStream_t stream);
 
-//! Does reduction op on sendbuff on all gpus and scatter the result to all gpus
-//! on recvbuff Reduction op (rcclRedOp_t) is done on data (of data type
-//! rcclDataType_t) in sendbuff of length = count on all gpus and stored in
-//! recvbuff on all gpus. The operation is launched on the stream provided. To
-//! do in-place reduce scatter, pass in same pointers to recvbuff and sendbuff (
-//! in-place all-reduce, sendbuff = recvbuff)
+//! Does reduction op on sendbuff on all gpus and scatter the result to all gpus'
+//! recvbuff. Reduction op (rcclRedOp_t) is done on data (of data type
+//! rcclDataType_t) in sendbuff of length = recvcount*num_gpus on all gpus and
+//! stored in recvbuff with length = recvcount on all gpus. The operation is
+//! launched on the stream provided. To do in-place reduce scatter, pass in
+//! same pointers to recvbuff and sendbuff (sendbuff = recvbuff)
 
 //! \param [in] sendbuff Source buffer
 //! \param [in] recvbuff Destination buffer
-//! \param [in] count Number of elements in buffer
+//! \param [in] recvcount Number of elements in receive buffer
 //! \param [in] datatype Data type of buffers
 //! \param [in] op Reduction operation on buffers
 //! \param [in] comm Communicator for current gpu
--- a/inc/rccl.h
+++ b/inc/rccl.h
@@ -190,6 +190,23 @@ rcclResult_t rcclReduce(const void* sendbuff, void* recvbuff, int count,
 rcclResult_t rcclAllGather(const void* sendbuff, int count, rcclDataType_t datatype,
                             void* recvbuff, rcclComm_t comm, hipStream_t stream);
 
+//! Does reduction op on sendbuff on all gpus and scatter the result to all gpus
+//! on recvbuff Reduction op (rcclRedOp_t) is done on data (of data type
+//! rcclDataType_t) in sendbuff of length = count on all gpus and stored in
+//! recvbuff on all gpus. The operation is launched on the stream provided. To
+//! do in-place reduce scatter, pass in same pointers to recvbuff and sendbuff (
+//! in-place all-reduce, sendbuff = recvbuff)
+
+//! \param [in] sendbuff Source buffer
+//! \param [in] recvbuff Destination buffer
+//! \param [in] count Number of elements in buffer
+//! \param [in] datatype Data type of buffers
+//! \param [in] op Reduction operation on buffers
+//! \param [in] comm Communicator for current gpu
+//! \param [in] stream HIP stream the op launches on
+rcclResult_t rcclReduceScatter(const void* sendbuff, void* recvbuff, int recvcount,
+                           rcclDataType_t datatype, rcclRedOp_t op,
+                           rcclComm_t comm, hipStream_t stream);
 #ifdef __cplusplus
 }  // end extern "C"
 #endif
--- a/src/rcclReduceScatter.cpp
+++ b/src/rcclReduceScatter.cpp
@@ -1,0 +1,313 @@
+/*
+Copyright (c) 2017 - Present Advanced Micro Devices, Inc.
+All rights reserved.
+*/
+
+/**
+ * @file rcclReduceScatter.cpp
+ * @brief rccl library implementation of rcclReduceScatter API
+ *
+ * This file contains implementation of rcclReduceScatter API.
+ *
+ */
+
+#include "rcclDataTypes.h"
+#include "rcclHelper.h"
+#include "rcclSetKernels.h"
+#include "rcclTracker.h"
+
+#include "rcclScalarReduceScatterRuntime.h"
+
+#include <string>
+#include <unordered_map>
+
+extern std::unordered_map<int, std::string> umap_red_op;
+extern std::unordered_map<int, std::string> umap_datatype;
+
+extern int RCCL_TRACE_RT;
+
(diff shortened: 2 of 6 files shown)


jedwards-AMD:
LGTM

inc/rccl.h
```diff
@@ -190,6 +190,23 @@ rcclResult_t rcclReduce(const void* sendbuff, void* recvbuff, int count,
 rcclResult_t rcclAllGather(const void* sendbuff, int count, rcclDataType_t datatype,
                             void* recvbuff, rcclComm_t comm, hipStream_t stream);
 
+//! Does reduction op on sendbuff on all gpus and scatter the result to all gpus
```
gilbertlee-amd: scatters

src/rcclReduceScatter.cpp
```diff
@@ -0,0 +1,313 @@
+/*
+Copyright (c) 2017 - Present Advanced Micro Devices, Inc.
```
gilbertlee-amd: 2018

src/rcclReduceScatter.cpp
```diff
@@ -0,0 +1,313 @@
...
+                stream, API_COLOR_END);
+    }
+
+    //! Check if buffer pointers are not null
+    if (sendbuff == nullptr || recvbuff == nullptr) {
+        return rcclInvalidDevicePointer;
+    }
+
+    //! Check if data type of buffers is valid or not
+    if (datatype >= rccl_NUM_TYPES) {
```
gilbertlee-amd: This check could fail if the passed in datatype was a negative number.   Do you need/want to check for that case too?  This assumes that the enumeration starts at 0.

src/rcclReduceScatter.cpp
```diff
@@ -0,0 +1,313 @@
...
+        return rcclInvalidType;
+    }
+
+    //! Check if op is valid or not
+    if (op >= rccl_NUM_OPS) {
```
gilbertlee-amd: Same issue as above.

jeffdaily: It might be worth filing a separate issue to fix all these checks.  I'm assuming this code was copy-and-pasted and modified based on the other implemented functions so these checks would appear elsewhere, as well.

src/rcclReduceScatter.cpp
```diff
@@ -0,0 +1,313 @@
...
+    //! If same comm is used on a different stream, synchronize it with current
+    //! stream before launching op.
+    PreEnqueueEventRecord(pcomm, stream);
+
+    //! Get tracker to current gpu
+    RingNode_t *pcurr_track = pcomm->track_;
+
+    //! If the number of gpus equal to 1, do a simple memory copy
+    if (num_gpus == 1) {
+        switch (datatype) {
```
gilbertlee-amd: There's quite a bit of duplicated code here.  I'd suggest having a inline function that returns the size of each of the rcclDataType_t, and then just having a single hipMemCpyAsync.  The out-of-bounds datatype check has already occurred so you can skip the default case.

src/rcclReduceScatter.cpp
```diff
@@ -0,0 +1,313 @@
...
+        default: { return rcclInvalidType; }
+        }
+
+        //! Track current stream so that op launched on different stream can be
+        //! synchronized with current stream
+        PostEnqueueEventRecord(pcomm, stream);
+        return rcclSuccess;
+    }
+
+    //! Check which op to launch
```
gilbertlee-amd: There's quite a bit of code here required to expand out the 2 templated arguments (datatype / operation).
An alternative solution for dealing with this is to "de-onion" and deal with one template argument at a time:

For example:
```
foo_datatype(rcclDataType_t datatype, rcclRedOp_t op, ...)
{
   switch (datatype)
   {
     case rcclChar: return foo_op<signed char, rccl_char16_t>(op, COMMON_ARGS);
     case rcclHalf: return foo_op<__fp16, rccl_half8_t>(op, COMMON_ARGS);
     ...
}

template <typename T, rcclDataType_t datatype>
foo_op(rcclRedOp_t op, COMMON_ARGS)
{
  switch (op)
  {
  case rcclSum: return RcclInternalReduceScatter<T, datatype, rcclSum>(COMMON_ARGS);
  case rcclProd: return RcclInternalReduceScatter<T, datatype, rcclProd>(COMMON_ARGS);
  ...
}
```

This might provide more clarity to the code, and avoid any nasty one-off typos.  Also, it might make life easier if we have to add any additional template parameters.

jeffdaily:
@gilbertlee-amd in general I agree with your comments.  However, I think the comments should be applied more generally throughout the entire code base.  I think this PR should focus on the new function and an issue be filed concerning a code review like yours.  That way we can capture the concerns you raised here but fix everything at once.

gilbertlee-amd:
Sounds good - I'll focus on reviewing the kernel.

inc/rccl.h
wenkaidu: will fix this

src/rcclReduceScatter.cpp
wenkaidu: this will be a valid checking. But it needs to be applied to the entire code. Let's take on this as separate task.

src/rcclReduceScatter.cpp
wenkaidu: agreed. let's fix this globally in another patch.

src/rcclReduceScatter.cpp
wenkaidu: this is the same copy/paste consistent with all other collectives. If we want to fix it, let's fix them all.

src/rcclReduceScatter.cpp
wenkaidu: Are there any performance or resource usage benefits of making these changes?

gilbertlee-amd: Unlikely.  Purely cosmetic.  Don't worry about it.

inc/rccl.h
wenkaidu: Please review function description patch

PR #55, opened 2018-11-28
Fix #54. Replace std::atomic with HIP atomic.

### LINKED ISSUE #54: Removing use of std::atomic
It seems rccl is currently using std::atomic. Although HCC is able to compile the code, deep inside, there are some issues for supporting std::atomic with HCC. For details, please see discussions about supporting std::atomic in HIP for hip-clang. https://github.com/ROCm-Developer-Tools/HIP/pull/764

As such, I suggest rccl to use hip atomics (https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/hcc_detail/hip_atomic.h) instead of std::atomics.

jeffdaily:
HIP atomics do not provide load and store operations. They are instead emulated using atomicAdd and atomicExch, respectively.

### DIFF
--- a/src/rcclTracker.cpp
+++ b/src/rcclTracker.cpp
@@ -25,12 +25,9 @@ RingNodePool_t::RingNodePool_t() {
     HIPCHECK(
         hipHostMalloc(&barrier_, sizeof(Barrier_t), hipHostMallocCoherent));
 
-    std::atomic_store_explicit(&(barrier_->bar_in), 0,
-                               std::memory_order_seq_cst);
-    std::atomic_store_explicit(&(barrier_->bar_out), 0,
-                               std::memory_order_seq_cst);
-    std::atomic_store_explicit(&(barrier_->times_done), 0,
-                               std::memory_order_seq_cst);
+    barrier_->bar_in = 0;
+    barrier_->bar_out = 0;
+    barrier_->times_done = 0;
 }
 
 //! @brief Default destructor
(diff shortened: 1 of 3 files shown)


jeffdaily:
If you wish to evaluate the changes, you can use the docker image `jeffdaily/rocm:1.9.2-rccl-hip-atomics`.

src/rcclTracker.cpp
```diff
@@ -25,12 +25,9 @@ RingNodePool_t::RingNodePool_t() {
     HIPCHECK(
         hipHostMalloc(&barrier_, sizeof(Barrier_t), hipHostMallocCoherent));
 
-    std::atomic_store_explicit(&(barrier_->bar_in), 0,
-                               std::memory_order_seq_cst);
-    std::atomic_store_explicit(&(barrier_->bar_out), 0,
-                               std::memory_order_seq_cst);
-    std::atomic_store_explicit(&(barrier_->times_done), 0,
-                               std::memory_order_seq_cst);
+    barrier_->bar_in = 0;
```
wenkaidu: should these use atomicExch as well?

jeffdaily: The HIP atomics use __device__ only, no __host__, so I can't use atomicExch here.  This is initializing the barrier values to 0 with hipHostMallocCoherent memory, and this is during rcclComm_t initialization.  Do we need to use an atomic store in this case?

jeffdaily: Outside of the host initializing these variables to 0, they are never again accessed by the host.  However, each GPU must be able to access these values.  I'm unsure, but I thought for atomics to work across GPUs the memory must be host mapped?

wenkaidu: Can we clarify across GPU atomic memory location with design team? At minimal, can we ensure consistent access to these variables by removing any access from host side, and only let GPU to set them through atomic methods?

PR #58, opened 2019-02-09
Add reinterpret_cast to call sites of hipHostMalloc

whchung:
To cope with potential upcoming changes in HIP runtime API where `hipMalloc` and `hipHostMalloc` becomes pure C functions, `reinterpret_cast<void**>()` must be introduced to properly match HIP interface.


whchung:
After discussion with stakeholders in HIP, it has been decided to tweak HIP headers. Therefore changes proposed in this PR is no longer needed.

PR #70, opened 2019-05-23
Add RCCL primitive testing

wenkaidu:


gilbertlee-amd:
Is there any way for us to include the "current" collective device code in use within the RCCL library instead of copy_kernel.h?

wenkaidu:
The key function is copy exact from src/collectives/device/common_kernel.h:
__device__ inline void ReduceCopy128b( const int w, const int nw, const int t,
    Pack128 * src0, Pack128 * src1, Pack128 * dest0, Pack128 * dest1,
    const int N) {

gilbertlee-amd:
> The key function is copy exact from src/collectives/device/common_kernel.h:
> device inline void ReduceCopy128b( const int w, const int nw, const int t,
> Pack128 * src0, Pack128 * src1, Pack128 * dest0, Pack128 * dest1,
> const int N) {

Yes - does this mean we can just include it, instead of making a copy?  That way if we change things, we don't need to synchronize with this one?

wenkaidu:
Both approach has plus and minus. Current approach is a simple portable test utilities can be easily used anywhere. While common_kernel.h includes core.h and introduces too many dependencies. I would prefer keep it simple for now.

PR #72, opened 2019-05-24
Increase number of rings with XGMI connection

wenkaidu:
Improve throughput for about 20%. Also remove P2P over PCIe which was
left enabled at initial release.

Signed-off-by: Wenkai Du <wenkai.du@amd.com>


jeffdaily:
What, if any, performance impact will this change make when `link_type != HSA_AMD_LINK_INFO_TYPE_XGMI`?

wenkaidu:
Performance will be worse on machines which have PCIe P2P.

PR #75, opened 2019-05-28
Implement HDP flush when transfer data over PCIe P2P

wenkaidu:


wenkaidu:
Please review updated patch which is more streamlined than previous version.

src/collectives/device/primitives.h
```diff
@@ -44,11 +44,13 @@ class PostFlag {
...
   const int fifo_size;
+  uint32_t * hdp_reg;
  public:
   __device__
-  PostFlag(volatile uint64_t* const flag, const int shift, volatile int* const fifo, const int fifo_size) : flag(flag), shift(shift), fifo(fifo), fifo_size(fifo_size) { }
+  PostFlag(volatile uint64_t* const flag, const int shift, volatile int* const fifo, const int fifo_size, uint32_t* hdp_reg = NULL)
+    : flag(flag), shift(shift), fifo(fifo), fifo_size(fifo_size), hdp_reg(hdp_reg) { }
   __device__
-  void post(uint64_t val) { STORE(flag, (val - shift)); }
+  void post(uint64_t val) { if (hdp_reg != NULL) STORE(hdp_reg, 0x1); STORE(flag, (val - shift)); }
```
gilbertlee-amd: This is rather mysterious.  Can you provide a comment which explains what this ends up doing?

src/include/core.h
```diff
@@ -153,6 +153,9 @@ struct ncclRing {
       int* userRanks;
       int* devUserRanks;
 
+      // Next GPU's HDP register address
+      uint32_t* hdp_reg;
```
gilbertlee-amd: Can we briefly explain here what the HDP register is used for?

src/collectives/device/primitives.h
wenkaidu: Done. The descriptions were given by Sean from his earlier email.

src/include/core.h
wenkaidu: Done

wenkaidu:
I fixed a few other typos. I think "acked" is correct. Maybe Sean can comment?

wenkaidu:
Added suggested texts

src/transport/p2p.cu
```diff
@@ -487,10 +482,62 @@ end:
...
+      }
+    }
+    return HSA_STATUS_SUCCESS;
+  };
+
+  char busId[NVML_DEVICE_PCI_BUS_ID_BUFFER_SIZE];
+  *hdp = NULL;
+  CUDACHECK(hipDeviceGetPCIBusId(busId, NVML_DEVICE_PCI_BUS_ID_BUFFER_SIZE, cudaDev));
+  hsa_agent_t agent;
+  agent.handle = convert_bdf(busId);
```
skeelyamd: Seems a bit confusing at first glance.  convert_bdf produces a uint16_t.  agent.handle should be an opaque pointer.  I see that you're using it to pass in search parameters for find_agent.  Would a union be clearer?

src/transport/p2p.cu
```diff
@@ -487,10 +482,62 @@ end:
...
+  hsa_iterate_agents(find_agent, (void*)&agent);
+  hsa_amd_hdp_flush_t hdpinfo;
+  hsa_status_t err = hsa_agent_get_info(agent, (hsa_agent_info_t)HSA_AMD_AGENT_INFO_HDP_FLUSH, &hdpinfo);
```
skeelyamd: If agent lookup on 517 fails to locate the device then this call will almost certainly segfault since agent.handle will have non-null garbage in it.  Adding a check (or at least assert) will help debugging if something does go wrong in bdf id comparison.

src/transport/p2p.cu
```diff
@@ -487,10 +482,62 @@ end:
...
 ncclResult_t p2pSendSetup(ncclTinfo_t* myOpaqueInfo, ncclTinfo_t* peerOpaqueInfo, struct ncclConnect* connectInfo, struct ncclRing* ring) {
   struct p2pInfo* myInfo = (struct p2pInfo*)myOpaqueInfo;
   struct p2pInfo* peerInfo = (struct p2pInfo*)peerOpaqueInfo;
   struct p2pConnectInfo info;
+  uint32_t linktype, hops;
+  if (hipExtGetLinkTypeAndHopCount(myInfo->cudaDev, peerInfo->cudaDev, &linktype, &hops) != hipSuccess) {
+    INFO(NCCL_INIT|NCCL_P2P,"Ring %02d : %d -> %d failed to get link type and hop count", ring->id, myInfo->rank, peerInfo->rank);
+    return ncclInternalError;
+  }
+  if (linktype != HSA_AMD_LINK_INFO_TYPE_XGMI) {
```
skeelyamd: Shouldn't hops be checked first?  If the link is indirect (more than 1 hop) then it's possible (on future systems) that the link type isn't unique and that an intermediate physical link may traverse PCIe.

Maybe the check for direct connection has already been done before this function is called?

src/collectives/device/primitives.h
```diff
@@ -44,11 +44,26 @@ class PostFlag {
...
  public:
   __device__
-  PostFlag(volatile uint64_t* const flag, const int shift, volatile int* const fifo, const int fifo_size) : flag(flag), shift(shift), fifo(fifo), fifo_size(fifo_size) { }
+  PostFlag(volatile uint64_t* const flag, const int shift, volatile int* const fifo, const int fifo_size, uint32_t* hdp_reg = NULL)
+    : flag(flag), shift(shift), fifo(fifo), fifo_size(fifo_size), hdp_reg(hdp_reg) { }
+  // Data transfers through PCIe are not guaranteed to occur in order to due to how writes
+  // may be combined when received by the remote GPU. This can cause an issue if the write
+  // to the flag signaling transfer completion gets reordered and occurs prior to the actual
+  // data transfer.
+  // HDP register is used to suppress PCIe re-ordering. The reordering we need to suppress is
```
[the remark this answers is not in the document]
wenkaidu: done

src/collectives/device/primitives.h
```diff
@@ -44,11 +44,26 @@ class PostFlag {
...
+  // between the HDP write combiner (where PCIe writes are acked) and local memory (where it
```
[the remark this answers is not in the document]
wenkaidu: done

src/collectives/device/primitives.h
```diff
@@ -44,11 +44,26 @@ class PostFlag {
...
+  // becomes visible to the GPU). Any PCIe reordering is handled by the s_waitcnt preceding
```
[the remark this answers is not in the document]
wenkaidu: done

src/transport/p2p.cu
wenkaidu: done

src/transport/p2p.cu
wenkaidu: done

src/transport/p2p.cu
wenkaidu: P2P connection has already been decided by calling link type and hop count in another path. Since there is no easy way to pass along the information, it is easier just to call the API again only to check if the link is XGMI.

src/transport/p2p.cu
```diff
@@ -493,17 +493,21 @@ static ncclResult_t getGpuHdpReg(int cudaDev, uint32_t** hdp) {
...
+    uint16_t id = ((union find_agent_args *)arg)->id;
     hsa_device_type_t type;
     hsa_agent_get_info(agent, HSA_AGENT_INFO_DEVICE, (void*)&type);
     if(type == HSA_DEVICE_TYPE_GPU) {
       uint16_t bdf_id = 1;
       hsa_agent_get_info(agent, (hsa_agent_info_t)HSA_AMD_AGENT_INFO_BDFID, &bdf_id);
       if(bdf_id == id) {
-        *found_agent=agent;
-        return HSA_STATUS_ERROR;
+        ((union find_agent_args *)arg)->agent=agent;
```
skeelyamd: Use of union here is inconsistent with other C style typecasts.  Should this be (find_agent_args*)arg to match the other casts or is the project style okay with both uses?

Also feel free to ignore this since it's entirely a style thing.

skeelyamd: Nevermind.  Code already has a mix.

PR #76, opened 2019-05-29
Make ncclFuncs static

yxsamliu:
This is necessary to constant propagate the function pointers
to eliminate the indirect function call.


wenkaidu:
Tested OK

PR #88, opened 2019-06-24
Fixing install script to actually install library when requested

stanleytsang-amd:
Also cleaned up unused command-line arguments and weird tabs in the install script too.


stanleytsang-amd:
> If I do "./install.sh -i", I don't see this file: /opt/rocm/lib/librccl.so
> If I manually build and install deb package, I got that file.
> ls -l /opt/rocm/lib/librccl.so
> lrwxrwxrwx 1 root root 22 Jun 24 16:42 /opt/rocm/lib/librccl.so -> ../rccl/lib/librccl.so
> 
> Also this file is missing:
> /opt/rocm/include/rccl.h

It looks like the install script was previously set up incorrectly - the install_prefix needs to be /opt/rocm. not /opt/rocm/rccl.  I've made the change and now it looks like librccl.so and rccl.h are installed properly as per your comment.  Let me know if it works for you now.

Also updated the README to reflect the changes, plus some missing information.

PR #89, opened 2019-06-25
Use hipLaunchCooperativeKernelMultiDevice API

wenkaidu:
Depends on HIP version with this pull request:
https://github.com/ROCm-Developer-Tools/HIP/pull/1165


wenkaidu:
> We should merge this into devel instead of master if it depends on specific version of HIP

At this time, this is for review only. This cannot be merged until the HIP change is on roc-master.

PR #94, opened 2019-07-04
Refactor primitive test to support multiple GPUs in rings

wenkaidu:


tools/rccl-prim-test/rccl_prim_test.cpp
```diff
@@ -122,21 +142,83 @@ do {                                                                           \
...
+  for (int i = 0; i < deviceCnt; i++)
+    printf("%1d ", ring[i]);
+  printf("\n");
+}
+
+static void findConnect(uint32_t *info, int *ring, int deviceCnt) {
+  int n = 0, curr = 0, best;
+  uint32_t temp[MAX_GPU*MAX_GPU];
+  for (int i = 0; i < deviceCnt*deviceCnt; i++) temp[i] = 0;
+  for (int i = 0; i < deviceCnt; i++) {
```
gilbertlee-amd: Why do we loop over i here?  It's not used.

wenkaidu: This line?
for (int i = 0; i < deviceCnt*deviceCnt; i++) temp[i] = 0;
It is used to initialize the temp matrix to all 0s.

gilbertlee-amd: Nevermind - That's the ring size

tools/rccl-prim-test/rccl_prim_test.cpp
```diff
@@ -122,21 +142,83 @@ do {                                                                           \
...
+        HIPCHECK(hipExtGetLinkTypeAndHopCount(i, j, &linktype, &info[i*deviceCnt+j]));
+      }
+      else
+        info[i*deviceCnt+j] = 0;
+    }
+  }
+  HIPCHECK(hipSetDevice(dev));
+}
+
+static void printRing(int id, int *ring, int deviceCnt) {
```
gilbertlee-amd: Can we pass in info so that we can also see the hop distance in between?

wenkaidu: You can always refer to distance which are already printed in connection matrix. Ring is to show the constructed GPU order.
Connection matrix:
 0  1  1  2 
 1  0  2  1 
 1  2  0  1 
 2  1  1  0 
Ring 0: 0 1 3 2 
Ring 1: 0 2 3 1

tools/rccl-prim-test/rccl_prim_test.cpp
```diff
@@ -122,21 +142,83 @@ do {                                                                           \
...
+        best = j;
+        hops = info[curr*deviceCnt+j];
+      }
     }
-     HIPCHECK(hipSetDevice(dev));
+    curr = best;
+  }
+}
+
+static int findNextGpu(int *ring, int gpu, int deviceCnt) {
```
gilbertlee-amd: It might not affect performance much, but it looks like this is being called inside nested for loops.
We can easily just create an array of whatever length and then fill this once and then just access nextGpu[myId] directly instead of having to search.

tools/rccl-prim-test/rccl_prim_test.cpp
```diff
@@ -188,194 +276,164 @@ int main(int argc,char* argv[])
...
-  HIPCHECK(hipMemcpyAsync(transfer_data_1, &h_transfer_data_1,
-                          sizeof(struct transfer_data_t), hipMemcpyHostToDevice,
-                          stream_1));
-  HIPCHECK(hipStreamSynchronize(stream_1));
+  float *buff[MAX_GPU*MAX_WORKGROUPS], *buff_coarse[MAX_GPU*MAX_WORKGROUPS];
+  struct transfer_data_t h_transfer_data[MAX_GPU], *transfer_data[MAX_GPU];
+  struct profiling_data_t *profiling_data[MAX_GPU], *d_profiling_data[MAX_GPU];
+  hipStream_t stream[MAX_GPU];
+
+  uint64_t N = 2097152;
```
gilbertlee-amd: Let's make this accessible via a cmd line parameter as well.

wenkaidu: done

PR #95, opened 2019-07-09
Adding explicit HDP flush when using RDMA via Infiniband

gilbertlee-amd:
- Performs an explicit HDP flush for GPU-based RDMA memory receives after IB verbs reports completion / local read-back finishes.
- Modifies pre-existing HDP register variable to distinguish between "next" GPU's flush register and "this" GPU's flush register (Although only 1 is ever used - next for P2P transfers, curr for IB RDMA transfers)


wenkaidu:
LGTM. There is also this commit pending review for some time:
https://github.com/ROCm-Developer-Tools/HIP/pull/1192
We can merge this for now. But we should move to HIP API later.

PR #100, opened 2019-07-17
Use hipExtLaunchMultiKernelMultiDevice API

wenkaidu:
Depends on HIP version with this pull request:
https://github.com/ROCm-Developer-Tools/HIP/pull/1232


wenkaidu:
Cannot merge until https://github.com/ROCm-Developer-Tools/HIP/pull/1232 is on amd-master

gilbertlee-amd:
We should merge this into devel instead of master

PR #102, opened 2019-07-22
Versioning updated to new rocm-cmake style and 2.7 release

saadrahim:


wenkaidu:
Is it possible to add commit hash?

saadrahim:
commit hash is standard now. It might be added in too many places as it stands. It is being added to the library name as well.

saadrahim:
You might not see commit hash until rocm 2.7 is released depending on the build environment. remove the rocm-cmake package to get commit hash.

PR #105, opened 2019-07-25
Remove extra "." from version string

wenkaidu:


wenkaidu:
Need this patch for 2.7 branch as well

PR #106, opened 2019-07-30
Add gfx908 target

wenkaidu:


saadrahim:
We  should merge to develop first.

PR #107, opened 2019-07-31
Fix build with hip-clang

mhbliao:
Two minor issues are solved:
+ Enclose the kernel function with parenthesis as hip-clang defines
  `hipLaunchKernelGGL` as macro.
+ Need to explicitly include <hsa.h> for hip-clang.


mhbliao:
This is the change necessary to compile RCCL with hip-clang.
BTW, do we still have time to cherry-pick this into 2.7 branch

yxsamliu:
These changes are necessary for hip-clang and are compatible with HCC.

PR #110, opened 2019-08-01
Revise the previous fix to use the canonical path to HSA.

mhbliao:
- This fix the build failures under certain environments.


mhbliao:
Sorry, have to revise the previous fix as failures are observed under certain environment.

PR #112, opened 2019-08-05
Get HDP register address from hipDeviceGetAttribute API

wenkaidu:


jeffdaily:
In which ROCm release does this HIP attribute become available?

wenkaidu:
> In which ROCm release does this HIP attribute become available?

The HIP patch is on amd-master now. So it should be on 2.8. Target same release for RCCL.

PR #116, opened 2019-08-12
Removing unnecessary device collective source files.

stanleytsang-amd:
All the preprocessor directive paths were being taken anyway in our current CMake structure.  Slightly faster compilation time (~3%), slightly smaller librccl.so size.  Much less clutter.


wenkaidu:
Instead of adding _dev.cpp, can we just use the .cu file, i.e. reduce.cu?

PR #119, opened 2019-08-14
RCCL 2.4 update

wenkaidu:


src/collectives/device/reduce_scatter.cu
```diff
@@ -1,6 +1,7 @@
 /*************************************************************************
  * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
  * Modifications Copyright (c) 2019 Advanced Micro Devices, Inc. All rights reserved.
+ * Modifications Copyright (c) 2019 Advanced Micro Devices, Inc. All rights reserved.
```
stanleytsang-amd: I must have accidentally added an extra AMD copyright statement here.  Can you remove the extra line? Sorry.

wenkaidu: Done

PR #122, opened 2019-08-15
Tune LL threshold for VEGA

wenkaidu:
Also move back abort check to after SPINS_BEFORE_CHECK_ABORT


src/include/enqueue.h
```diff
@@ -15,6 +15,7 @@
 #define NCCL_LL_CHANNEL_THRESHOLD 8 // Per thread size before we start increasing nrings
 #define NCCL_THREAD_THRESHOLD 256  // Per thread size before we switch to non-LL
 #define NCCL_THREAD_THRESHOLD_PREVOLTA 32 // Per thread size before we switch to non-LL for pre-Volta archs
+#define NCCL_THREAD_THRESHOLD_VEGA 8 // Per thread size before we switch to non-LL for VEGA
```
gilbertlee-amd: Can you provide some reasoning behind this number?  Essentially, doesn't this just lower the limit at which we would use LL?

wenkaidu: This probably should have been changed earlier. We increase number of LL threads from 64 to 256, thus each thread get less data. 32/4 = 8.

PR #129, opened 2019-08-29
display each workgroup ,links and directions with throughputs

rpathani:


wenkaidu:
Can you please refactor this pull request, so that it will not include any merge commit?

rpathani:
> Can you please refactor this pull request, so that it will not include any merge commit?
Try to do squashing to merged all commit into one but seems it was not working..so forced pushed into it..

PR #133, opened 2019-10-03
detect the hcc version and conditionally add -hc-function-calls

scchan:


scchan:
Not for that particular PSDB but we'll get a newer hcc with a version bump to 2.10 to address that

PR #146, opened 2019-10-30
check for fine grain support at top of p2pCanConnect

jeffdaily:
This fixes a TensorFlow unit test where the HSA_FORCE_FINE_GRAIN_PCIE env var is not set, HIP_VISIBLE_DEVICES=0 is set, and two ranks are created on the same device.  The p2p transport should always be skipped if fine grain memory is not requested.

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -171,11 +171,14 @@ void parseHsaForceFineGrainVramPcie() {
     errno = 0;
     int64_t v = strtoll(str, NULL, 0);
     if (errno || (v != 0 && v != 1)) {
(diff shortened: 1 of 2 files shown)


wenkaidu:
HSA_FORCE_FINE_GRAIN_PCIE is not needed when XGMI is available. So this case needs to be taken care of.

jeffdaily:
The issue is this check here:
https://github.com/ROCmSoftwarePlatform/rccl/blob/648c1ee7cc3ca6755bb817e6c2c834707aaeabf0/src/transport/p2p.cc#L113-L117
The case we need to resolve is when the same device is passed to `p2pCanConnect` for both devices.  Does XGMI matter in this case?  Will XGMI be used for P2P from a device to itself?  If not, then the check I put in should simply move down into the code block that checks for same device ID.

PR #154, opened 2019-11-18
Add bfloat16 support in RCCL

wenkaidu:
RCCL API version is updated to 2.5.0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,11 @@
 
 cmake_minimum_required(VERSION 2.8.12)
 
+# We use C++14 features, this will add compile option: -std=c++14
+set( CMAKE_CXX_STANDARD 14 )
+# Without this line, it will add -std=gnu++14 instead, which has some issues.
+set( CMAKE_CXX_EXTENSIONS OFF )
+
 set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "")
 
 project(rccl CXX)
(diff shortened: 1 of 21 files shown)


makefiles/version.mk
```diff
@@ -1,6 +1,6 @@
 ##### version
 NCCL_MAJOR   := 2
-NCCL_MINOR   := 4
-NCCL_PATCH   := 8
+NCCL_MINOR   := 5
+NCCL_PATCH   := 0
```
iotamudelta: We don't expect NCCL to have a 2.5? That may end up confusing if they release 2.5 w/o bf16 support?

jeffdaily: If bf16 support is going to be added as an extra ROCm feature to an otherwise compatible NCCL 2.4.8 API, I second this concern.  Can we instead add a preprocessor symbol to the rccl.h header indicating bf16 support?  Then applications could check for this symbol to determine if the feature is available, rather than checking an artificial version number that now diverges from upstream.

wenkaidu: Removed changes to API version.
Added #define RCCL_BFLOAT16 1 in rccl.h instead

wenkaidu:
Hi all, based on feedback I reverted API version change. Use RCCL_BFLOAT16 ==1 instead.
Please help review again.

wenkaidu:
All uni test passed:
[==========] 1842 tests from 9 test cases ran. (2097039 ms total)
[  PASSED  ] 1842 tests.

jithunnair-amd:
Hi Wenkai, I tested these changes with pytorch and they work seamlessly! Can you please make sure these changes make it into ROCm 3.0 so that we can enable bfloat16 in pytorch?
cc @iotamudelta

PR #164, opened 2020-01-02
Changes for multiple ROCm installation

paulfreddy:
1. Set soversion to match release version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,18 +173,8 @@ else()
   target_link_libraries(rccl PUBLIC hip::hip_hcc ${hcc_LIBRARIES} numa)
 endif()
 
-## Set the VERSION and SOVERSION values
-set (LIB_VERSION_MAJOR 1 )
-set (LIB_VERSION_MINOR 0 )
-if ( ${ROCM_PATCH_VERSION} )
-    set (LIB_VERSION_PATCH ${ROCM_PATCH_VERSION} )
-else()
-    set (LIB_VERSION_PATCH 0 )
-endif()
-
-set ( LIB_VERSION_STRING "${LIB_VERSION_MAJOR}.${LIB_VERSION_MINOR}.${LIB_VERSION_PATCH}" )
-set_property ( TARGET rccl PROPERTY VERSION "${LIB_VERSION_STRING}" )
-set_property ( TARGET rccl PROPERTY SOVERSION "${LIB_VERSION_MAJOR}" )
+#Setup librccl.so version
+rocm_set_soversion(rccl "2.7")
 
 ## Linker Flags
 ## Add --enable-new-dtags to generate DT_RUNPATH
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -170,6 +173,26 @@ else()
   target_link_libraries(rccl PUBLIC hip::hip_hcc ${hcc_LIBRARIES} numa)
 endif()
 
+## Set the VERSION and SOVERSION values
```
saadrahim: from rocBLAS: rocm_set_soversion( rocblas ${rocblas_SOVERSION} )

Something like that will do this section more efficiently I believe.

saadrahim: https://github.com/ROCmSoftwarePlatform/rocBLAS/search?q=rocblas_SOVERSION&unscoped_q=rocblas_SOVERSION

paulfreddy: Done.

install.sh
```diff
@@ -80,7 +80,7 @@ while true; do
     esac
     done
 
-rocm_path=/opt/rocm/bin
+rocm_path=$install_prefix/bin
```
gilbertlee-amd: This isn't quite correct.  If the user decides to install to a new location (other than where a ROCm installation exists), this will fail because no compiler will be there.

CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ set(AMDGPU_TARGETS gfx900;gfx906;gfx908 CACHE STRING "List of specific machine t
 find_package(ROCM
              REQUIRED
              PATHS
-             /opt/rocm)
+             /opt/rocm $CMAKE_INSTALL_PREFIX)
```
gilbertlee-amd: I'm not clear on what this is supposed to do.  Could you explain the reasoning behind this?

paulfreddy: Agreed this can be removed since CMAKE already put INSTALL_PREFIX as a search path.

CMakeLists.txt
```diff
@@ -170,6 +173,16 @@ else()
   target_link_libraries(rccl PUBLIC hip::hip_hcc ${hcc_LIBRARIES} numa)
 endif()
 
+#Setup librccl.so version
+rocm_set_soversion(rccl "2.7")
```
saadrahim: SO version for RCCL should be 1.1 I believe.

install.sh
paulfreddy: Is this install.sh used in the current form?

CMakeLists.txt
paulfreddy: You are right this library version was 1.0, this need to be 1.0

install.sh
gilbertlee-amd: Yes

paulfreddy:
@saadrahim  @gilbertlee-amd ,

Any more comments on this PR. Hope this is OK to be merged.

install.sh
gilbertlee-amd: I fail to see how this has been resolved.

saadrahim: I think it has, that change is removed. Only one line I see

paulfreddy: Yes, After looking at it closely, that change was not relevant for this PR and removed it.
Will submit another PR with a proper fix if that is needed.

gilbertlee-amd: Ah - my mistake.  I clicked on view changes and saw the outdated changes.

PR #165, opened 2020-01-06
Do not use struct ncclColl in kernel launch parameter

wenkaidu:
This helps in reducing register spills


wenkaidu:
> Do you have any idea how this affects latency for smaller array sizes?
Overall differences are below run to run variations. If looking really carefully by running test 3 times each and check for minimal numbers, the numbers are 31.53 vs 32.23

PR #173, opened 2020-02-21
Fix incorrect CR8 detection

wenkaidu:
Also change level of ring graph print to help debugging


wenkaidu:
This is an urgent fix. The logic should be ||, not &&.

PR #185, opened 2020-03-16
rccl-prim-test: add all-to-all benchmark

wenkaidu:
For gfx908, support simple detection of ring topology.
Call ReduceOrCopyMulti directly from kernel.
Also simplify code by removing kernel start synchronization option
which has no effect on throughput measurements.


wenkaidu:
> all2all operation will run all the operation or is it suppose to do something else?

All2all testing does P2P read from all remote GPUs directly and sum all the data. Results are write back to all GPUs.

PR #196, opened 2020-04-27
Adding Full Logging to CI

saadrahim:


wenkaidu:
I think we should only add NCCL_DEBUG=INFO.

PR #199, opened 2020-04-29
Enable parallel jobs for hip-clang build

wenkaidu:


CMakeLists.txt
```diff
@@ -164,6 +164,12 @@ if("${HIP_COMPILER}" MATCHES "clang")
   endforeach()
   target_link_libraries(rccl PRIVATE -fgpu-rdc)
   target_include_directories(rccl PRIVATE /opt/rocm/hsa/include)
+  find_program( hipcc_executable hipcc )
+  execute_process(COMMAND bash "-c" "${hipcc_executable} -help | grep 'parallel-jobs'" OUTPUT_VARIABLE hipcc_parallel_jobs)
+  if("${hipcc_parallel_jobs}" MATCHES "parallel-jobs")
+    target_compile_options(rccl PRIVATE -parallel-jobs=4 PRIVATE -Wno-format-nonliteral)
```
saadrahim, line 170: one question: why did the compiler team choose 4?

wenkaidu: @yxsamliu Will we get shorter build time with more than 4 parallel jobs? Must compiling and linking options use same number of jobs?

wenkaidu: Tested using 8 parallel jobs. It was not faster. So merging the patch as is.

yxsamliu: it will launch a parallel job for each gpu arch. If you have 4 gpu archs and specify 8 it will only launch 4 parallel jobs but if you specify 3 it will only launch 3 parallel jobs.

PR #215, opened 2020-05-22
Defaulting to hipcc in install.sh script

saadrahim:
Hot fix for release branch. This needs to be merged back to develop once rocm 3.5 is live.


wenkaidu:
cmake_common_options is not taken. Can you also fix this as well?

PR #216, opened 2020-06-04
Making hip-clang the default compiler; documentation update

stanleytsang-amd:
For now, I've provided users the option to build with hcc if they still wish by replacing the --hip-clang option with --hcc.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -13,7 +13,7 @@ function display_help()
     echo "    [-p|--package_build] Build RCCL package."
     echo "    [-t|--tests_build] Build unit tests, but do not run."
     echo "    [-r|--run_tests] Run unit tests (must be built already.)"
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -13,7 +13,7 @@ function display_help()
     echo "    [-p|--package_build] Build RCCL package."
     echo "    [-t|--tests_build] Build unit tests, but do not run."
     echo "    [-r|--run_tests] Run unit tests (must be built already.)"
-    echo "    [--hip-clang] Build library using hip-clang compiler."
```
saadrahim, line 16: Can we keep a silent --hip-clang option? Devops passes this to the CI and we don't want to break builds. Once all libraries update and devops fixes their CI, we can remove this.

stanleytsang-amd: Will do.

PR #217, opened 2020-06-05
Fixing CI as install.sh script should not install dependencies withou…

saadrahim:
…t user request


saadrahim:
This is needed to fix CI for RCCL.

PR #220, opened 2020-06-17
ib-test: support host memory allocation through posix_memalign

wenkaidu:


tools/ib-test/ib_test.cpp
```diff
@@ -333,16 +353,30 @@ int main(int argc,char* argv[])
...
+    if (use_gdr_read) {
+      NCCLCHECK(ncclCudaCalloc(&sendDevBuffer, sendBuffSize, 1));
+      printf("Allocated sendDevBuffer %p of %d bytes, sliceSteps %d\n",
+                sendDevBuffer, sendBuffSize, sliceSteps);
+    }
+    else {
+#ifdef USE_MEMALIGN
+      int page_size = getpagesize();
+      posix_memalign((void **)&sendHostBuffer, page_size, sendBuffSize);
+      hipHostRegister(sendHostBuffer, sendBuffSize, hipHostRegisterMapped);
```
stanleytsang-amd, line 369: Is there any particular reason why we don't have CUDACHECK() around any the hipHostRegister and hipHostGetDevicePointer calls?

wenkaidu: good catch! fixed.

PR #221, opened 2020-06-18
Use posix_memalign for network buffer allocation on host memory

wenkaidu:


src/transport/net.cc
```diff
@@ -93,10 +96,21 @@ ncclResult_t netSendSetup(struct ncclTopoSystem* topo, struct ncclTopoGraph* gra
   }
   char line[16];
   if (resources->buffSizes[LOC_HOSTMEM]) {
+#ifdef USE_MEMALIGN
+    int page_size = getpagesize();
+    posix_memalign((void **)&resources->hostBuffer, page_size, resources->buffSizes[LOC_HOSTMEM]);
```
souravzzz, line 101: We should probably check the retval from posix_memalign since it can fail if OOM

wenkaidu: Done

PR #229, opened 2020-07-07
Match RCCL package name to API version

wenkaidu:


saadrahim:
Will version number of the package drop? That may impact upgrading.

wenkaidu:
> Will version number of the package drop? That may impact upgrading.

I don't think so. If version drops, a lot of code will break as they check for API version greater than certain numbers.

PR #234, opened 2020-07-15
Changing GTest inclusion in cmake to use find_package

saadrahim:
Will add error message asking user to install cmake.


saadrahim:
This fixes all our client packaging issues we were having.

PR #237, opened 2020-07-20
Download GTest if not found in system

saadrahim:


stanleytsang-amd:
I just tried this PR on a machine that doesn't have googltest installed (ixt-rack-87) and it doesn't seem to be able to build the unit tests.  While it detects the absence of googletest, it's unable to find where the gtest headers are installed:

/home/statsang/WORKSPACE/clean/rccl/test/CorrectnessTest.hpp:12:10: fatal error: 'gtest/gtest.h' file not found
#include <gtest/gtest.h>

PR #242, opened 2020-07-27
Topology tuning for 4P2H on Rome

wenkaidu:


src/graph/search.cc
```diff
@@ -771,6 +771,137 @@ static ncclResult_t parseChordalRing(struct ncclTopoSystem* system, char **str)
   return ncclSuccess;
 }
 
+static bool getGpuNetCount(struct ncclTopoSystem* system, int id, int *ngpu, int *nnet) {
+  *ngpu = 0; *nnet = 0;
+  int i;
+  for (i = 0; i < system->nodes[CPU].count; i++)
+    if (system->nodes[CPU].nodes[i].id == id) break;
+  if (i >= system->nodes[CPU].count)
+    return false;
```
gilbertlee-amd, line 780: I'd recommend using ncclTopoIdToIndex(system, CPU, id, index) instead

src/graph/search.cc
```diff
@@ -771,6 +771,137 @@ static ncclResult_t parseChordalRing(struct ncclTopoSystem* system, char **str)
...
+  int n, m;
+  int ngpus = system->nodes[GPU].count;
+  *gpu1 = -1; *gpu2 = -1;
+  int c1, c2;
+  for (c1 = 0; c1 < system->nodes[CPU].count; c1++)
+    if (system->nodes[CPU].nodes[c1].id == cpu1) break;
+  for (c2 = 0; c2 < system->nodes[CPU].count; c2++)
+    if (system->nodes[CPU].nodes[c2].id == cpu2) break;
+  if (c1 >= system->nodes[CPU].count || c2 >= system->nodes[CPU].count)
+    return false;
```
gilbertlee-amd, line 798: Another place to use ncclTopoIdToIndex

src/graph/search.cc
wenkaidu: Done

src/graph/search.cc
wenkaidu: Done

PR #244, opened 2020-07-30
Adding static library building option.

stanleytsang-amd:
Users can now choose to build RCCL as a static library by passing in -s or --static to the install script.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -9,7 +9,7 @@ def runCompileCommand(platform, project, jobName)
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}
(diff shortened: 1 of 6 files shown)


stanleytsang-amd:
> I will add CI coverage to your fork. That we can test prior to merging.

Sounds good.

saadrahim:
Static library tests do not work? Is there a point to merging this until things are fixed?

stanleytsang-amd:
> Static library tests do not work? Is there a point to merging this until things are fixed?

They don't build because the hipcc perl script has a bug that introduces extra escape characters into the linker, causing it to fail to build unit tests.  As far as RCCL is concerned, the updates in the PR is fine, and I'm assuming we need to have these changes in develop in time for 3.8?

PR #251, opened 2020-08-12
Cleaning up CI code be removing overrides

saadrahim:


saadrahim:
This is just CI code. We are refactoring things so that we do not have to add commit as much but a few more commits needed today in separate PRs.

PR #256, opened 2020-08-18
Add NPS4 support on some models

wenkaidu:


wenkaidu:
> This detection code is getting quite complicated.
> Would it be possible to pull it out into a different file for easier maintenance / merging?

Yes, I am thinking about a plug-in to make it cleaner and less likely to break existing models

PR #258, opened 2020-08-24
Only use software barrier for synchronization

wenkaidu:


src/collectives/device/primitives.h
```diff
@@ -84,9 +84,7 @@ class ncclPrimitives {
 
   inline __device__ void barrier() {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-    if (nthreads == gridDim.x) {
-      __syncthreads();
-    } else if (wid == 0) {
+    if (wid == 0) {
```
souravzzz, line 87: What's the impact of this PR? Does it improve performance?

wenkaidu: No, there should be no change in performance. All collectives using same barrier mechanism will improve code coverage and help to eliminate potential bugs.

PR #260, opened 2020-08-26
Add RCCL Static Lib Creation with -fgpu-rdc

aaronenyeshi:
RCCL uses -fgpu-rdc to compile its source objects. When linking
the RCCL static library, the link and archive step must do through
hipcc and uses the flag --emit-static-lib. When compiling
UnitTests, the librccl.a must be consumed through -l and -L.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -25,7 +25,17 @@ if(BUILD_TESTS)
 
   add_executable(UnitTests ${TEST_SOURCES})
   target_include_directories(UnitTests PRIVATE /opt/rocm ${GTEST_INCLUDE_DIRS})
-  target_link_libraries(UnitTests PRIVATE ${GTEST_BOTH_LIBRARIES}  PRIVATE rccl)
+  target_link_libraries(UnitTests PRIVATE ${GTEST_BOTH_LIBRARIES})
+
+  # UnitTests using static library of rccl requires passing rccl
+  # through -l and -L instead of command line input.
+  if(BUILD_STATIC)
+    add_dependencies(UnitTests rccl)
+    target_link_libraries(UnitTests PRIVATE dl rt numa -lrccl -L${CMAKE_BINARY_DIR})
+    target_link_libraries(UnitTests PRIVATE amdhip64 amd_comgr hsa-runtime64::hsa-runtime64)
+  else()
+    target_link_libraries(UnitTests PRIVATE rccl)
+  endif()
 else()
   message("Not building unit tests")
 endif()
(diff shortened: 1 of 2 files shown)


aaronenyeshi:
Tested that this patch creates `librccl.a` and builds `UnitTests` successfully if you use command: `./install.sh -ts`. UnitTests' 2 GPU tests are passing for me, but I didn't run them for too long. Tested on build 3565-STG2 docker image.

saadrahim:
Do static builds work in the standard mainline docker containers or do we need a separate static container?

saadrahim:
I want to update our CI's static tests so that this PR is correctly tested in the math library CI system.

aaronenyeshi:
> Do static builds work in the standard mainline docker containers or do we need a separate static container?

This patch was tested in the standard mainline docker container 3565. I'm not sure about the separate static containers.

test/CMakeLists.txt
```diff
@@ -25,7 +25,16 @@ if(BUILD_TESTS)
...
   add_executable(UnitTests ${TEST_SOURCES})
   target_include_directories(UnitTests PRIVATE /opt/rocm ${GTEST_INCLUDE_DIRS})
-  target_link_libraries(UnitTests PRIVATE ${GTEST_BOTH_LIBRARIES}  PRIVATE rccl)
+  target_link_libraries(UnitTests PRIVATE ${GTEST_BOTH_LIBRARIES})
+
+  # UnitTests using static library of rccl requires passing rccl
+  # through -l and -L instead of command line input.
+  if(BUILD_STATIC)
+    add_dependencies(UnitTests rccl)
+    target_link_libraries(UnitTests PRIVATE dl rt numa -lrccl -L${CMAKE_BINARY_DIR})
```
yxsamliu, line 34: Does rccl itself depend on these libraries? If so, shouldn't they be specified through target_link_libraries(rccl INTERFACE ...) ?

aaronenyeshi: Right, RCCL depends on these libraries, but normally it should have been brought in when we set link to hip::device (which includes them from either hsa-runtime::hsa-runtime64 or other deps). But I couldn't figure out how to force CMake to resolve something like hip::device into its list of link deps. 

Also, we may need to support librccl.a as an input file instead of using -L and -l?

aaronenyeshi: It looks like I need to explicitly put the static libraries into the PRIVATE libraries for UnitTests. I'm not sure how to work around it, but it doesn't seem like its picking up all the dependencies in static mode. I checked hip::device, and the dependency on hsa-runtime64::hsa-runtime64 is a PRIVATE one.

aaronenyeshi:
> Do static builds work in the standard mainline docker containers or do we need a separate static container?

@saadrahim - this new patch I pushed was tested in the static container. And I ran some of the 2GPU unit tests on my machine. I've changed some of the syntax to support cmake 3.10 which is used in the static images.

yxsamliu (approved):

stanleytsang-amd (approved):

PR #265, opened 2020-09-04
Adding the ability to force install dependencies (namely gtest); gtest library installation fix for centos

stanleytsang-amd:
A googletest dependency check in find_package(GTest 1.10) will always pass if there is gtest already installed in the typical system directories (ie. /usr/local/lib etc.) regardless of whether or not it is 1.10 installed, which is required for our unit tests.  This is because gtest does not supply the proper cmake package config files to indicate the version.

I've re-enabled the -d/--dependency option in the install script which will force install googletest 1.10 into the build folder.   This should be used concurrently with -t to ensure the unit test build correctly picks up the gtest library dependency.   Also, I added a check to determine where the gtest library is installed, as I encountered an issue in our centos docker image where it was installed in build/release/gtest/lib64/ (vs. build/release/gtest/lib in ubuntu).

Also updated documentation to indicate new flag and also updated unit test notes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ project(rccl CXX)
 set(AMDGPU_TARGETS gfx803;gfx900;gfx906;gfx908 CACHE STRING "List of specific machine types for library to target")
 
 option(BUILD_TESTS "Build test programs" OFF)
+option(INSTALL_DEPENDENCIES "Force install dependencies" OFF)
 
 include(cmake/Dependencies.cmake)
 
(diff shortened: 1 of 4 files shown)


wenkaidu:
"./install.sh -d" needs sudo. Can this be avoided?

stanleytsang-amd:
> "./install.sh -d" needs sudo. Can this be avoided?

You know what, we actually might be able to avoid sudo...if we don't need chrpath and openmp any more as dependencies, then we don't need to install those packages.  chrpath was only required for unit tests but we don't use it any more, it seems like?  I think the same for openmp as well.   We don't require either of those to build the RCCL library itself, right?

wenkaidu:
I don't remember the history of chrpath. But even it is needed, I think it would be better to print out a message, abort the script and let user install it manually with sudo.

wenkaidu (approved):

PR #278, opened 2020-10-13
Removing unnecessary flags from CI

gilbertlee-amd:


saadrahim (approved):

.jenkins/common.groovy
```diff
@@ -22,7 +21,7 @@ def runTestCommand (platform, project)
     def command = """#!/usr/bin/env bash
                 set -x
                 cd ${project.paths.project_build_prefix}/build/release/test
-                NCCL_DEBUG=INFO HSA_FORCE_FINE_GRAIN_PCIE=1 ./UnitTests --gtest_output=xml --gtest_color=yes
+                NCCL_DEBUG=INFO ./UnitTests --gtest_output=xml --gtest_color=yes
```
wenkaidu, line 24: We should keep HSA_FORCE_FINE_GRAIN_PCIE=1 flag. It will make test faster on non-XGMI machines.

PR #285, opened 2020-10-19
Fix incorrect pointer checking for scatter and gather

wenkaidu:


gilbertlee-amd (approved):

sarunyap:
@wenkaidu Will this check the recv buffer for Scatter and send buffer for Gather on all ranks?

wenkaidu:
> @wenkaidu Will this check the recv buffer for Scatter and send buffer for Gather on all ranks?
It should only skip send buffer checking for non-root broadcast and scatter, receiver buffer checking for non-root reduce and gather. So the conditions in your list will not be skipped.

sarunyap (approved):

PR #286, opened 2020-10-20
Fixing typo.  Pinned host memory uses C not P

gilbertlee-amd:


stanleytsang-amd (approved):

wenkaidu (approved):
It would be better if we can include component's name in title if it is not RCCL library. I have been adding like topo_expl and rccl-prim-test etc in title for this purpose.

PR #298, opened 2020-11-10
Use target id of xnack off

wenkaidu:


gilbertlee-amd (approved):

saadrahim (approved):
This is what the other libraries are doing.

PR #301, opened 2020-11-12
Skip unused peer connection in scatter and gather

wenkaidu:


sarunyap (approved):
I tested this patch and the problem that a program gets stuck when running Allgather and Scatter repeatedly is gone.

PR #303, opened 2020-11-13
gtest: add scatter to combined calls and use loops

wenkaidu:


sarunyap (approved):
LGTM

gilbertlee-amd (changes requested):
We may want to validate results after each iteration, to avoid the case where results are wrong on any of the iterations aside from the last one.  For example, if the first iteration were wrong for some reason.

wenkaidu:
> We may want to validate results after each iteration, to avoid the case where results are wrong on any of the iterations aside from the last one. For example, if the first iteration were wrong for some reason.

done!

gilbertlee-amd (approved):

PR #310, opened 2020-12-15
Fix Rome PCIe 2 node topology generation

wenkaidu:


src/graph/search.cc
```diff
@@ -426,7 +426,7 @@ ncclResult_t ncclTopoSearchRecNet(struct ncclTopoSystem* system, struct ncclTopo
         for (int i = 0; i<system->nodes[GPU].count; i++)
           if (paths[i].count < paths[f].count) f = i;
         int t = 1 << 10;
-        NCCLCHECK(ncclTopoSearchTryGpu(system, graph, saveGraph, 0, backToNet, backToFirstRank, FORCED_ORDER_PCI, &t, NET, n, f));
+        NCCLCHECK(ncclTopoSearchTryGpu(system, graph, saveGraph, 0, backToNet, backToFirstRank, (f == 0) ? FORCED_ORDER_PCI : 0, &t, NET, n, f));
```
gilbertlee-amd, line 429: Can you explain this change?
- If GPU 0 is closest to the NIC, then do FORCED_ORDER_PCI, otherwise perform normal search?

wenkaidu: yes. FORCED_ORDER_PCI means the order will be 0, 1, 2 .. 7. When search started with 4, it becomes 4, 1, 2, 3, 4, 5, 6, 7, which is incorrect. A few months earlier, I had a fix for this, but it caused other issues which has to be reverted. We can discuss offline on this.

gilbertlee-amd (approved):

PR #317, opened 2021-02-11
Update install.sh

pramenku:
Install.sh having hard code like /opt/rocm/bin/hipcc for rocm_path and default_path=/opt/rocm
This will work only when we have standalone rocm installed. If anyone has installed, side-by-side, they will face below error.

Can we keep like ROCM_PATH=$ROCM_PATH  instead of default_path as variable name and 
ROCM_BIN_PATH=$ROCM_PATH/bin ,rocm_path can be replaced with ROCM_BIN_PATH.

This way, we will have option to export ROCM_PATH as env variable as per need and use the script. 
I have also tried locally, it is working.  ROCM_PATH is common variable name, we are having.

Error when side-by-side install is done for driver.
# ./install.sh -dtr 2>&1 | tee /dockerx/6519_rccl-test.log
CMake Error at /usr/share/cmake/Modules/CMakeDetermineCXXCompiler.cmake:48 (message):
Could not find compiler set in environment variable CXX:
/opt/rocm/bin/hipcc.
Call Stack (most recent call first):
CMakeLists.txt:12 (project)

CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
See also "/root/driver/rccl/build/release/CMakeFiles/CMakeOutput.log".

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -23,7 +23,6 @@ function display_help()
 # #################################################
 # global variables
 # #################################################
-ROCM_PATH=$ROCM_PATH
 build_package=false
 install_prefix=$ROCM_PATH
 build_tests=false
--- a/install.sh
+++ b/install.sh
@@ -23,9 +23,9 @@ function display_help()
 # #################################################
 # global variables
 # #################################################
-default_path=/opt/rocm
+ROCM_PATH=$ROCM_PATH
 build_package=false
-install_prefix=$default_path
+install_prefix=$ROCM_PATH
 build_tests=false
 run_tests=false
 run_tests_all=false
@@ -103,7 +103,7 @@ while true; do
     esac
     done
 
-rocm_path=/opt/rocm/bin
+ROCM_BIN_PATH=$ROCM_PATH/bin
 
 # /etc/*-release files describe the system
 if [[ -e "/etc/os-release" ]]; then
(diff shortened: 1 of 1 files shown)


wenkaidu:
Since the script now requires setting ROCM_PATH, can you also add this into README.md?
If I do "git log", there are some strange characters in 2 lines:
"Can we keep like ROCM_PATH=$ROCM_PATH  instead of \ufffd<80><9C>default_path\ufffd<80><9D> as variable name "
"I have also tried locally, it\ufffd<80><99>s working"

install.sh
```diff
@@ -23,9 +23,9 @@ function display_help()
 # #################################################
 # global variables
 # #################################################
-default_path=/opt/rocm
+ROCM_PATH=$ROCM_PATH
```
wenkaidu, line 26: ROCM_PATH?=/opt/rocm
This can keep CI machine happy

install.sh
```diff
@@ -23,9 +23,9 @@ function display_help()
...
 build_package=false
-install_prefix=$default_path
+install_prefix=$ROCM_PATH
```
wenkaidu, line 28: install_prefix=${ROCM_PATH:="/opt/rocm"}

install.sh
wenkaidu, line 26: delete this?

pramenku: Thanks, Done

pramenku (commented):
> 
> 
> Since the script now requires setting ROCM_PATH, can you also add this into README.md?
> If I do "git log", there are some strange characters in 2 lines:
> "Can we keep like ROCM_PATH=$ROCM_PATH instead of \ufffd<80><9C>default_path\ufffd<80><9D> as variable name "
> "I have also tried locally, it\ufffd<80><99>s working"

Removed the special character

wenkaidu (approved):

install.sh
pramenku: Thanks. Just renamed  "install_prefix" to "ROCM_PATH". It's working well now.

pramenku:
Please also cherry pick the changes in "master" branch. QA used master branch in Mainline. Thanks.

PR #326, opened 2021-03-06
Add GPU memory usage tracker

wenkaidu:


src/include/alloc.h
```diff
@@ -37,13 +37,28 @@ static ncclResult_t ncclCalloc(T** ptr, size_t nelem) {
   return ncclSuccess;
 }
 
+struct __attribute__ ((aligned(64))) allocationTracker {
+  uint64_t totalAlloc;
```
gilbertlee-amd, line 41: Can we just use a union with one field being 64 chars?

src/include/alloc.h
```diff
@@ -37,13 +37,28 @@ static ncclResult_t ncclCalloc(T** ptr, size_t nelem) {
...
 static ncclResult_t ncclCudaCalloc(T** ptr, size_t nelem, bool isFineGrain = false) {
   if (isFineGrain)
     CUDACHECK(hipExtMallocWithFlags((void**)ptr, nelem*sizeof(T), hipDeviceMallocFinegrained));
   else
     CUDACHECK(hipMalloc(ptr, nelem*sizeof(T)));
   CUDACHECK(hipMemset(*ptr, 0, nelem*sizeof(T)));
+  int dev;
+  CUDACHECK(hipGetDevice(&dev));
+  if (dev < MAX_ALLOC_TRACK_NGPU) {
+    __atomic_fetch_add(&allocTracker[dev].totalAlloc, 1, __ATOMIC_SEQ_CST);
```
gilbertlee-amd, line 59: The number of allocs is being counted, but I don't think it's being reported anywhere.

src/init.cc
```diff
@@ -58,6 +58,8 @@ NCCL_PARAM(CheckPointers, "CHECK_POINTERS", 0);
 ncclNet_t* ncclNet = NULL;
 ncclCollNet_t* ncclCollNet = NULL;
 
+struct allocationTracker allocTracker[MAX_ALLOC_TRACK_NGPU];
```
gilbertlee-amd, line 61: Might need to initialize this with = {};

src/include/alloc.h
```diff
@@ -37,13 +37,28 @@ static ncclResult_t ncclCalloc(T** ptr, size_t nelem) {
...
+    __atomic_fetch_add(&allocTracker[dev].totalAllocSize, nelem*sizeof(T), __ATOMIC_SEQ_CST);
```
gilbertlee-amd, line 60: These values aren't reset when comm gets destroyed - is that intentional?

src/include/alloc.h
wenkaidu: done

src/include/alloc.h
wenkaidu: The information is not useful outside RCCL. Thus I decided not to print it.

src/include/alloc.h
wenkaidu: I missed it. I added reset to init, as sometimes framework do not call comm destroy

src/init.cc
wenkaidu: done

src/include/alloc.h
gilbertlee-amd: But why even track then?  How do you check that value?

gilbertlee-amd (approved):

src/include/alloc.h
wenkaidu: We can check with gdb. Anyway, original thought was to use counter to track hipMalloc/hipFree pair so we can detect leak. But this requires create a wrapper and change all location of hipFree. We can discuss offline to see if this is worth doing.

PR #359, opened 2021-04-28
Fixing potential race-condition in env var parameter macro

gilbertlee-amd:
Fixing a possible race condition when utilizing un-cached mode for unit testing with RCCL_TEST_ENV_VARS=ENABLE.

The potential race condition is as follows:
- Thread 1 gets lock and sets static "value" variable to non-default value and releases lock
- Thread 2 gets lock and initializes static "value" variable to default value
- Thread 1 returns static "value" variable set to default value (not the non-default value)

With this fix, a "localValue" modified only while locked echoes the value of "value" and is returned outside of the lock instead of "value".

### DIFF
--- a/src/include/param.h
+++ b/src/include/param.h
@@ -83,9 +83,11 @@ pthread_mutex_t rcclParamMutex##name = PTHREAD_MUTEX_INITIALIZER; \
 int64_t rcclParam##name() { \
   static_assert(default_value != -1LL, "default value cannot be -1"); \
   static int64_t value = -1LL; \
+  int64_t localValue; \
   pthread_mutex_lock(&rcclParamMutex##name); \
+  localValue = value; \
   char* en = getenv("RCCL_TEST_ENV_VARS"); \
-  if (value == -1LL || (en && (strcmp(en, "ENABLE") == 0))) { \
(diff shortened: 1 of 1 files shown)


stanleytsang-amd:
Fix looks good to me! But won't we need to update NCCL_PARAM as well?
https://github.com/ROCmSoftwarePlatform/rccl/blob/47bbad04e5ff41b3c1457a24134204a65427a460/src/include/param.h#L57

wenkaidu (approved):

gilbertlee-amd:
NCCL_PARAM doesn't requires a change because "value" will only ever change once.
Once the first thread sets value != -1LL, it never ever gets changed again (skips the if statement).  This is done safely while holding the mutex, so it should be ok.  Could you confirm this analysis?

stanleytsang-amd:
> NCCL_PARAM doesn't requires a change because "value" will only ever change once.
> Once the first thread sets value != -1LL, it never ever gets changed again (skips the if statement). This is done safely while holding the mutex, so it should be ok. Could you confirm this analysis?

Ah, I missed that.  You're right.  All is good.

stanleytsang-amd (approved):

PR #366, opened 2021-05-12
Sync up to NCCL 2.9.6

wenkaidu:


wenkaidu:
I mistakenly squashed commits when merging. I have to revert and do it again correctly

PR #443, opened 2021-10-21
Build AllReduce only mode

stanleytsang-amd:
You can now pass in --build_allreduce_only to the install script or set BUILD_ALLREDUCE_ONLY to cmake if not using the install script.  

With this mode on, only AllReduce + sum + float32 is generated for RCCL.  When combined with building unit tests, only the AllReduce-specific unit tests are built and only float and sum are run.

On rack-87, total build times with and without --build_allreduce_only are:

Library only
With: 5m22.345s
Without: 10m59.227s

Library + unit test executables
With: 5m42.022s
Without: 11m42.876s 

So about a 50% reduction in total build time in both cases.  AllReduce tests pass.

### DIFF
--- a/src/collectives/device/common.h
+++ b/src/collectives/device/common.h
@@ -16,7 +16,6 @@
 
 #define __syncwarp()
 
-#if !defined(BUILD_ALLREDUCE_ONLY)
 #define NCCL_FUNC5(func, algo, redop, type) \
   NCCL_FUNC_NAME(func, algo, LL,     redop, type), \
   NCCL_FUNC_NAME(func, algo, LL,     redop, type), \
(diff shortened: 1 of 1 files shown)


src/enqueue.cc
```diff
@@ -192,7 +192,9 @@ static ncclResult_t setupLaunch(struct ncclQueueInfo* eqInfo, int usingCudaGraph
       }
       if (elem->funcIndex != FUNC_INDEX_P2P && eqInfo->elemList->count() == 1) elem->active = 0;
     }
-
+#if defined(BUILD_ALLREDUCE_ONLY)
+    params->func = (void *)ncclKerns[0];
+#endif
```
wenkaidu, line 197: Most likely not needed. It has been set in line #190

PR #462, opened 2021-11-05
Fixing cmake_install_prefix search to include /opt/rocm-xxxx

stanleytsang-amd:
Now do a regex search for /opt/rocm*

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -81,7 +81,7 @@ if(BUILD_TESTS)
   # HIPCC adds /opt/rocm/lib as RPATH, even though the install process is supposed to
   # remove RPATH.  It also occurs before any user-specified rpath, which effectively overrides the user rpath.
   #  As a work-around, set the correct RPATH for the unit test executable as a post-install step
-  if (CMAKE_INSTALL_PREFIX STREQUAL "/opt/rocm")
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -81,7 +81,7 @@ if(BUILD_TESTS)
   # HIPCC adds /opt/rocm/lib as RPATH, even though the install process is supposed to
   # remove RPATH.  It also occurs before any user-specified rpath, which effectively overrides the user rpath.
   #  As a work-around, set the correct RPATH for the unit test executable as a post-install step
-  if (CMAKE_INSTALL_PREFIX STREQUAL "/opt/rocm")
+  if (CMAKE_INSTALL_PREFIX MATCHES "/opt/rocm*")
```
saadrahim, line 84: Can you control the order? What if you have 5.0 and 4.5 in the same system? This may result in unexpected builds.

wenkaidu (approved):

test/CMakeLists.txt
stanleytsang-amd: I'm not sure if we need to get that sophisticated here.  All we need to know here is that if the install prefix is set to any /opt/rocm (which is the default).  This check is just for determining where the unit tests were built (by default in the build/release folder, or in some other user specified non-/opt/rocm location).  This should not result in any unexpected builds.

PR #509, opened 2022-03-04
Force ring algorithm on single node

wenkaidu:


edgargabriel (approved):
I am assuming this gave the best results in your tests?

wenkaidu:
> I am assuming this gave the best results in your tests?

No, it made no difference in final result. But without it, tree algorithm was accidentally selected with some data sizes on single node.

PR #512, opened 2022-03-08
Packages for test and benchmark executables on all supported OSes using CPack.

lawruble13:


saadrahim (approved):

lawruble13:
The changes to test/CMakeLists.txt are required for installation. When CMake attempts to install the UnitTests executable it tries to rewrite its RPATH, which fails if it is specified with `chrpath`. Using the CMake variables avoids this problem.

PR #515, opened 2022-03-15
Replace hardcoded /opt/rocm with ROCM_PATH env var

minsii:
We install the ROCM package under home directory so that multiple staffs can share the same machine with different ROCM version.  However, several places in RCCL hardcode /opt/rocm. This patch tries to fix it.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -72,7 +72,7 @@ endif()
 
 
 # Find or download/install rocm-cmake project
-find_package(ROCM QUIET CONFIG PATHS /opt/rocm)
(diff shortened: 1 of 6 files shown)


wenkaidu (approved):

minsii:
@wenkaidu Thank you for the quick confirmation. Can we also include this fix in future RCCL releases? We expect to use the newer version of RCCL very soon on new AMD systems.

wenkaidu:
> @wenkaidu Thank you for the quick confirmation. Can we also include this fix in future RCCL releases? We expect to use the newer version of RCCL very soon on new AMD systems.

This should have been fixed already. If you check current develop branch, there should not be any hard coded /opt/rocm

PR #518, opened 2022-03-28
Update tuning parameters

wenkaidu:


edgargabriel:
@wenkaidu I am happy to provide the review if you want, but I am not sure I fully grasp the impact (and consequences) of this pr. That being said, I would curious on the process to change the treeCorrectionFactors. Did you execute a series of tests with different values, or is it the result of an analytical evaluation?

PR #529, opened 2022-04-07
Fixing the broken library soft link

nunnikri:
librrcl.so is is having a dead soft link. Which is causing issues with tensor flow build. Fixing it.


lawruble13:
Thanks @nunnikri, re-adding this was unintentional, so I'm happy to remove it once testing confirms functionality.

nunnikri:
Thanks @lawruble13. I tested locally, seems fine. Lets wait for the automated testing as well

wenkaidu (approved):

PR #540, opened 2022-04-27
fix the cmake logic to handle both old and new include

edgargabriel:
directory structure. I would suggest this as an alternative to pr 534, since it allows rccl to compile on both the old and the new directory layout.


CMakeLists.txt
```diff
@@ -222,11 +222,17 @@ if(COLLTRACE)
   add_definitions(-DENABLE_COLLTRACE)
 endif()
 
-CHECK_INCLUDE_FILE_CXX("${ROCM_PATH}/rocm_smi/include/rocm_smi/rocm_smi64Config.h" HAVE_ROCM_SMI64CONFIG)
```
saadrahim, line 225: Please see https://ontrack-internal.amd.com/browse/SWDEV-294432. rocm-smi-lib should have a cmake interface with find_package support.

edgargabriel: @saadrahim thank you for the pointer, is this support already available, or is it something to look out for in the near future? Not entirely clear based on the ticket.

wenkaidu: rocm-smi-lib cmake was added at ROCm 5.0. Since some setup are still using ROCm 4.5, we still need to keep backward compatibility.

PR #544, opened 2022-04-28
fix cmake logic to handle old and new include dirs

edgargabriel:
Starting from rocm 5.2 there is a reorganization of the include directories. This pr allows to compile RCCL on both the old and the new directory layout. This solution is using find_package() for identifying correct settings for rocm_smi starting from rocm-5.2, and the original (manual) settings for all previous releases.

Tested with rocm-5.2, 5.1.1, 5.0.2, and 4.5.2.


wenkaidu (approved):

saadrahim (approved):
Thanks for incorporating my suggestions.

PR #545, opened 2022-04-29
Adding debug CI build mode via label

gilbertlee-amd:


.jenkins/precheckin.groovy
```diff
@@ -17,7 +17,6 @@ def runCI =
 
     prj.timeout.test = 1440
     prj.paths.build_command = './install.sh -t '
-    prj.paths.build_command_debug = './install.sh -t --build-debug '
```
saadrahim: https://github.com/ROCmSoftwarePlatform/rocJenkins/blob/pong/src/com/amd/project/rocProject.groovy

there is no build_command_debug.

gilbertlee-amd: Yes - 2nd checkin addressed that.  Would it be better to have one in rocProject?

saadrahim: As long as it works, just use it for now. We are going to come back to all this code in the future and remove it from each individual repository and place in a central jenkins repo.

PR #550, opened 2022-05-19
Moving opt-in custom signal handler from UnitTests into RCCL

gilbertlee-amd:
- Moving the custom signal handler out of UnitTests and into the RCCL library
- Opt in via RCCL_ENABLE_SIGNALHANDLER=1
- Additional information provided if BFD library is installed (e.g. apt-get install binutils-dev for Ubuntu)


akolliasAMD:
IS there a way to add this `Additional information provided if BFD library is installed` to the readme or some readme?

edgargabriel (approved):

Issue #567, opened 2022-06-21
Missing roc:: namespace in rccl-config.cmake


dmikushin:
Hello, this is a bug report concerning RCCL packaged as a part of ROCm framework, as of 5.1.1. The file `rccl-config.cmake` should be changed as show below:

```diff
--- rccl-config.cmake.old  2022-06-21 20:35:34.815053058 +0000
+++ rccl-config.cmake  2022-06-21 18:58:15.774500081 +0000
@@ -91,14 +91,14 @@
 
 include(${rccl_TARGET_FILE})
 
-set(rccl_LIBRARIES rccl)
+set(rccl_LIBRARIES roc::rccl)
 
-set(rccl_LIBRARY rccl)
+set(rccl_LIBRARY roc::rccl)
 
-set(RCCL_LIBRARIES rccl)
+set(RCCL_LIBRARIES roc::rccl)
 
-set(RCCL_LIBRARY rccl)
+set(RCCL_LIBRARY roc::rccl)
 
-set(rccl_LIBRARIES rccl)
+set(rccl_LIBRARIES roc::rccl)
 
-set(rccl_LIBRARY rccl)
+set(rccl_LIBRARY roc::rccl)
```

Background. The missing namespace is necessary, in order to pick up targets defined in `rccl-targets-noconfig.cmake`. Without the namespace, CMake simply issues `-lrccl` to the linker command, which will fail. For example, PyTorch fails like this:

```
/opt/rh/devtoolset-7/root/usr/libexec/gcc/x86_64-redhat-linux/7/ld: cannot find -lrccl
```

In other libraries the namespace is present, for example in rocblas. So, this patch simply aligns RCCL CMake config with the other ROCm libraries.

The file `rccl-config.cmake` is not present in this repository, please help me to report it where it belongs to.

pfultz2:
This issue is that it lists the target as `rccl` instead of `roc::rccl` in `rocm_export_targets`: 

https://github.com/ROCmSoftwarePlatform/rccl/blob/develop/CMakeLists.txt#L351

dmikushin:
Someone made a point that end users should just link against the ROCm, and then it correctly brings the whole collection of libraries into the link command. That indeed works. On the other hand, the issue happens when the user is trying to link just against RCCL_LIBRARY.

doctorcolinsmith:
@pfultz2 Can this issue be closed?  Looks like the fix was completed.

dmikushin:
[...]

PR #570, opened 2022-06-22
Adding the missing roc:: namespace

dmikushin:
Adding the missing roc:: namespace, effectively changing the value of `RCCL_LIBRARY` from rccl to roc::rccl.

The important difference is that rccl is treated as a symbolic "-lrccl" by linker (and fail the linking
due to a missing library search path), while roc::rccl is a target name, which can resolve into an absolute
library path.

This patch corrects the export name as explained by @pfultz2 in #567 , similarly to as in [rocBLAS](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/CMakeLists.txt#L596)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,7 +348,7 @@ rocm_install(FILES ${PROJECT_BINARY_DIR}/include/rccl/rccl.h
 rocm_export_targets(NAMESPACE
                     roc::
                     TARGETS
-                    rccl
+                    roc::rccl
                     DEPENDS
                     hip)
 if(BUILD_FILE_REORG_BACKWARD_COMPATIBILITY)


gilbertlee-amd (approved):

saadrahim (changes requested):
Please update the changelog as this is major change that can impact end users. This may need to be done by a member of the RCCL team.

dmikushin:
@saadrahim , I've pushed a changelog entry proposal. @wenkaidu , @akolliasAMD does it look good to you?

dmikushin:
@saadrahim Greetings, does this look OK to you?

saadrahim (approved):

PR #596, opened 2022-08-11
Expand ncclTreeBasePostset for multi-rank

edgargabriel:
add multi-rank support for ncclTreeBasePostset. The algorithm maintains the same properties as the non-multirank version, i.e. the pair of GPUs communicating with each other stays the same.

This fix resolves the hang observed on Hayabusa when using all GPUs with the multi-rank feature.


edgargabriel:
Closing this pull request. Concept works fundamentally, but need to refine the code, since we cannot assume that ranks on a GPU are consecutive.

PR #614, opened 2022-09-09
Updating stream caching

gilbertlee-amd:
- Adding new optional stream to ncclCudaCalloc
- Adding a sideStream to ncclComm with same lifetime as communicator
- Using sideStream in calls to ncclCudaCalloc where comm is available
- Changing a test cudaCalloc to a cudaMalloc

### DIFF
--- a/src/transport/p2p.cc
+++ b/src/transport/p2p.cc
@@ -115,7 +115,7 @@ ncclResult_t p2pCanConnect(int* ret, struct ncclTopoSystem* topo, struct ncclTop
     // Check that legacy IPC support is available (WSL WAR)
     char *dummy;
     hipIpcMemHandle_t ipc;
-    NCCLCHECK(ncclCudaCalloc(&dummy, CUDA_IPC_MIN));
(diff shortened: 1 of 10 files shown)


src/transport/p2p.cc
```diff
@@ -115,7 +115,7 @@ ncclResult_t p2pCanConnect(int* ret, struct ncclTopoSystem* topo, struct ncclTop
     // Check that legacy IPC support is available (WSL WAR)
     char *dummy;
     hipIpcMemHandle_t ipc;
-    NCCLCHECK(ncclCudaCalloc(&dummy, CUDA_IPC_MIN));
+    CUDACHECK(hipMalloc(&dummy, CUDA_IPC_MIN));
```
wenkaidu, line 118: why this is changed?

gilbertlee-amd: This dummy memory is immediately freed after this.  It doesn't need to be calloc-ed.

wenkaidu (approved):

PR #615, opened 2022-09-09
add binary tree

edgargabriel:
In addition, introduce the ability to have 2 trees at the same time. Only for allreduce at the moment.


src/init.cc
```diff
@@ -1027,17 +1030,23 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, ncclUniqueId* comm
...
 
-  char line[1024];
+  char line[1024], binline[1024];
   line[0]='\0';
+  binline[0]='\0';
   for (int c=0; c<comm->nChannels; c++) {
     struct ncclTree* tree = &comm->channels[c].tree;
-    snprintf(line+strlen(line), 1023-strlen(line), " [%d] %d/%d/%d->%d->%d",
+    struct ncclTree* binTree = &comm->channels[c].binTree;
+    snprintf(line+strlen(line), 1023-strlen(line), " [%d] %d/%d/%d->%d->%d\n",
```
wenkaidu, line 1039: need to remove \n

edgargabriel: done

wenkaidu (approved):

PR #623, opened 2022-09-20
Relax memory ordering used in atomic operations.

whchung:


wenkaidu:
Need to confirm it won't break multinode

PR #624, opened 2022-09-20
Tweak unroll factors.

whchung:
Remove #pragma unroll 1 for places known not unrollable.
Instill #pragma unroll on places known to be unrollable.

### DIFF
--- a/src/collectives/device/prims_ll.h
+++ b/src/collectives/device/prims_ll.h
@@ -183,7 +183,7 @@ private:
 
   template<int BeginIx>
   __device__ void readLLBeginAll(int offset, ncclLLFifoLine(&line)[MaxRecv]) {
-    #pragma unroll 1
+    #pragma unroll
     for (int i=BeginIx; i < MaxRecv; i++) {
       if (i < fan.nrecv()) {
         union ncclLLFifoLine* src = recvPtr(i) + offset;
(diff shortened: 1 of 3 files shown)


wenkaidu (approved):

src/collectives/device/common.h
```diff
@@ -449,7 +449,6 @@ struct RunWork {
     int wid = threadIdx.x / WARP_SIZE;
     ncclWorkElem* we = w->header.type == ncclWorkTypeRegColl ? &w->regElems[0].elem : &w->elems[0];
     int stride = w->header.type == ncclWorkTypeRegColl ? sizeof(ncclWorkElemReg) : sizeof(ncclWorkElem);
-    #pragma unroll 1
```
wenkaidu, line 452: Suggest to revert this change. Original unroll 1 is in NCCL. Changing this may have unexpected effect.

src/collectives/device/onerank_reduce.cu
```diff
@@ -16,7 +16,7 @@ namespace {
     ncclWork *w = &ncclShmem.work;
     int tid = threadIdx.x;
     int tn = blockDim.x;
-    #pragma unroll 1
+    #pragma unroll
```
wenkaidu, line 19: Suggest to revert this change. Original unroll 1 is in NCCL. Changing this may have unexpected effect.

PR #626, opened 2022-09-21
HOTFIX: Remove backwards compatibility headers on Windows

lawruble13:
The backwards compatibility headers should not be generated or installed on Windows.


lawruble13:
Opened to incorrect branch.

PR #630, opened 2022-09-26
Add LL128 tuning

wenkaidu:


whchung:
@wenkaidu Probably need to document the rationale behind this PR as it's not easily understand the idea. I can see it introduces a new option, and it's tied to a specific message size. But why we need to do that is not obvious yet. I can imagine there are empirical experimental results to justify this PR and perhaps it should be documented.

akolliasAMD (approved):

gilbertlee-amd (approved):

PR #641, opened 2022-10-20
Sync up with NCCL 2.14.3

wenkaidu:


edgargabriel (approved):
compiled and run the branch on H26 with ROCm 5.2, looks good as far as I can see.

PR #662, opened 2022-12-09
UnitTest: add test cases for 2.14 API (ncclCommInitRankConfig and ncc…

PedramAlizadeh:
…lCommFinalize for non-blocking communicator)


test/common/TestBedChild.cpp
```diff
@@ -177,6 +178,29 @@ namespace RcclUnitTesting
...
+        } 
+        do 
+        {
+          ncclCommGetAsyncError(this->comms[localRank], &ncclAsyncErrInit);
+        } while(ncclAsyncErrInit != ncclSuccess);
+        if (ncclAsyncErrInit != ncclSuccess)
+        {
+          ERROR("Rank %d on child %d fails NCCL call %s with code %d\n", globalRank, this->childId, "ncclCommGetAsyncErrorInit", ncclAsyncErrInit);
+          status = TEST_FAIL;
+          break;
```
gilbertlee-amd, line 201: To actually "test" out this feature, I think we can move this outside of the local rank loop (and only tested if nonblocking init requested).  Additionally, we need to have some sort of counter to prevent infinite loop, if for some reason, ncclCommGetAsyncError never reaches ncclSuccess.

test/common/TestBedChild.cpp
```diff
@@ -187,10 +211,38 @@ namespace RcclUnitTesting
...
+        ncclResult_t Group_End_state = ncclGroupEnd();
+        for (int localRank = 0; localRank < numGpus; ++localRank)
+        {
+          ncclResult_t ncclAsyncErrGroupEnd;
+          if (Group_End_state == ncclInProgress) 
+          {
+            do 
+            {
+              ncclCommGetAsyncError(this->comms[localRank], &ncclAsyncErrGroupEnd);
+            } while(ncclAsyncErrGroupEnd == ncclInProgress);
```
gilbertlee-amd, line 230: Add counter to prevent infinite loop

test/common/TestBedChild.cpp
```diff
@@ -187,10 +211,38 @@ namespace RcclUnitTesting
...
+          {
+            ERROR("Child process %d fails NCCL call %s with code %d\n", this->childId, "ncclCommGetAsyncErrorGroupEnd", ncclAsyncErrGroupEnd);
+            return TEST_FAIL; 
+          }
+        }
+      }
+      else 
+      { 
+        // In case of blocking communication just call ncclGroupEnd
+        CHILD_NCCL_CALL(ncclGroupEnd(), "ncclGroupEnd");
```
gilbertlee-amd, line 242: Perhaps it would make more sense dealing with the blocking vs non-blocking calls by incorporating ncclCommGetAsyncError directly into the CHILD_NCCL_CALL macro

gilbertlee-amd (changes requested):

gilbertlee-amd (approved):

PR #683, opened 2023-02-13
NPKit: improve clock calibration and fix GPU clock API

yzygitzh:


src/include/npkit/npkit.h
```diff
@@ -14,6 +14,12 @@
 #include "npkit/npkit_event.h"
 #include "npkit/npkit_struct.h"
 
+#if defined(__gfx1100__) || defined(__gfx1101__) || defined(__gfx1102__)
```
wenkaidu, line 17: since __builtin_amdgcn_s_memrealtime is available in gfx9 and gfx10, we can use following to be future proof:
#ifdef __GFX9__ || __GFX10__
#define NPKIT_GET_GPU_TIMESTAMP __builtin_amdgcn_s_memrealtime
#else
#define NPKIT_GET_GPU_TIMESTAMP clock64
#endif

src/misc/npkit.cc
```diff
@@ -17,19 +17,30 @@ NpKitEvent** NpKit::cpu_event_buffers_ = nullptr;
...
+  uint64_t cpu_timestamp = *cpu_timestamp_in;
+  uint64_t gpu_timestamp_2 = NPKIT_GET_GPU_TIMESTAMP();
+  *gpu_timestamp = gpu_timestamp_1 + (gpu_timestamp_2 - gpu_timestamp_1) / 2;
+  *cpu_timestamp_out = cpu_timestamp;
+}
+
+static volatile bool cpu_timestamp_update_thread_should_stop = false;
+static volatile uint64_t* volatile_cpu_timestamp = nullptr;
 
 void NpKit::CpuTimestampUpdateThread() {
```
wenkaidu, line 37: In general, since RTC is available on GFX9 and GFX10, we would prefer to use known RTC clock instead of CPU time stamp. This is more accurate and also save system resources (no extra CPU thread).

src/include/npkit/npkit.h
yzygitzh: Fixed.

src/misc/npkit.cc
yzygitzh: Currently there is no CPU timestamp thread anymore. This is only used in initialization phase.

wenkaidu (approved):

PR #696, opened 2023-02-28
Warn user on incorrect system settings

wenkaidu:


src/init.cc
```diff
@@ -108,6 +109,18 @@ static ncclResult_t ncclInit() {
     NCCLCHECK(bootstrapNetInit());
     NCCLCHECK(ncclNetPluginInit());
 
+    char strValue[MAX_STR_LEN];
+    NCCLCHECK(ncclTopoGetStrFromSys("/proc/sys/kernel", "numa_balancing", strValue));
+    if (strcmp(strValue, "1") == 0)
+      WARN("NUMA auto blancing enabled! Disable by \"sudo sysctl kernel.numa_balancing=0\"");
```
gilbertlee-amd, line 115: balancing

gilbertlee-amd (changes requested):
minor typo

src/init.cc
wenkaidu: fixed

gilbertlee-amd (approved):

PR #727, opened 2023-04-15
Do not add barrier after send

whchung:
On AMD GCN ISA, the semantic of nontemporal store already guarantee data fully visible on remote GPU. There shouldn't be no barriers needed after send, under all-to-all communication pattern.

Empirical performance benchmarking shows improvement in smaller message sizes with correct results.

### DIFF
--- a/src/collectives/device/msccl_kernel_impl.h
+++ b/src/collectives/device/msccl_kernel_impl.h
@@ -312,7 +312,7 @@ __device__ __forceinline__ void mscclRunInterpreter(
               NpKit::CollectGpuEventLDS(NPKIT_EVENT_MSCCL_SEND_ENTRY, thisNelem*sizeof(T), 0, NPKIT_GET_GPU_TIMESTAMP());
             }
 #endif	
-          prims.sendWithBarrier(srcOffset, thisNelem); // LL.send is the only situation where there is no barrier at the end.
+          prims.send(srcOffset, thisNelem); // LL.send is the only situation where there is no barrier at the end.
 
 #if defined(ENABLE_NPKIT) && defined(ENABLE_NPKIT_EVENT_MSCCL_SEND_EXIT)
             if (tid == 0) {


yzygitzh:
Hi,

This barrier was designed to avoid data corruption when data from the same address is touched by different threads.
Imagine there is 1 thread block which has 2 threads, and they're doing send first and then recv.
There are 4 data slots (0, 1, 2, 3).
For send, thread 0 is going to send (0, 1) and thread 1 is going to send (2, 3).
For recv, thread 0 is going to recv (0) and thread 1 is going to recv (1).
Now, without barrier, thread 1 might go ahead of thread 0 and wrongly overwrites data slot 1.

This is not a common case while some special algorithms we use will have different send size and recv size. One potential way to solve this is to differentiate "send w/ barrier" and "send w/o barrier" in MSCCL algorithm interface.

src/collectives/device/prims_ll.h
```diff
@@ -744,8 +744,6 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
   // MSCCL primitives
   __device__ void sendWithBarrier(intptr_t inpIx, int eltN) {
     send(inpIx, eltN);
-    // This is the only primitive.instruction where there is no barrier at the end, add it
-    barrier();
```
nusislam, line 748: Will removing this barrier cause any synchronization issue?

whchung:
Closing this PR after yielding no perceivable improvements on target machines.

whchung:
Reviving the PR as it proves to be helpful on upcoming SKUs. The logic needs to modified so it's only applied under LL protocl though.

yzygitzh:
> Reviving the PR as it proves to be helpful on upcoming SKUs. The logic needs to modified so it's only applied under LL protocl though.

Let me help drive a PR to introduce "swb" (send with barrier) instruction in XML to enable this while not breaking non-safe algorithms.

wenkaidu (approved):

PR #742, opened 2023-05-17
Allow skipping doneEvent inside MSCCL.

whchung:
Added a RCCL_MSCCL_ENABLE_DONE_EVENT env var, set it be 1 by default.

The env var is to control whether to use doneEvent when invoking MSCCL kernels.

Skipping doneEvent would cause the firmware to skip L2 cache flush, resulting in overall performance improvement.

### DIFF
--- a/src/misc/msccl/msccl_setup.cc
+++ b/src/misc/msccl/msccl_setup.cc
@@ -13,6 +13,8 @@
 #include "msccl/msccl_setup.h"
 #include "msccl/msccl_status.h"
 
+RCCL_PARAM(MscclEnableDoneEvent, "MSCCL_ENABLE_DONE_EVENT", 0);
+
 ncclResult_t mscclSetupCount(struct mscclAlgo* hostAlgo, ncclComm_t comm, size_t count, ncclDataType_t dataType) {
   mscclStatus& status = mscclGetStatus();
   status.stepSize = comm->buffSizes[hostAlgo->protocol] / NCCL_STEPS;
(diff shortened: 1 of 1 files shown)


wenkaidu (approved):

src/misc/msccl/msccl_setup.cc
```diff
@@ -260,7 +264,14 @@ ncclResult_t mscclSetupKernel(const void* sendBuff, void* recvBuff, size_t count
     ncclComm_t comm, hipStream_t stream) {
   mscclStatus& status = mscclGetStatus();
 
-  if (status.lastStream != stream && status.lastStream != nullptr) {
+  bool enableDoneEvent =
+#ifndef HIP_EVENT_DISABLE_FENCE
+	  (rcclParamMscclEnableDoneEvent() == 1);
+#else
+          true;
```
pash-msft, line 271: Please correct me if my understanding is not correct, if HIP_ENABLE_DISABLE_EVENT is set, then enableDoneEvent should be set to false.

src/misc/msccl/msccl_setup.cc
```diff
@@ -13,6 +13,10 @@
 #include "msccl/msccl_setup.h"
 #include "msccl/msccl_status.h"
 
+#ifndef HIP_EVENT_DISABLE_FENCE
```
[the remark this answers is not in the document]
whchung: With `HIP_EVENT_DISABLE_FENCE`, we shouldn't need any logic introduced in this commit, so we use the macro to guard this parameter.

src/misc/msccl/msccl_setup.cc
whchung: With `HIP_EVENT_DISABLE_FENCE`, `doneEvent` will be instantiated with a bit hinting the hardware not to flush L2 cache, so we can safely set `enableDoneEvent` as `true`.

pash-msft (approved):

PR #757, opened 2023-05-25
device: change unroll factor

nusislam:
The default value of the unroll factor is 2. Changing the unroll factor to 4 provides better performance for most of the collectives.


nusislam:
The test failures do not seem related to this PR.

wenkaidu (approved):

PR #762, opened 2023-05-31
Removing init_nvtx.cc from source list

gilbertlee-amd:
- To fix undefined symbol: nvtxPayloadEnumRegister

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,7 +376,7 @@ set(SRC_FILES
   src/include/trees.h
   src/include/utils.h
   src/init.cc
-  src/init_nvtx.cc
+#  src/init_nvtx.cc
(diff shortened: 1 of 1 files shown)


wenkaidu (approved):

CMakeLists.txt
```diff
@@ -376,7 +376,7 @@ set(SRC_FILES
   src/include/trees.h
   src/include/utils.h
   src/init.cc
-  src/init_nvtx.cc
+#  src/init_nvtx.cc
```
nusislam, line 379: Do you like to keep it commented or remove entirely?

gilbertlee-amd: I prefer to keep it commented out, so that we purposelessly know it was left out.

PR #773, opened 2023-06-13
improve compilation time and create timetrace plot

BertanDogancay:


wenkaidu (approved):

BertanDogancay:
* Added a function to generate some of the device files (all_reduce, reduce, reduce_scatter and msccl_kernel) in Dependencies.cmake and share the work among the threads to reduce the compilation time.
* Commented out the contents of the original files located in src/collectives/device for all_reduce, reduce, reduce_scatter, msccl_kernel and commented out their paths in CMakeLists.txt.
* Deleted ncclTypes in common.h (not used).
*  Changed msccl_kernel.cu to msccl_kernel_impl.h (need it to generate files in CMake) and added the path in CMakeLists.txt.
* Added time-trace flag in install.sh. If enabled, will generate html file to visualize the build time of RCCL.
* Created a python script to extract the data from the build log file and generate plot (located in tools/time-trace).
* Created rccl-TimeTrace.sh to check for dependencies to run the python script and connect it to RCCL.
* Added max-jobs flag in install.sh to enable max number of procs when building. If not used, default value is 16.

nusislam (approved):

PedramAlizadeh (approved):

PR #778, opened 2023-06-20
Disable Colltrace for --fast option

BertanDogancay:
* Disabled generating additional debug kernels when using the fast option in install.sh.
* Removed the --max-jobs flag and added the --limit-nprocs flag. The default value is now set to the maximum number of processors. The limiting functionality is intended for use in the CI environment.
* Modified time_trace_generator.py and rccl-TimeTrace.sh to avoid specifying static paths to folder and file locations.
* Modified topo.h, common.h, and enqueue.cc to resolve compiler errors when colltrace is disabled.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,8 +16,8 @@ function display_help()
     echo "       --fast                  Quick-build RCCL (local gpu arch only, no backtrace support)"
     echo "    -h|--help                  Prints this help message"
     echo "    -i|--install               Install RCCL library (see --prefix argument below)"
+    echo "    -l|--limit-nprocs          Limit the number of procs to 16 while building"
     echo "       --local_gpu_only        Only compile for local GPU architecture"
-    echo "       --max-jobs              Use nproc instead of default number of 16"
     echo "       --no_clean              Don't delete files if they already exist"
     echo "       --npkit-enable          Compile with npkit enabled"
     echo "    -p|--package_build         Build RCCL package"
@@ -53,7 +53,7 @@ build_static=false
 build_tests=false
 build_verbose=0
 time_trace=false
-enable_all_jobs=false
+enable_all_jobs=true
 enable_ninja=""
 
 # #################################################
@@ -63,7 +63,7 @@ enable_ninja=""
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
-    GETOPT_PARSE=$(getopt --name "${0}" --longoptions address-sanitizer,build_allreduce_only,dependencies,debug,disable_backtrace,fast,help,install,local_gpu_only,no_clean,npkit-enable,package_build,prefix:,rm-legacy-include-dir,run_tests_all,run_tests_quick,tests_build,time-trace,max-jobs,verbose --options hidptrs -- "$@")
+    GETOPT_PARSE=$(getopt --name "${0}" --longoptions address-sanitizer,build_allreduce_only,dependencies,debug,disable_backtrace,fast,help,install,limit-nprocs,local_gpu_only,no_clean,npkit-enable,package_build,prefix:,rm-legacy-include-dir,run_tests_all,run_tests_quick,tests_build,time-trace,verbose --options hidptrs -- "$@")
 else
     echo "Need a new version of getopt"
     exit 1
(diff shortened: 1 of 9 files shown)


wenkaidu (approved):

install.sh
```diff
@@ -16,8 +16,8 @@ function display_help()
     echo "       --fast                  Quick-build RCCL (local gpu arch only, no backtrace support)"
```
gilbertlee-amd, line 16: This line should be updated to reflect the options that occur when fast is picked

install.sh
```diff
@@ -192,6 +192,7 @@ fi
 # Build local GPU arch only
 if [[ "$build_local_gpu_only" == true ]]; then
     cmake_common_options="${cmake_common_options} -DBUILD_LOCAL_GPU_TARGET_ONLY=ON"
+    cmake_common_options="${cmake_common_options} -DCOLLTRACE=OFF"
```
gilbertlee-amd, line 195: this shouldn't be done here - should be part of fast.

install.sh
```diff
@@ -86,8 +86,8 @@ while true; do
          --fast)                     build_bfd=false; build_local_gpu_only=true; shift ;;
```
gilbertlee-amd, line 86: This is where the the option for disabling collective trace should go

gilbertlee-amd (changes requested):

install.sh
```diff
@@ -16,8 +16,8 @@ function display_help()
     echo "       --fast                  Quick-build RCCL (local gpu arch only, no backtrace support)"
     echo "    -h|--help                  Prints this help message"
     echo "    -i|--install               Install RCCL library (see --prefix argument below)"
+    echo "    -l|--limit-nprocs          Limit the number of procs to 16 while building"
     echo "       --local_gpu_only        Only compile for local GPU architecture"
```
gilbertlee-amd, line 20: Should be able to disable collective trace here with its own flag, not just only through --fast

install.sh
gilbertlee-amd: This hasn't been fixed still

tools/time-trace/rccl-TimeTrace.sh
```diff
@@ -47,4 +47,4 @@ mv temp_file.txt "$directory/.ninja_log"
 mv "$directory/.ninja_log" "$directory/time_trace.log"
 
 # Run the python program
-python3 time_trace_generator.py --min_val 5 --include_linking
\ No newline at end of file
+python3 time_trace_generator.py --min_val 5 --include_linking --log_file_path ../../build/release/time_trace.log
```
gilbertlee-amd, line 50: This continues to assume that files will be in ../build/release, which isn't always the case.  This directory might not even exist, or a debug library might have been build in ../build/debug.  
The log file directory needs to be an input parameter to this script, which can default to ../build/release if not specified.

gilbertlee-amd (changes requested):

install.sh
```diff
@@ -84,6 +85,7 @@ while true; do
     -d | --dependencies)             install_dependencies=true;                                          shift ;;
          --debug)                    build_release=false;                                                shift ;;
          --disable_backtrace)        build_bfd=false;                                                    shift ;;
+         --disable-colltrace)        collective_trace=false;                                             shift ;;
```
nusislam, line 88: Instead of adding a new commit addressing the reviewer's comments, it is possible to amend the existing commits. That way, when the PR is merged, you do not have to merge the intermediate versions of the commits, you will be merging the final version only. In my opinion, this way we will be merging a cleaner version of the PR and it is better in terms of long-term maintenance of the repo.

BertanDogancay: @nusislam What about squash and merge? Wouldn't that also work?

nusislam: Probably not. If you have one commit addressing the reviews for three initial commits, squashing them will result into a single commit. But we need to keep the three initial commits. In open-source repositories like MPICH, they usually do interactive `rebase` to `amend` individual commits.

wenkaidu: I think if the PR is for one feature or issue and there isn't need to keep multiple commit in final version, then "squash and merge" should be fine. If in some cases we want to keep multiple commits for any reason, then we need to do rebase and force push

gilbertlee-amd (approved):

PR #804, opened 2023-07-13
Add GPU P2P ping-pong latency test tool

yzygitzh:


wenkaidu (approved):

tools/p2p-latency-test/p2p_latency_test.cpp
```diff
@@ -0,0 +1,112 @@
...
+      return -1;
+    }
+    if (self_mode == PING_MODE) {
+      PingKernel<<<1, 1, 0, stream>>>(local_flag, remote_flag, time_delta);
+    } else {
+      PongKernel<<<1, 1, 0, stream>>>(local_flag, remote_flag, time_delta);
+    }
+    hipStreamSynchronize(stream);
+    if (self_mode == PING_MODE) {
+      printf("Ping-pong latency in us: %g\n", double(*time_delta) / NUM_LOOPS_RUN * 1e6 / VEGA_GPU_RTC_FREQUENCY);
```
nusislam, line 104: Are you reporting the round-trip latency? The one way latency should be round-trip/2/

yzygitzh: I was reporting round-trip latency, but I changed to one-way latency now since it's a little bit misleading. VEGA_GPU_RTC_FREQUENCY is gfx90a's rtc frequency, I made it a variable to support gfx94x as well.

nusislam (approved):

PR #825, opened 2023-07-26
ll_latency_test: fix time calculation

wenkaidu:


tools/p2p-latency-test/ll_latency_test.cpp
```diff
@@ -91,7 +91,8 @@ __global__ void PingKernel(LLFifoLine* local_flag, LLFifoLine* remote_flag, uint
     while (readLL(local_flag+tid, i) != i);
   }
   uint64_t end_time = wall_clock64();
-  *time_delta = end_time - start_time;
+  __syncthreads();
+  if (tid == 0) *time_delta = end_time - start_time;
```
nusislam, line 95: Instead of taking the latency of thread-0, should we take the max latency among all the threads?

wenkaidu: __syncthreads() will wait for all threads to reach the barrier, thus we will measure max latency

nusislam: The `end_time` is taken before calling `__syncthreads()`. The time required to sync among the threads is not being measured in the latency.

nusislam (approved):

tools/p2p-latency-test/ll_latency_test.cpp
gilbertlee-amd: Only thread 0 needs to collect wall_clock64().
if (tid == 0) *time_delta = wall_clock64() - start_time;

wenkaidu: I know, but think not worth it to create diversion. Would rather let all thread do the same

gilbertlee-amd: You potentially add additional latency if other wavefronts (not thread 0's) all have to compute wall_clock64() before thread 0's wavefront.

wenkaidu: as you wish...

PR #857, opened 2023-08-18
p2p/ll-latency-test: convert to single thread tests

wenkaidu:


nusislam:
Why are you converting it to a single process test? Isn't multi-process test the most common use-case?

wenkaidu:
> Why are you converting it to a single process test? Isn't multi-process test the most common use-case?

purpose to make it easier working with other HW debug tools

PR #860, opened 2023-08-21
Fix static_assert string literal that contains a "\%".

ddpagan:
This is no longer valid. They can only be simple escape sequences. Removing '\' fixes issue. Assert message now compiles and emits the '%' as expected.


wenkaidu (approved):

PedramAlizadeh (approved):

ronlieb:
@wenkaidu  what is the process to get this visible to the nPSDB  for amd-stg-open and amd-mainline-open ?

PR #866, opened 2023-08-25
rename of NCCL_TREES variable

akolliasAMD:


wenkaidu (approved):

src/graph/search.cc
```diff
@@ -887,7 +887,7 @@ ncclResult_t ncclTopoCompute(ncclTopoSystem* system, struct ncclTopoGraph* graph
   }
 
   str = getenv("NCCL_RINGS");
-  char* strTrees = getenv("NCCL_TREES");
+  char* strTrees = getenv("RCCL_TREES");
```
gilbertlee-amd, line 890: Is there any reason why we aren't using RCCL_PARAM from (include/param.h)

gilbertlee-amd (changes requested):

PR #881, opened 2023-09-07
Remove reliance on deprecated gcnArch

astatide:
This is a draft for removing the now deprecated gcnArch; we use CMake to determine if we're compiling against a version of ROCm that supports gcnArchName and handles architecture checking appropriately.  It includes a few helper functions as drop ins for the functionality we used gcnArch for before; sometimes to enable flags, and sometimes to set frequencies.  If we're using a version of ROCm that allows us to directly query the frequency, we directly query it.

Comments appreciated!  I'm especially looking for feedback on whether `archinfo.cc/h` should be moved, or if the functionality should be merged into another file, etc.


src/clique/CliqueManager.cc
```diff
@@ -266,12 +266,12 @@ void CliqueManager::SetByteLimits()
...
   if (m_allReduceByteLimit == 0)
   {
-    switch (m_gcnArch)
-    {
-    case 906: m_allReduceByteLimit =  16777216; break;
-    case 908: m_allReduceByteLimit =   8388608; break;
-    default:  m_allReduceByteLimit =  16777216; break;
-    }
+    if (strncmp(m_gcnArchName, "gfx906", 6) == 0)
+      m_allReduceByteLimit = 16777216; // TODO: do we really need this, since it's covered by the default?  Check if it's more of a documentation / assumption thing.
```
gilbertlee-amd, line 270: Clique isn't used anymore, but this was done explicitly so that it's easy to tell that it had been "optimized" for 906.
Please leave it in.

src/graph/topo.cc
```diff
@@ -369,7 +369,17 @@ ncclResult_t ncclTopoAddNic(struct ncclXmlNode* xmlNic, struct ncclTopoSystem* s
...
+  if (strcmp(gpu->gpu.gcn, "906") == 0) {
+    //NCCLCHECK(xmlSetAttr(gpu->gpu, "gcn", "gfx906"));
+    gpu->gpu.gcn = "gfx906";
+  } else if (strcmp(gpu->gpu.gcn, "908") == 0) {
+    //NCCLCHECK(xmlSetAttr(gpu->gpu, "gcn", "gfx908"));
+    gpu->gpu.gcn = "gfx908";
+  } else if (strcmp(gpu->gpu.gcn, "910") == 0) {
+    //NCCLCHECK(xmlSetAttr(gpu->gpu, "gcn", "gfx90a"));
+    gpu->gpu.gcn = "gfx90a";
+  }
```
gilbertlee-amd, line 382: Is there any reason to keep the extra NCCLCHECK comments?

src/graph/topo.h
```diff
@@ -224,17 +225,13 @@ static ncclResult_t ncclTopoDevToRank(struct ncclTopoSystem* system, int dev, in
...
-    case 941:
-    case 942:
-      return GFX94X_XGMI_WIDTH;
-    default:
-      return VEGA_XGMI_WIDTH;
-  }
+static float ncclTopoXGMISpeed(const char* gcn) {
+  if (strncmp(gcn, "gfx90a", 6) == 0)
+    return MI200_XGMI_WIDTH;
+  else if (strncmp(gcn, "gfx940", 6) == 0 || strncmp(gcn, "gfx941", 6) == 0 || strncmp(gcn, "gfx942", 6) == 0) // TODO check if we can just use strncmp(gcn, "gfx94", 5)
```
gilbertlee-amd, line 231: Yes.  gfx94 should suffice.

src/clique/CliqueManager.cc
```diff
@@ -266,12 +266,12 @@ void CliqueManager::SetByteLimits()
   m_allReduceByteLimit = rcclParamAllReduceCliqueByteLimit();
   if (m_allReduceByteLimit == 0)
   {
-    switch (m_gcnArch)
-    {
-    case 906: m_allReduceByteLimit =  16777216; break;
-    case 908: m_allReduceByteLimit =   8388608; break;
-    default:  m_allReduceByteLimit =  16777216; break;
-    }
+    if (strncmp(m_gcnArchName, "gfx906", 6) == 0)
```
gilbertlee-amd, line 269: strncmp(archname, target, 6) gets repeated a lot through the code/

I think you can just add a helper function in archinfo,h

Perhaps
bool ArchMatches(char const* arch, char const* target)
{
  return strncmp(arch, target, strlen(target));
}

This should work even if you want to just compare for "gfx94"

src/init.cc
```diff
@@ -173,15 +174,11 @@ RCCL_PARAM(KernelCollTraceEnable, "KERNEL_COLL_TRACE_ENABLE", 0);
...
-  hipDeviceProp_t devProp;
   double vega_gpu_rtc_freq;
 
   memset(head, 0, sizeof(int)*MAXCHANNELS);
-  hipError_t status = hipGetDeviceProperties(&devProp, comm->cudaDev);
-  if (devProp.gcnArch/10 == 94 && status == hipSuccess)
-    vega_gpu_rtc_freq = 1.0E8;
-  else
-    vega_gpu_rtc_freq = 2.5E7;
+  // Many AMD cards currently supported by RCCL have the same frequency; we're testing for the lone set of exceptions.
```
gilbertlee-amd, line 180: This comment doesn't quite make sense at this location.
Perhaps you mean to put it in getDeviceWallClockRateInKhz

src/include/archinfo.h
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+#include <hip/hip_runtime_api.h>
+#include <hip/hip_runtime.h>
+*/
+
+void gcnArchNameFormat(char *gcnArchName, char* out);
+void gcnArchConvertToGcnArchName(int gcnArch, char* out);
+int getGcnArchName(int deviceId, char* out);
+double getDeviceWallClockRateInKhz(int deviceId);
```
gilbertlee-amd, line 34: I think generally for code that we write (not NCCL), we try to use CamelCase for functions (starting with capital), and leave lowercase camelcase for variables.

gilbertlee-amd (changes requested):

src/graph/topo.cc
astatide: hahaha, nope, just forgot to delete em.  Thanks!

src/clique/CliqueManager.cc
astatide: ah, that's a great idea!  Thank you.

tools/topo_expl/utils.cpp
```diff
@@ -862,17 +862,17 @@ ncclResult_t initTransportsRank_1(struct ncclComm* comm, struct allGatherInfo *a
   NCCLCHECK(ncclTopoIdToIndex(comm->topo, GPU, comm->busId, &idx));
   allGather3Data[rank].nc = 2;
   if (comm->topo->nodes[GPU].count == comm->topo->nRanks &&
-       comm->topo->nodes[GPU].nodes[idx].gpu.gcn == 906 && allXgmi)
+       strncmp(comm->topo->nodes[GPU].nodes[idx].gpu.gcn, "gfx906", 6) == 0 && allXgmi)
```
gilbertlee-amd, line 865: IsArchMatch could go here too

gilbertlee-amd (changes requested):

gilbertlee-amd (approved):

PR #886, opened 2023-09-12
Gcn arch name

astatide:
New PR to do a squash merge instead of a regular one... text copied from the original PR.

We use CMake to determine if we're compiling against a version of ROCm that supports gcnArchName and handles architecture checking appropriately. It includes a few helper functions as drop ins for the functionality we used gcnArch for before; sometimes to enable flags, and sometimes to set frequencies.


src/clique/CliqueManager.cc
```diff
@@ -368,23 +368,18 @@ ncclResult_t CliqueManager::GetNumChannelsToUse(ncclFunc_t const coll,
     {
       // NOTE: These are currently based on collected data and not necessarily ideal for all hardware
       int numChannels;
-      switch (m_gcnArch)
-      {
-      case 906:
+      if (ArchName(m_gcnArchName, "gfx906")) {
```
gilbertlee-amd, line 371: IsArchMatch?

astatide: ... how in the hell did that even compile?

gilbertlee-amd (approved):

PR #893, opened 2023-09-15
Reduce NPKit latency overhead in MSCCL kernel

wenkaidu:


whchung:
From the look of it, we use LDS to track NPKIT data instead of global to reduce overhead, right?

wenkaidu:
> From the look of it, we use LDS to track NPKIT data instead of global to reduce overhead, right?

yes, only dump to global at kernel exit. This will minimize impact during kernel execution

akolliasAMD (approved):

PR #911, opened 2023-10-06
Change MSCCL kernel signature to allow kernel arguments be preloaded via SGPR

whchung:
Use newly introduced compiler option `-mllvm --amdgpu-kernarg-preload-count=16` to teach the compiler preload kernel arguments via SGPR.

Modify MSCCL kernel signature so only vanilla pointers are passed to MSCCL kernels to leverage this feature.

Modify NPKIT tracing logic to accommodate modified `mscclWork` data structure.

Please squash it when merging.


src/collectives/device/msccl_kernel_impl.h
wenkaidu, line 1: Need to add:
  if (tid == 0)
    *mscclShmem.work.workFifoDone = mscclShmem.work.workFifoDoneAck;

whchung: added one commit to adopt this logic.

wenkaidu (approved):

PR #920, opened 2023-10-18
remove gcnArch support

nileshnegi:
`gcnArch` will be removed in ROCm 6.0, and `gcnArchName` will be used instead. `gcnArchName` was introduced in ROCm 3.7.0, so removing all occurrences of `gcnArch`.

### DIFF
--- a/src/graph/xml.cc
+++ b/src/graph/xml.cc
@@ -636,15 +636,22 @@ ncclResult_t ncclTopoGetXmlFromGpu(struct ncclXmlNode* pciNode, uint32_t rocmDev
   int sm;
   NCCLCHECK(xmlGetAttrInt(gpuNode, "sm", &sm));
 
-  int gcn;
+  const char* gcn;
+  const char* gcnArchName;
   NCCLCHECK(xmlGetAttrIndex(gpuNode, "gcn", &index));
   if (index == -1) {
     hipDeviceProp_t devProp;
     CUDACHECK(hipGetDeviceProperties(&devProp, 0));
-    gcn = devProp.gcnArch;
-    NCCLCHECK(xmlSetAttrInt(gpuNode, "gcn", gcn));
+    //extract only the releveant info from the gcnArchName attribute
+    //e.g.: convert "gfx908:sramecc+:xnack-" to "gfx908"
+    char gcnArchNameSubstr[6];
+    GcnArchNameFormat(devProp.gcnArchName, gcnArchNameSubstr);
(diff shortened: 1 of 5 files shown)


wenkaidu:
What's the purpose of this PR? Anything missing from Audrey's https://github.com/ROCmSoftwarePlatform/rccl/pull/886?

src/graph/xml.cc
```diff
@@ -636,15 +636,15 @@ ncclResult_t ncclTopoGetXmlFromGpu(struct ncclXmlNode* pciNode, uint32_t rocmDev
...
 
-  int gcn;
+  const char* gcn;
   NCCLCHECK(xmlGetAttrIndex(gpuNode, "gcn", &index));
   if (index == -1) {
     hipDeviceProp_t devProp;
     CUDACHECK(hipGetDeviceProperties(&devProp, 0));
-    gcn = devProp.gcnArch;
-    NCCLCHECK(xmlSetAttrInt(gpuNode, "gcn", gcn));
+    gcn = devProp.gcnArchName;
```
wenkaidu, line 644: To avoid RCCL build from source getting broken on older ROCm releases, we need to use HIP_NO_GCNARCHNAME to switch between gcnArch and gcnArchName.

nileshnegi: had this in the [previous commit](https://github.com/ROCmSoftwarePlatform/rccl/commit/fea5d1ce3eb5584dc40387874b0454526599d554)... should we still keep it given that gcnArch was deprecated in ROCm 3.7?

src/graph/xml.cc
```diff
@@ -636,15 +636,15 @@ ncclResult_t ncclTopoGetXmlFromGpu(struct ncclXmlNode* pciNode, uint32_t rocmDev
...
+    NCCLCHECK(xmlSetAttr(gpuNode, "gcn", gcn));
   }
-  NCCLCHECK(xmlGetAttrInt(gpuNode, "gcn", &gcn));
+  NCCLCHECK(xmlGetAttrStr(gpuNode, "gcn", &gcn));
```
wenkaidu, line 647: Changing "gcn" from int to str breaks topo xml loading function, I tested on gfx908 and saw segfault.

wenkaidu (changes requested):

wenkaidu (approved):

wenkaidu (approved):

gilbertlee-amd (approved):

PR #923, opened 2023-10-19
Remove LDS based software barriers from MSCCL

wenkaidu:


src/collectives/device/msccl_kernel_impl.h
```diff
@@ -89,10 +89,10 @@ __device__ static void store(U *dst, U val) {
 #endif
 }
 
-inline __device__ static void barrier(int nthreads, uint64_t* barrier_next, uint64_t* barriers) {
+inline __device__ static void barrier(int nthreads) {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-  if (nthreads != WARP_SIZE)
-    barrier_by_group();
+  assert(nthreads == NCCL_MAX_NTHREADS);
+  __asm__ __volatile__("s_waitcnt vmcnt(0) lgkmcnt(0)\ns_barrier");
```
whchung, line 95: instead of using inline assembly, `__syncthreads()` will likely be better as it produces the same instructions but allow compiler options to optimize.

wenkaidu: __syncthreads no longer waits for vmcnt. It caused data corruption in RCCL. Thus RCCL has been forced to use inline asm. I feel it is safer to do the same here

whchung (approved):

PedramAlizadeh (approved):

PR #945, opened 2023-11-02
msccl: add templated kernel

wenkaidu:


src/include/msccl/msccl_struct.h
```diff
@@ -144,6 +144,8 @@ struct mscclAlgo {
   bool inPlace;
   // Whether this algorithm is suitable for out-of-place.
   bool outOfPlace;
+  // Keep a bit mask of used types
+  uint64_t typeMask;
```
edgargabriel, line 148: it looks like with hipcc they typeMask access is unaligned, we could maybe add a 'char[6] alignBuf' or similar. Here is the output of this structure compiled with hipcc:

```
egabriel@smc01:~/tmp$ hipcc bool-size.cc
egabriel@smc01:~/tmp$ ./a.out
Size of bool is 1
Start 0x7ffdeefd8a38 inPlace 0x7ffdeefd8a38 outofPlace 0x7ffdeefd8a39 typeMask 0x7ffdeefd8a40
```
and the corresponding testcode 

```
int main (int argc, char **argv)
{
  printf("Size of bool is %lu\n", sizeof(bool));
  struct mscclAlgo algo;
  printf("Start %p inPlace %p outofPlace %p typeMask %p\n", &algo, &algo.inPlace, &algo.outOfPlace, 
 &algo.typeMask);

  return 0;
}
```

wenkaidu: I will change to uint8_t then. it is sufficient for existing list of ops

edgargabriel (approved):

PR #959, opened 2023-11-08
msccl: enable basic collective trace

wenkaidu:
To avoid increasing number of kernels, colltrace is only enabled with RCCL_MSCCL_FORCE_FULLOPS=1


nusislam (approved):
Good to go when the tests pass.

PR #962, opened 2023-11-09
Optimize MSCCL reduce primitive switching for gfx942

yzygitzh:


src/collectives/device/prims_ll.h
```diff
@@ -773,10 +773,12 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
     return mscclGenericOp<0,1,0,0>(&srcs, 1, &dsts, 1, eltN);
   }
   __device__ void reduce(T** srcs, int nsrcs, T** dsts, int ndsts, int eltN) {
+#if !defined(__gfx942__)
```
wenkaidu: I don't think this is necessary since mscclRunInterpreter is templated. The "else" path of "Proto::Id == NCCL_PROTO_LL" should not have been included by compiler. Did you see any difference with and without this "#if"?

yzygitzh: You're right, I ran some experiments to compare and didn't find difference. Removed this in latest commit.

wenkaidu (approved):

whchung:
jenkins: retest this please.

whchung:
@wenkaidu The CI is stuck at `sles15sp1 && 4gfx908` stage. If it's not serious, suggest merge the PR.

PR #963, opened 2023-11-09
Check to support older ROCm versions

BertanDogancay:


CMakeLists.txt
```diff
@@ -119,6 +119,26 @@ execute_process(
...
+  COMMAND         bash "-c" "cat ${ROCM_PATH}/.info/version"
+  OUTPUT_VARIABLE rocm_version_string
+)
+string(REGEX MATCH "([0-9]+)\\.([0-9]+)\\.([0-9]+)" rocm_version_matches ${rocm_version_string})
+if (rocm_version_matches)
+    set(ROCM_MAJOR_VERSION ${CMAKE_MATCH_1})
+    set(ROCM_MINOR_VERSION ${CMAKE_MATCH_2})
+    set(ROCM_PATCH_VERSION ${CMAKE_MATCH_3})
+
+    message(STATUS "ROCm version: ${ROCM_MAJOR_VERSION}.${ROCM_MINOR_VERSION}.${ROCM_PATCH_VERSION}")
```
nileshnegi, line 133: are you ignoring the digits after `-` like in `6.0.0-12568` for the patch version?

wenkaidu (approved):

CMakeLists.txt
BertanDogancay: yes, I think the digits after '-' are not important in this case

PR #967, opened 2023-11-15
Fix MSCCL work FIFO allocation with HIP graph enabled

yzygitzh:

PR #968, opened 2023-11-15
msccl: allocate scratch as ext-scope fine-grained

wenkaidu:


src/misc/msccl/msccl_setup.cc
```diff
@@ -69,7 +69,7 @@ ncclResult_t mscclSetupScratch(struct mscclAlgo* hostAlgo, hipStream_t stream) {
   size_t sizeNeeded = (status.nBytes * (size_t)(hostAlgo->nScratchChunks)) / (size_t)(hostAlgo->nChunksPerLoop);
   if (sizeNeeded > status.scratchBufferSize){
     NCCLCHECK(ncclCudaFree(status.scratchBuffer));
-    NCCLCHECK(ncclCudaCalloc((char**)&status.scratchBuffer, sizeNeeded));
+    NCCLCHECK(ncclCudaCalloc((char**)&status.scratchBuffer, sizeNeeded, stream, true));
```
whchung: From experiences learned in #967, it's better to use `ncclCudaMalloc` to cope with graph capture, as there are `hipStreamSynchronize` calls inside `ncclCudaCalloc`.

whchung (changes requested):
Need to switch from `ncclCudaCalloc` to `ncclCudaMalloc` to avoid hitting `hipStreamSynchronize` inside graph capturing.

gilbertlee-amd (approved):

nusislam (approved):

src/misc/msccl/msccl_setup.cc
wenkaidu: done

whchung (approved):

PR #974, opened 2023-11-18
Workarounds rid stream sync related HIP APIs in MSCCL initialization logic.

whchung:


src/collectives/msccl.cc
```diff
@@ -28,7 +28,7 @@ ncclResult_t mscclLoadAlgo(const char *mscclAlgoFilePath, mscclAlgoHandle_t *msc
 
   struct mscclAlgo* devAlgo;
   NCCLCHECK(ncclCudaMalloc(&devAlgo, 1));
-  CUDACHECK(hipMemcpy(devAlgo, hostAlgo, sizeof(struct mscclAlgo), hipMemcpyHostToDevice));
+  CUDACHECK(hipMemcpyAsync(devAlgo, hostAlgo, sizeof(struct mscclAlgo), hipMemcpyHostToDevice));
```
wenkaidu, line 31: should HIP stream be passed in here?

whchung: @wenkaidu I’m not familiar with the logic here. I just blindly changed it to allow workloads passing graph capturing stage. Please review and advise better ways of implementation. Thanks.

yzygitzh: @wenkaidu @whchung If passing stream to hipMemcpyAsync, it will be captured into graph and be executed in each graph launch, causing extra overhead.

I think the proper way to solve this is to call hipMemcpyAsync in a side stream and then synchronize it, like in https://github.com/ROCmSoftwarePlatform/rccl/blob/develop/src/include/alloc.h#L184-L218. However, this is currently not allowed in ROCm.

Another temp workaround can be moving algorithm loading from mscclInternalSchedulerSelectAlgo to mscclInternalSchedulerInit, i.e. load all algorithms during mscclInit (also RCCL init). This requires RCCL init is not in graph capturing.

yzygitzh: Created #982 for the temp workaround.

whchung:
Per discussion with @wenkaidu and @yzygitzh , the PR is superseded by #982. Close it hereby. I will leave the `exp_bandaid` branch open as it's still used by some Dockerfile inflight.

+ @dllehr-amd for awareness.

PR #982, opened 2023-11-28
Move MSCCL algorithm loading to initialization to workaround HIP graph conflict

yzygitzh:
…h conflict


whchung:
@yzygitzh , @wenkaidu , from CI logs it seems this PR shares the same failure patterns with #974 . So it could be the switch from `Calloc` to `Malloc` which triggered some butterfly effects.

wenkaidu:
Can we initialize channels based on loaded xmls on init?

yzygitzh:
> Can we initialize channels based on loaded xmls on init?

Refactored the PR to move both number of channel decision and algorithm loading to initialization phase.

wenkaidu (approved):

Issue #958, opened 2023-11-08
RCCL ignores AMDGPU_TARGETS and builds for all hardcoded arches anyway


AngryLoki:
Hi, while updating build script for rccl-5.7.1 in Gentoo, I noticed that RCCL ignores `-DAMDGPU_TARGETS=...` parameter, which makes compilation for a single GPU significantly slower than it could be.

AMDGPU_TARGETS works for every ROCm package I checked (miopen, hipCUB, rocBLAS, hipRAND, rocFFT, and so on), the only difference, why it does not work in rccl is because it uses `FORCE` in `set(AMDGPU_TARGETS` definition:
https://github.com/ROCmSoftwarePlatform/rccl/blob/rocm-5.7.1/CMakeLists.txt#L65

Can you remove this `FORCE` in some future release? Thanks!

nileshnegi:
Fixed in https://github.com/ROCm/rccl/commit/414884c6cbc18f40c880d9b7bed39132e9f349fa

PR #989, opened 2023-11-29
Remove FORCE from AMDGPU_TARGETS

nileshnegi:
Modified CMakeLists.txt to remove FORCE for AMDGPU_TARGETS and check compilation support only for target GPU archs.
Related to #958

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,24 +47,26 @@ include(CheckIncludeFiles)
 include(CheckSymbolExists)
 include(cmake/Dependencies.cmake) # GTest, rocm-cmake, rocm_local_targets
 
-# Determine which GPU architectures to build for
+# Build only for local GPU architecture
 if (BUILD_LOCAL_GPU_TARGET_ONLY)
   message(STATUS "Building only for local GPU target")
   if (COMMAND rocm_local_targets)
     rocm_local_targets(DEFAULT_GPUS)
   else()
-    message(WARNING "Unable to determine local GPU targets.  Falling back to default GPUs")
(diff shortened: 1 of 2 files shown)


wenkaidu (approved):

wenkaidu (approved):

CMakeLists.txt
```diff
@@ -56,15 +56,17 @@ if (BUILD_LOCAL_GPU_TARGET_ONLY)
     message(WARNING "Unable to determine local GPU targets.  Falling back to default GPUs")
   endif()
 endif()
-if(COMMAND rocm_check_target_ids)
+if (NOT DEFINED AMDGPU_TARGETS)
```
lawruble13: This check is redundant, setting a cache variable has no effect if it's already set.

CMakeLists.txt
```diff
@@ -56,15 +56,17 @@ if (BUILD_LOCAL_GPU_TARGET_ONLY)
...
+  set(AMDGPU_TARGETS "${DEFAULT_GPUS}" CACHE STRING "AMD GPU targets to compile for")
```
lawruble13: Note `AMDGPU_TARGETS` is also deprecated, so we probably shouldn't be using it as an intermediate. If we must still have it, I would prefer to have it behave as it does on the other libraries (see e.g. [rocRAND](https://github.com/ROCmSoftwarePlatform/rocRAND/blob/126731e25586054e5c50cb5012577dd7d51249c0/CMakeLists.txt#L91)).

lawruble13 (approved):

PR #995, opened 2023-12-01
Fix mscclAlgoHandle not initialized issue

yzygitzh:


whchung (approved):
@yzygitzh Thanks for spotting this bug!

wenkaidu (approved):

PR #997, opened 2023-12-01
Relax workgroup barrier implementation for MSCCL send/recv ops.

whchung:
Introduce MSCCL template parameter to send/recv protocol primitives.
Alter the implementation for LL protocol so under MSCCL mode, do not issue `s_waitcnt`.

The core idea is that in MSCCL kernels each workgroup / wavefront is basically reading data from disjoint places from HBM, and write to remote GPUs. We just need to ensure all workitems reach the same spot, and we don't care if the XGMI traffic is completed. It's the job of the receiving end which be handled by the read loop in LL protocol to check the flag value.

Only MSCCL + LL protocol would be affected, other combinations (MSCCL + Simple, or RCCL) won't be affected.

Empirical testing show 0.4~0.5us improvement for small message sizes where MSCCL + LL protocol is in use, with correct results.

### DIFF
--- a/src/collectives/device/prims_ll.h
+++ b/src/collectives/device/prims_ll.h
@@ -82,8 +82,7 @@ private:
 
   inline __device__ void barrier() {
 #if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
-    if (nthreads != WARP_SIZE)
-      barrier_by_group();
+    barrier_by_group();
 #else
     asm volatile ("bar.sync %1, %0;" :: "r"(nthreads), "r"(15-group));
 #endif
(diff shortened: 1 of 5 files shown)


src/collectives/device/prims_ll.h
```diff
@@ -89,6 +89,13 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
 #endif
   }
 
+  inline __device__ void msccl_barrier() {
+#if defined(__HIP_PLATFORM_HCC__) || defined(__HCC__) || defined(__HIPCC__)
+    __asm__ __volatile__("s_barrier");
```
wenkaidu: __builtin_amdgcn_s_barrier()

wenkaidu (approved):

PR #1009, opened 2023-12-07
Tune MSCCL all-reduce algorithm

yzygitzh:
Extend to 64 thread blocks for some sizes for better bandwidth utilization. Up to 25% performance improvement observed.


whchung (approved):

whchung:
@wenkaidu I have verified these new XMLs indeed significantly improve performance in some of recent message sizes we care about. As there is no change in the source codes, I plan to merge it once CI passes. Is it okay with you?

nileshnegi (approved):

PR #1010, opened 2023-12-08
Improve MSCCL scratch allocation

yzygitzh:
1) Fix not handling zero-sized scratch buffer allocation request, causing warnings in alloc.h.
2) Improve allocation logic to allocate only 2^n sizes and reuse same buffer for sizes in the same 2^n range.


whchung (approved):
LGTM. Thanks for improving the scratch buffer allocation logic!

wenkaidu (approved):

PR #1013, opened 2023-12-09
[Experimental] Adding experimental XCC affinity mode

gilbertlee-amd:
Adding opt-in experimental XCC affinity mode for MI300
This should be considered experimental and not guaranteed to work in current/future versions of ROCm.


src/init.cc
wenkaidu, line 1: Using cudaDev will not work when only one HIP device is visible for the running process. We should add a PCIe busID mapping table, then looking up src/dst ranks from PCIe busIDs.

src/include/rccl_vars.h
```diff
@@ -25,6 +25,37 @@ THE SOFTWARE.
 
 #include "param.h"
 
-RCCL_PARAM_DECLARE(EnableHipGraph);  // Opt-in environment variable for enabling hipGraph
+RCCL_PARAM_DECLARE(EnableHipGraph);      // Opt-in environment variable for enabling hipGraph
+RCCL_PARAM_DECLARE(ExperimentalXccMode); // Opt-in environment variable for enabling experimental XCC filtering
+
+int32_t __inline__ rcclGetPreferredXcc(int src, int dst)
```
wenkaidu: The table should be based on src and dst ranks which are unique and not impacted by HIP_VISIBLE_DEVICES and ROCR_VISIBLE_DEVICES

src/include/rccl_vars.h
```diff
@@ -25,6 +25,37 @@ THE SOFTWARE.
...
+
+  constexpr int32_t table2[8][8] = {{0,1,4,6,2,7,3,5},
+                                    {1,0,4,6,2,5,3,7},
+                                    {4,6,0,1,5,2,7,3},
+                                    {6,7,1,0,4,2,5,3},
+                                    {2,3,4,5,0,6,1,7},
+                                    {6,4,2,3,7,0,5,1},
+                                    {2,3,6,4,1,7,0,5},
+                                    {4,6,2,3,5,1,7,0}};
+
```
wenkaidu: Need to add PCIe mapping to ranks, which can be used to looked actual rank IDs.

src/init.cc
```diff
@@ -1052,11 +1053,26 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   NCCLCHECKGOTO(fillInfo(comm, comm->peerInfo+rank, comm->commHash), ret, fail);
   NCCLCHECKGOTO(bootstrapAllGather(comm->bootstrap, comm->peerInfo, sizeof(struct ncclPeerInfo)), ret, fail);
 
+  comm->busIdIdx = 0;
   for (int i = 0; i < nranks; i++) {
-    if ((i != rank) && (comm->peerInfo[i].hostHash == comm->peerInfo[rank].hostHash) && (comm->peerInfo[i].busId == comm->peerInfo[rank].busId)) {
-      WARN("Duplicate GPU detected : rank %d and rank %d both on CUDA device %lx", rank, i, comm->peerInfo[rank].busId);
-      ret = ncclInvalidUsage;
-      goto fail;
+    if ((i != rank) && (comm->peerInfo[i].hostHash == comm->peerInfo[rank].hostHash)) {
```
wenkaidu, line 1058: correct place to calculate target GPUs should be at end of ncclTransportP2pSetup, because connections are established only during 1st collective call for sendrecv/alltoall type of collectives. comm->connectSend is the data structure to find target GPUs info. It has multiple entries. We should only need to care about index 0 (all reduce type) and index 1 (sendrecv/alltoall type)

PR #1015, opened 2023-12-10
msccl: fix data corruption with MTYPE_RW

wenkaidu:


src/collectives/device/msccl_kernel_impl.h
```diff
@@ -460,7 +460,7 @@ __device__ __forceinline__ void mscclRunInterpreter(
           return;
       }
       if (t->hasDependence && tid == nthreads-1)
-        __atomic_store_n(&mscclFlags[bid].flag, (uint64_t) COMPUTE_FLAG(workIndex, iter, step), __ATOMIC_RELAXED);
+        __atomic_store_n(&mscclFlags[bid].flag, (uint64_t) COMPUTE_FLAG(workIndex, iter, step), t->type == MSCCL_REDUCE ? __ATOMIC_RELEASE : __ATOMIC_RELAXED);
```
nusislam, line 463: Why is the scope different based on the operation type?

wenkaidu: Latest FW/KFD enabled MTYPE_RW coarse grained memory (which was MTYPE_CC previously). During MSCCL 2-step all pair all reduce, "reduce" operator running on some CUs writes data to coarse grained output buffer, then other CUs send data from the output buffer to remote. This requires output buffer memory coherence after reduce step.
RCCL doesn't have this kind of use case, thus is not affected.
@yzygitzh maybe able to update XML to avoid any send from output buffer. In that case, we will be able to remove this fence.

akolliasAMD (approved):

src/collectives/device/msccl_kernel_impl.h
nusislam: @yzygitzh - Do you have any comments?

yzygitzh: I made a quick try on the XMLs mentioned by @wenkaidu , but seems the performance will be worse than original XMLs plus this patch.

PR #1017, opened 2023-12-12
Fully disable MSCCL when machine is not matched

yzygitzh:


wenkaidu (approved):

src/init.cc
akolliasAMD, line 1: It appears that the changes on this file breaks npkit. Seg Fault appears when gathering npkit trace using all reduce, at the end. We will investigate a bit more in the morning. (The issue appeared  after the second commit)

yzygitzh: Thanks for the fault location info. It seems to be using freed comm structure issue, I pushed a new commit trying to fix this.

akolliasAMD (approved):

PR #1019, opened 2023-12-15
Rework barriers and adjust scope of atomics

wenkaidu:


src/collectives/device/prims_ll.h
```diff
@@ -125,15 +117,11 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
       __asm__ __volatile__("s_wakeup");
       if (sendConnFifoPtr) {
         int size = ((sendConnHead & NCCL_LL_CLEAN_MASK) == NCCL_LL_CLEAN_MASK) ? stepLines*sizeof(union ncclLLFifoLine) : nbytes;
-        __atomic_store_n(sendConnFifoPtr+sendConnHead%NCCL_STEPS, (size), __ATOMIC_SEQ_CST);
+        __atomic_store_n(sendConnFifoPtr+sendConnHead%NCCL_STEPS, (size), __ATOMIC_RELAXED);
```
nusislam, line 120: Why is it OK to do an atomic store with `RELAXED` here?

wenkaidu: this store tells proxy thread about upcoming data packet before GPU shader starts writing any data. So it should be fine without cache flush.

nusislam (approved):

PR #1045, opened 2024-01-15
Adding regression test

AtlantaPepsi:
Does an AllReduce of 5KB data between 2 devices under 1 process, blocks regression that makes execution time greater than 5ms.

### DIFF
--- a/test/StandaloneTests.cpp
+++ b/test/StandaloneTests.cpp
@@ -143,8 +143,10 @@ namespace RcclUnitTesting {
 
   TEST(Standalone, RegressionTiming)
   {
+    // timing
     using namespace std::chrono;
     using Clock = std::chrono::high_resolution_clock;
+    int usElapsed, iter = 20;
 
     // Check for 2 GPUs
     int numGpus;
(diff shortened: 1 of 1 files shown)


wenkaidu:
1. need to go through all protocols.
2. need to add a few warm up cycles, then take timing measurements from average of 10-20 iterations.

test/StandaloneTests.cpp
```diff
@@ -140,4 +140,105 @@ namespace RcclUnitTesting {
...
+        for (int rank = 0; rank < numRanks; rank++) {
+          HIPCALL(hipMemcpy(cpuOutput.data(), gpuOutput[rank], N * sizeof(int), hipMemcpyDeviceToHost));
+          HIPCALL(hipDeviceSynchronize());
+          for (int i = 0; i < N; i++)
+            ASSERT_EQ(cpuOutput[i], cpuExpected[i]);
+        }
+      }
+
+      EXPECT_LT(msElapsed/15.0, 5);
+      printf("sada%d\n", msElapsed);
```
wenkaidu: Logs always show "0". Can you make msElapsed as microseconds and float? We would like to see some numbers which varies depends on protocol to make sure it works as expected.

wenkaidu (approved):

test/StandaloneTests.cpp
```diff
@@ -140,4 +140,105 @@ namespace RcclUnitTesting {
...
+        std::vector<int> cpuOutput(N);
+        for (int rank = 0; rank < numRanks; rank++) {
+          HIPCALL(hipMemcpy(cpuOutput.data(), gpuOutput[rank], N * sizeof(int), hipMemcpyDeviceToHost));
+          HIPCALL(hipDeviceSynchronize());
+          for (int i = 0; i < N; i++)
+            ASSERT_EQ(cpuOutput[i], cpuExpected[i]);
+        }
+      }
+
+      EXPECT_LT(usElapsed/15.0, 5000);
```
gilbertlee-amd: iter isn't hard coded, yet this is.  Should pick one or the other.

gilbertlee-amd (changes requested):

gilbertlee-amd (approved):

PR #1046, opened 2024-01-15
Add option to force enable network transport on single node

wenkaidu:


nusislam (approved):
Good to go when the tests pass.

PR #1062, opened 2024-01-25
msccl: ensure memory coherence after data receive

wenkaidu:


src/collectives/device/msccl_kernel_impl.h
```diff
@@ -372,7 +372,7 @@ __device__ __forceinline__ void mscclRunInterpreter(
           return;
       }
       if (t->hasDependence && tid == nthreads-1)
-        __atomic_store_n(&mscclFlags[bid].flag, (uint64_t) COMPUTE_FLAG(workIndex, iter, step), t->type == MSCCL_REDUCE ? __ATOMIC_RELEASE : __ATOMIC_RELAXED);
+        __atomic_store_n(&mscclFlags[bid].flag, (uint64_t) COMPUTE_FLAG(workIndex, iter, step), (t->type == MSCCL_REDUCE || t->type == MSCCL_RECV) ? __ATOMIC_RELEASE : __ATOMIC_RELAXED);
```
nusislam, line 375: Do `recv` operations have any dependence for any collectives?

wenkaidu: "recv" can be receive from remote GPU send or RDMA send. It appears that RDMA path needs memory coherence atomic. However, the 2 paths can not be easily distinguished within MSCCL kernel. Thus I added atomic to both.

nusislam: That's alright. But my concern is this `if` statement is executed only when the operation depends on another operation (`t->hasDependence`). Do `recv` operations have dependence on any other operations on single/multi node?

nusislam (approved):

PR #1077, opened 2024-02-09
2.18.5 fix

wenkaidu:


PedramAlizadeh (approved):
Tested on the multi-node machine and it fixes the perf regression for multi-node Alltoall and Alltoallv

PR #1093, opened 2024-02-23
Adding FP16 to all unit test cases

AtlantaPepsi:
All unit tests runtime combined increased by about 10%


wenkaidu (approved):

gilbertlee-amd (changes requested):
This looks like too much overlap in coverage.
For example, it shouldn't be required to have both ncclFloat16 for both inplace and outofplace.

gilbertlee-amd (approved):

PR #1096, opened 2024-02-29
Add support for using contiguous memory for GPU direct RDMA

wenkaidu:
Enabled by env var RCCL_NET_CONTIGUOUS_MEM=1

### DIFF
--- a/src/clique/CliqueManager.cc
+++ b/src/clique/CliqueManager.cc
@@ -190,7 +190,7 @@ ncclResult_t CliqueManager::Init(ncclUniqueId const* commId, int suffix)
       hipIpcMemHandle_t handle;
       // Allocate fine-grained device memory on rank 0 and get IPC handle for it
       // Re-usable barrier consists of (globalCount / globalSense) pair of integers
-      NCCLCHECKGOTO(ncclCudaCalloc(&m_fineGrainBarrierMem, NCCL_MAX_OPS * 2 * sizeof(int), nullptr, true), res, dropback);
+      NCCLCHECKGOTO(ncclCudaCalloc(&m_fineGrainBarrierMem, NCCL_MAX_OPS * 2 * sizeof(int), nullptr, hipDeviceMallocFinegrained), res, dropback);
       if (hipIpcGetMemHandle(&handle, m_fineGrainBarrierMem) != hipSuccess)
       {
         WARN("Unable to get IPC handle for barrier memory");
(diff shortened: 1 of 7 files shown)


src/transport/net.cc
```diff
@@ -521,7 +525,18 @@ static ncclResult_t sharedNetBuffersInit(struct ncclProxyState* proxyState, int
     if (sameProcess == 0 || ncclCuMemEnable()) {
       NCCLCHECK(ncclP2pAllocateShareableBuffer(state->size, &state->ipcDesc, (void**)&state->cudaBuff));
     } else {
-      NCCLCHECK(ncclCudaCalloc(&state->cudaBuff, state->size, nullptr, cuda));
+#if defined(HIP_UNCACHED_MEMORY)
+#if defined(HIP_CONTIGUOUS_MEMORY)
+      NCCLCHECK(ncclCudaCalloc(&state->cudaBuff, state->size, nullptr,
+        cuda ? (rcclParamNetContiguousMem() ? hipDeviceMallocContiguous : hipDeviceMallocUncached) : hipDeviceMallocDefault));
```
nusislam, line 531: What is the significance of `hipDeviceMallocContiguous`? How is it different than the existing memory type support in RCCL.

src/transport/p2p.cc
```diff
@@ -218,7 +218,11 @@ ncclResult_t ncclP2pAllocateShareableBuffer(size_t size, ncclIpcDesc *ipcDesc, v
 #endif
   } else {
     // Allocate a CUDA buffer and generate an IPC handle for it
-    NCCLCHECK(ncclCudaCalloc((char **)ptr, size, nullptr, true));
+#if defined(HIP_UNCACHED_MEMORY)
+    NCCLCHECK(ncclCudaCalloc((char **)ptr, size, nullptr, hipDeviceMallocUncached));
+#else
+    NCCLCHECK(ncclCudaCalloc((char **)ptr, size, nullptr, hipDeviceMallocFinegrained));
```
nusislam, line 224: How is this `Finegrained` different than `Uncached`? Are these buffer type classifications true for all supported AMD GPUs?

src/clique/CliqueManager.cc
```diff
@@ -190,7 +190,7 @@ ncclResult_t CliqueManager::Init(ncclUniqueId const* commId, int suffix)
       hipIpcMemHandle_t handle;
       // Allocate fine-grained device memory on rank 0 and get IPC handle for it
       // Re-usable barrier consists of (globalCount / globalSense) pair of integers
-      NCCLCHECKGOTO(ncclCudaCalloc(&m_fineGrainBarrierMem, NCCL_MAX_OPS * 2 * sizeof(int), nullptr, true), res, dropback);
+      NCCLCHECKGOTO(ncclCudaCalloc(&m_fineGrainBarrierMem, NCCL_MAX_OPS * 2 * sizeof(int), nullptr, hipDeviceMallocFinegrained), res, dropback);
```
nusislam, line 193: Some of these changes are not related to `contiguous` memory type.

src/transport/net.cc
wenkaidu: hipDeviceMallocContiguous should be same as current hipDeviceMallocUncached with added property of being contiguous in physical address

src/transport/p2p.cc
wenkaidu: hipDeviceMallocUncached is identical to hipDeviceMallocFinegrained on MI100 and MI200. On MI300, hipDeviceMallocFinegrained is identical to hipDeviceMallocDefault.

src/clique/CliqueManager.cc
wenkaidu: "clique" is not currently enabled. I made the change here just to make sure API call is still correct in case we want to reenable in future.

nusislam (approved):

PR #1113, opened 2024-03-14
Num of channels 942 fix

akolliasAMD:


src/graph/connect.cc
```diff
@@ -589,6 +589,12 @@ ncclResult_t ncclTopoPostset(struct ncclComm* comm, int* firstRanks, int* treePa
 
   int minNchannels = ncclMinNchannels();
 
+  if (IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx942")) {
+    int managed = 0;
```
gilbertlee-amd: The reasoning behind this logic needs to be properly documented. 

Additionally, are there any guarantees that hipDeviceGetAttribute will only ever return 1, and not some number greater than 1?  Otherwise it's best to modify this code so that it just checks against non-zero, instead of using it directly.

gilbertlee-amd (changes requested):

PR #1128, opened 2024-03-28
[BUILD] Enable RCCL build with amdclang++

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Enable the use of `amdclang++` instead of `hipcc` for building RCCL.

- Update `CXX` and `C` compiler and flags in `toolchain-linux.cmake`
- Update `CMakeLists.txt` for `amdclang++`-related checks
- Replace `hipcc --version` with `hipconfig -v` in CMakeLists.txt to query the HIP version, as `amdclang++ --version` does not report HIP version
- Use `set(CMAKE_REQUIRED_LIBRARIES hip::device)` and `unset(CMAKE_REQUIRED_LIBRARIES)` around any `check_symbol_exists` calls for hipDevice symbols.

## Approval Checklist


AtlantaPepsi (approved):

CMakeLists.txt
```diff
@@ -115,12 +123,21 @@ message(STATUS "HIP runtime:      ${HIP_RUNTIME}")
...
+    OUTPUT_VARIABLE hipcc_version_string)
+  message(STATUS "hipcc version:    ${hipcc_version_string}")
+elseif("${CMAKE_CXX_COMPILER}" MATCHES ".*amdclang")
+  find_program(hipcc_executable amdclang)
+  message(STATUS "amdclang executable: ${hipcc_executable}")
+  execute_process(
+    COMMAND         bash "-c" "${hipcc_executable} --version | grep 'AMD clang version' | awk -F\" \" '{ printf $4}'"
+    OUTPUT_VARIABLE hipcc_version_string)
+  message(STATUS "amdclang version:    ${hipcc_version_string}")
+endif()
```
corey-derochie-amd: Not a functional change, but it would aid comprehension if we renamed all the `hipcc_` variables to something more generic.

wenkaidu: Did we add CI testing for this option?

BertanDogancay: We might want to reduce the amount of duplicate code here. The first if/else statement should be enough to gather info to generalize the second if/else block.

nileshnegi: > Did we add CI testing for this option?

no... waiting on info about ROCm version which enforces switch from hipcc to amdclang++

nileshnegi: > We might want to reduce the amount of duplicate code here. The first if/else statement should be enough to gather info to generalize the second if/else block.

may add an else block for exit if the compiler is neither hipcc nor amdclang

BertanDogancay: > > We might want to reduce the amount of duplicate code here. The first if/else statement should be enough to gather info to generalize the second if/else block.
> 
> may add an else block for exit if the compiler is neither hipcc nor amdclang

I don't think you need an if/else block here at all. Once you have the required info from the first if/else statement in this PR, this one can just be generalized. When cmake tries to find a package and unsuccessful, it will fail and exit anyway.

CMakeLists.txt
```diff
@@ -576,7 +582,7 @@ endif()
 if(${HIP_CONTIGUOUS_MEMORY})
   target_compile_definitions(rccl PRIVATE HIP_CONTIGUOUS_MEMORY)
 endif()
-if(${hipcc_version_string} VERSION_GREATER_EQUAL "5.7.31920")
+if(${compiler_version_string} VERSION_GREATER_EQUAL "5.7.31920")
```
nileshnegi: @lawruble13 this one seems to be related to hipcc version... is there an equivalent for amdclang version?

CMakeLists.txt
```diff
@@ -606,7 +612,7 @@ if(DEMANGLE_DIR)
   target_compile_definitions(rccl PRIVATE "HAVE_CPLUS_DEMANGLE=1")
   target_compile_definitions(rccl PRIVATE "HAVE_DECL_BASENAME=1")
 endif()
-if(${hipcc_version_string} VERSION_GREATER_EQUAL "6.1.33591")
+if(${compiler_version_string} VERSION_GREATER_EQUAL "6.1.33591")
```
nileshnegi: @lawruble13 this one seems to be related to hipcc version... is there an equivalent for amdclang version?

CMakeLists.txt
nileshnegi: reduced if/else blocks

PR #1130, opened 2024-04-03
Update Depends with correct HIP Runtime package name

arvindcheru:
Proposed Changes:

- hip-rocclr package is deprecated
- This is replaced with latest package hip-runtime-amd.
- This change is to update the package depends of RCCL to the latest package name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -718,7 +718,7 @@ if(BUILD_ADDRESS_SANITIZER)
 else()
   set(DEPENDS_HIP_RUNTIME "hip-runtime-amd" )
 endif()
-rocm_package_add_dependencies(DEPENDS "${DEPENDS_HIP_RUNTIME} >= 3.5.0" "rocm-smi-lib >= 4.0.0")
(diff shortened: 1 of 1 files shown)


nileshnegi:
`HIP_RUNTIME` reported by `find_package(hip REQUIRED)` [here](https://github.com/ROCm/rccl/blob/develop/CMakeLists.txt#L114), shows up as `rocclr` during the cmake stage. Should this be updated as well?

arvindcheru:
> `HIP_RUNTIME` reported by `find_package(hip REQUIRED)` [here](https://github.com/ROCm/rccl/blob/develop/CMakeLists.txt#L114), shows up as `rocclr` during the cmake stage. Should this be updated as well?

Thank you for sharing the inputs, we are checking this separately and the scope of this change may make this to be considered and handled separately.

The changes proposed with this PR, at the moment limiting scope to the depends string update of RCCL package. But are we good in suggesting to remove this print msg (for HIP_RUNTIME) also, or it is a true statement to say this flag is not being used internally in RCCL other than the print and can be removed?

wenkaidu (approved):

PR #1132, opened 2024-04-03
rccl_prim_test: increase max number of workgroups and test iterations

wenkaidu:


PedramAlizadeh (approved):

corey-derochie-amd:
I was wondering, in what context / environment is this typically run? Would it be possible or appropriate to make these values parameters rather than committing hard-coded values?

wenkaidu:
> I was wondering, in what context / environment is this typically run? Would it be possible or appropriate to make these values parameters rather than committing hard-coded values?

rccl_prim_test is not actively maintained. I normally only does minimal fix to get it running for specific testing.

PR #1133, opened 2024-04-04
Fix buffer overflow when parsing kernel cmdline

wenkaidu:


src/init.cc
```diff
@@ -127,7 +127,7 @@ static ncclResult_t ncclInit() {
         FILE* file;
         if ((file = fopen("/proc/cmdline", "r")) != NULL) {
           if (feof(file) == 0 && ferror(file) == 0) {
-            int len = fread(strValue, 1, 1024, file);
+            int len = fread(strValue, 1, 2047, file);
```
edgargabriel, line 130: is reading 2047 (instead of 2048) bytes on purpose?

wenkaidu: yes, last byte is reserved for NULL character if valid data reaches 2047.

edgargabriel (approved):

gilbertlee-amd (approved):

PR #1140, opened 2024-04-11
Rail optimization for rings

gilbertlee-amd:
## Details

**What were the changes?**  
Modified ring generation strategy to be friendlier to rail-optimized network topologies

Existing RCCL rings incur extra traffic on rail-optimized network topologies, and this change should eliminate / remove most of the extra traffic.

Tried to modify as little code as possible by adding on to the existing model detection code.  Three main changes are made: the determination of nodes (via hostname), and then alternating the direction of the lines used for rings across nodes.
For example, Node 0 might normally have a line that proceeds NIC 0->0->1->2->3->NIC 3 which would normally be connected to NIC 0->0->1->2->3->NIC 3 on Node 1.  Instead, Node 0 will use NIC 0->0->1->2->3->NIC 3 but Node 1 will use NIC 3->3->2->1->0->NIC 0 which means that the NIC to NIC connections will stay on the same rail.  This should not have any impact on traffic patterns on non rail-optimized network topologies.  Lastly, additional support for odd-number of nodes is performed. 

**Additional Documentation:**  
Experiments have been done using topology explorer tool.
SWDEV-454102


wenkaidu:
I think this commit worth being mentioned in CHANGELOG?
Also can you add an entry in topo_expl with odd number of nodes? For example, {5, "topo_8p_940.xml",        "5 nodes gfx940 8P"},

CHANGELOG.md
```diff
@@ -2,15 +2,42 @@
 
 Full documentation for RCCL is available at [https://rccl.readthedocs.io](https://rccl.readthedocs.io)
 
-## Unreleased
+## Unreleased - RCCL 2.18.6 for ROCm 6.1.0
+### Changed
+- Compatibility with NCCL 2.18.6
+- Modifying rings to be rail-optimized topology friendly
```
wenkaidu: is this going into 6.1 or 6.2? Must be 6.2 as RCCL API version is 2.19.4 at current develop

corey-derochie-amd: Correct, this is a race condition on the fact that we're waiting for 6.1 to be cut before updating the CHANGELOG in `develop`. The change should go in a new "Unreleased" section.

tools/topo_expl/topo_expl.cpp
```diff
@@ -158,6 +158,7 @@ NodeModelDesc model_descs[] = {
   {2, "topo_8p_940vm.xml",      "2 nodes gfx940 VM"},
   {2, "topo_8p_940_16n.xml",    "2 nodes gfx940 16 NICs"},
   {2, "topo_8p1h_6.xml",        "2 nodes 8P1H Alt."},
+  {5, "topo_8p_940.xml",        "5 nodes gfx940 8P"},
```
wenkaidu, line 161: Please increase https://github.com/ROCm/rccl/blob/develop/tools/scripts/topo_val.sh#L24 by one, so the script will test latest addition.

wenkaidu (approved):

PR #1145, opened 2024-04-15
fixing the issue of failure due to the missed XML files

mberenjk:
Fixing the failure of msccl due to missing the xml files. if the files are missing, it will switch to rccl mode.

### DIFF
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -39,13 +39,41 @@ static const char* mscclUnitTestAlgoShareDirPath = "../share/rccl/msccl-unit-tes
 
 
 bool checkXMLFolderExists() {
-  DIR *dp = nullptr;
-  dp = opendir(mscclAlgoDefaultDir);
-  if (dp == nullptr) {
-      WARN("MSCCL XML files are not found, MSCCL failed, chaging the mode to RCCL!");
-      return false;
+  const char* mscclAlgoDir = getenv(mscclAlgoDirEnv);
+  const char* mscclAlgoShareDir = nullptr;
+  std::string mscclAlgoDirStr;
+  std::string mscclAlgoShareDirStr;
+  const char *fullDirPath = nullptr;
+  if (mscclAlgoDir == nullptr)
+  {
+    // Try to find default algorithm directory based on librccl.so path
+    Dl_info dl_info;
+    struct link_map *link_map_ptr = nullptr;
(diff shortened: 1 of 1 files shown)


src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -28,9 +28,29 @@ static const char* mscclAlgoFilePathEnv = "MSCCL_ALGO_FILE_PATH";
...
+static const char* mscclAlgoShareDirPath = "../share/rccl/msccl-algorithms";
+static const char* mscclUnitTestAlgoShareDirPath = "../share/rccl/msccl-unit-test-algorithms";
+
+
+bool checkXMLFolderExists() {
+  DIR *dp = nullptr;
+  dp = opendir(mscclAlgoDefaultDir);
+  if (dp == nullptr) {
+      WARN("MSCCL XML files are not found, MSCCL failed, chaging the mode to RCCL!");
+      return false;
```
corey-derochie-amd: Is it sufficient to check for the existence of the folder, or are there files we should check for as well?

wenkaidu: Here is the place where all attempts of loading xml failed. Currently this results in fatal error in RCCL init. I guess the goal is to make it non-fatal instead? But why this is better idea and what issue this tries to solve?

wenkaidu: https://github.com/ROCm/rccl/blob/develop/src/misc/msccl/msccl_lifecycle.cc#L133

mberenjk: @corey-derochie-amd when there are no xml files in the folder, it still continues to rccl, there is no crash happened.
@wenkaidu This PR tries to avoid the fatal error and switch to rccl when hitting no msccl-algorithm folder, the folder can be either in a share folder in lib directory, or near librccl in build directory. if both fails, in the initial check, it will return failure and switch to rccl. Does this make sense?

wenkaidu:
Since MSCCL supports multi-node as well, there can be cases that xml files are missing on one or more nodes, but not all nodes. There needs to be handling for this situation as well. In general, I don't think silent failure on missing MSCCL xml is good idea: it can be frustrating to user who debug performance issue then just finding out due to missing xml. I would rather let app fail on any installation issues upfront.

mberenjk:
This PR needs to be closed since we decided to report failure if msccl-algorithm does is missed.

PR #1155, opened 2024-04-24
SWDEV-455705: Fix an UB that could lead to miscompilation

shiltian:


danpetreamd:
LGTM.

danpetreamd (approved):

src/device/msccl_kernel_impl.h
```diff
@@ -130,7 +130,7 @@ __device__ __forceinline__ void mscclRunInterpreter(
   // initialize ncclShmem and mscclShmem.work
   int channelId = mscclShmem.mscclTB.channelId;
   {
-    void *dst, *src;
+    void *dst = nullptr, *src = nullptr;
```
wenkaidu: why this is necessary? "bytes" set to 0 for all threads. dst/src should be ignored without valid "bytes"

wenkaidu (changes requested):

src/device/msccl_kernel_impl.h
shiltian: Because use of uninitialized variable is UB and will lead to miscompilation when optimization is on. The use is at the call site instead of callee, even though in the callee the variables might not be actually used.

danpetreamd: looks like bytes is only initialized to `0`.
the `switch...case` below sets it to various values.

wenkaidu: this code has been working for months. why it is failing now? there are many many uninitialized variables in the code. Are these safe or all need to be changed?

danpetreamd: I'm not sure I understand your question.
If you mean to say that checking `bytes!=0` should be enough, yeah, I agree with that.
But IMHO I like being explicit about `dst` and `src` as well, just in case the assumption that `bytes!=0` is enough changes in the future.

shiltian: I'd say it just happens to be working for months. The compiler's behavior towards UB can easily change since it is "undefined", especially when the optimization is on, so one should not rely on UB for correctness.

IMHO I'd recommend to fix all UBs to avoid such issues in the future.

wenkaidu: RCCL has similar "UB", but there is additional check on bytes: https://github.com/ROCm/rccl/blob/develop/src/device/common.h#L294
Will RCCL kernel have same issue or that bytes check helped?

wenkaidu: BTW, the same logic works for NVIDIA compiler: https://github.com/NVIDIA/nccl/blob/master/src/device/common.h#L150
The requirement for initializing all stack variables is kind of unacceptable.

shiltian: It doesn't mean it is right. Again, it just "happens" to work.

https://stackoverflow.com/questions/39681349/is-passing-an-uninitialized-variable-to-another-function-ub

wenkaidu: question: it seems RCCL kernel doesn't have this issue. Is it because:
https://github.com/ROCm/rccl/blob/develop/src/device/common.h#L291
or,
https://github.com/ROCm/rccl/blob/develop/src/device/common.h#L294?
If we do same in MSCCL kernel, will it solve the issue?

wenkaidu: > It doesn't mean it is right. Again, it just "happens" to work.
> 
> https://stackoverflow.com/questions/39681349/is-passing-an-uninitialized-variable-to-another-function-ub

Thanks for the link. I think "if (bytes) copyToShmem8(tid%WARP_SIZE, dst, src, bytes);" should fix the issue of passing-an-uninitialized-variable-to-another-function?

shiltian: That should work. Do you want me to do that?

wenkaidu: > That should work. Do you want me to do that?

sure. that's better as RCCL/NCCL does same.

wenkaidu (approved):

PR #1178, opened 2024-05-14
Make WSL detection thread safe

wenkaidu:
## Details

**What were the changes?**  
Make WSL detection thread safe

Make WSL detection thread safe

Use thread local variable

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/misc/rocm_smi_wrap.cc
+++ b/src/misc/rocm_smi_wrap.cc
@@ -23,7 +23,7 @@ THE SOFTWARE.
 #include "core.h"
 #include "utils.h"
 
-__thread static int is_wsl2 = -1;
+static __thread int is_wsl2 = -1;
 
 #define ROCMSMICHECK(cmd) do {               \
   rsmi_status_t ret = cmd;                   \
--- a/src/misc/rocm_smi_wrap.cc
+++ b/src/misc/rocm_smi_wrap.cc
@@ -23,7 +23,7 @@ THE SOFTWARE.
 #include "core.h"
 #include "utils.h"
 
-static int is_wsl2 = -1;
+__thread static int is_wsl2 = -1;
 
 #define ROCMSMICHECK(cmd) do {               \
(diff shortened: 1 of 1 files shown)


src/misc/rocm_smi_wrap.cc
```diff
@@ -23,7 +23,7 @@ THE SOFTWARE.
 #include "core.h"
 #include "utils.h"
 
-static int is_wsl2 = -1;
+__thread static int is_wsl2 = -1;
```
nileshnegi: should it be `static __thread int is_wsl2 = -1;`? as per [https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html](https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html)
or the order does not matter?

wenkaidu: I am not language lawyer. But I think you may be right. I will change the order

edgargabriel: the other question is whether to use __thread vs. thread_local, the C++ construct for that.

wenkaidu: all other variables are defined using __thread in rccl/nccl. so better keep the consistency, I think

edgargabriel: then I will have to change my code :-)

edgargabriel: actually, it doesn't matter for this pr, but for mine it requires `thread_local`, since it is on a `std::vector` variable. I cannot compile it if I change to __thread, it gives me an error that it has a 'non-trivial destructor'

edgargabriel (approved):

src/misc/rocm_smi_wrap.cc
wenkaidu: that should be fine as well

nileshnegi (approved):

akolliasAMD (approved):

PR #1184, opened 2024-05-22
add 256 channels support

nusislam:
## Details
By default RCCL can support 64 channels. This PR extends the number of supported channels to 256.

**Work item:** [_"Internal", or link to GitHub issue (if applicable)._](https://ontrack-internal.amd.com/browse/SWDEV-462962)

**What were the changes?**  
Major change is in the data structures keeping track of channels.

Some workloads can benefit from increased number of channels on certain platforms.

By default, channelMask was a 64-bit integer in RCCL. To support increased number of channels, this PR modified the channelMask to be a `struct` with multiple 64-bit integers. The PR has modifications for all the codes that need to use channelMask for connection and scheduling.

**Additional Documentation:**  
sendrecv operations are limited to 64 channels due to a bug (will be investigated later)
multi-node is also limited to 64 channels (will be worked on later)
all non-gfx94 GPUs are limited to 32 channels (same as/inherited from current RCCL) 

## Approval Checklist


wenkaidu:
1. Previously when we increased max channels to 64, it caused performance regression on MI100 and MI200. Thus below code was added to address this issue: https://github.com/ROCm/rccl/blob/develop/src/graph/connect.cc#L293

src/device/common.h
```diff
@@ -226,22 +227,31 @@ static __forceinline__ __device__ void ncclRedopPtrDeref(struct ncclWorkElem* we
...
-      if (channelMask & (1ull<<x)) {
-        int y = __popcll(channelMask & ((1ull<<x)-1));
+      if (channelMask.masks[i] & (1ull<<x)) {
+	y = __popcll(channelMask.masks[i] & ((1ull<<x)-1));
+	y = y + total;
         if (blockIdx.x == y) ncclShmem.channelId = x;
       }
     }
+    total = __popcll(channelMask.masks[i]);
+  }*/
```
wenkaidu: channelId cannot be blockIdx for release version. It needs to be counted from channelMasks

src/graph/connect.cc
```diff
@@ -290,7 +290,7 @@ static ncclResult_t connectTrees(struct ncclComm* comm, int* treeToParent, int*
   int t0u, t0d0, t0d1, t0ChildType, t1u, t1d0, t1d1, t1ChildType;
   int* ttp, *ttc0, *ttc1;
   NCCLCHECK(ncclGetDtree(nNodes, node, &t0u, &t0d0, &t0d1, &t0ChildType, &t1u, &t1d0, &t1d1, &t1ChildType));
-  if (comm->nChannels <= IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx94") ? (MAXCHANNELS/2) : (MAXCHANNELS/4)) {
+  if (comm->nChannels <= IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx94") ? (MAXCHANNELS/8) : (MAXCHANNELS/16)) {
```
wenkaidu: (MAXCHANNELS/2) : (MAXCHANNELS/16)

wenkaidu:
Can you check if more than 64 channels will break mirrorBits in https://github.com/ROCm/rccl/blob/develop/src/graph/topo.h#L264?

src/transport.cc
```diff
@@ -108,23 +112,26 @@ ncclResult_t ncclTransportP2pSetup(struct ncclComm* comm, struct ncclTopoGraph*
...
+    struct channelMasks sendMask = comm->connectSend[sendPeer+comm->nRanks*(connIndex == NCCL_CONN_IDX_P2P_NET ? NCCL_CONN_IDX_P2P_NET : 0)];
 
     // Data[i] contains all ncclConnect information for all send and receive connections with a given send and recv peer
     // This data is packed in the array based on the number of sendChannels and recvChannels connected with these peers
     // The first N entries contain recvData, connection information for recv connections
     // The next M entries contain sendData, connection information for send connections
     // It's not guaranteed that each entry of data has the same number of total or send/recv specific connections
     int p = i-(done+1);
-    if (recvMask || sendMask) NCCLCHECK(ncclCalloc(data+p, 2*MAXCHANNELS));
+    if ((recvMask.masks[0]) || (sendMask.masks[0])) NCCLCHECK(ncclCalloc(data+p, 2*MAXCHANNELS));
```
wenkaidu: shouldn't all masks need to checked for non-zero? I also suspect data+p needs multiply as well to match increased masks

src/transport.cc
```diff
@@ -208,7 +220,7 @@ ncclResult_t ncclTransportP2pSetup(struct ncclComm* comm, struct ncclTopoGraph*
             }
             TIME_STOP(4);
           }
-          if (sendMask || recvMask) {
+          if (sendMask.masks[0] || recvMask.masks[0]) {
```
wenkaidu: same here. Need to check all masks for non-zero value

wenkaidu (changes requested):

src/transport.cc
nusislam: `data` does not have anything to do with masks. the total size is already increased because MAXCHANNEL is now 256

wenkaidu: ok. still need a loop to check all masks. We cannot assume if mask[0] is empty all other will be same

src/include/comm.h
```diff
@@ -172,6 +172,10 @@ struct ncclNvlsMcHandleList {
   size_t size;
 };
 
+struct channelMasks {
+        uint64_t masks[4];
```
wenkaidu: let's also set MAXCHANNELS/64 here. This save 16 bytes in kernargs

wenkaidu (approved):

wenkaidu:
One more thing: it is likely debug buffer with GRAPH will overflow. Buffers sizes need to be double again similar to what was done in https://github.com/ROCm/rccl/pull/1035/

src/graph/paths.cc
```diff
@@ -918,7 +918,11 @@ ncclResult_t ncclTopoComputeP2pChannels(struct ncclComm* comm) {
   // fill the whole space of nChannels. To do so we mirror the bits in the
   // nChannels space.
   for (int c=0; c<comm->p2pnChannels; c++) {
-    comm->p2pChannels[c] = mirrorBits(c, comm->p2pnChannels);
+    //update mirrorbits to support more than 64 p2p channels
+    if (comm->p2pnChannels == 2*CHANNEL_LIMIT)
+      comm->p2pChannels[c] = mirrorBits(c, comm->p2pnChannels);
+    else
+      comm->p2pChannels[c] = mirrorBits(c, max(64,comm->p2pnChannels/2));
```
wenkaidu: I made a mirrorBits test app. it actually works fine to 128 channels. The logic here is wrong instead.

#include <stdio.h>

static int mirrorBits(int val, int pow2) {
  int mirror = 0;
  for (int b=1, mb=(pow2>>1); b<pow2; b<<=1, mb>>=1) if (val & b) mirror |= mb;
  return mirror;
}


int main()  {
	int nchans = 128;
	int mirrors[nchans];
	for (int c = 0; c < nchans; c ++) {
		mirrors[c] = mirrorBits(c, nchans);
		printf("c %d mirror %d\n", c, mirrors[c]);
	}
	for (int c = 0; c < nchans; c ++) {
		for (int d = 0; d < nchans; d ++) {
			if (c == d) continue;
			if (mirrors[c] == mirrors[d])
				printf("Error! Duplication detected at %d and %d\n", c, d);
		}
    }
    printf("All check passed!\n");
}

wenkaidu (changes requested):

src/graph/paths.cc
nusislam: Does not work for RCCL pt2pt tests. I would like to suggest a separate investigation for the pt2pt operations.

wenkaidu: so p2pnchannels cannot be more than 64 for now? if this is the case, we should limit it to 64 at the place where it is calculated. Not initializing the channels but not use it somehow

wenkaidu (approved):

PR #1203, opened 2024-06-06
Enable multi-threading for MSCCL

corey-derochie-amd:
## Details

**Work item:** Internal

**What were the changes?**  
Enabled RCCL to run in MSCCL mode using multi-threading instead of multi-process.

Many workloads already use multi-threading and would like the benefits of MSCCL without switching to multi-process.

The MSCCL code path was tracking static state, and was tracking thread-local state in a case where the threads used for initialization were not the same as the threads used to execute the collective. To model the multi-process case in the multi-threaded case, the static state was instead stored in a map for each rank to be accessed by each thread. Thread-local state remains thread-local but does not need to be carried over from an older context. Heuristics that disabled MSCCL for this mode were disabled.  
Unit tests were updated to support a single-process / multi-threaded mode using OpenMP by setting `UT_MULTITHREADED=1` and `UT_PROCESS_MASK=1`.

**Additional Documentation:**  
This behaviour can be confirmed using NCCL_DEBUG=INFO to see that the MSCCL functions are being called instead of the RCCL functions.

## Approval Checklist


src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -250,7 +228,7 @@ ncclResult_t mscclInit(ncclComm_t comm) {
     status.lastStream = nullptr;
     NCCLCHECK(mscclInitWorkFifoStatus(&(status.defaultWorkFifoStatus)));
 
-    mscclInitialized.store(true, std::memory_order_release);
+    mscclInitialized(comm->rank) = true;
```
wenkaidu: It would be better to have a function that dedicated for setting initialized to true. With that, we will able to implement lock to avoid race conditions and detect programming error like initializing msccl twice on same rank.

corey-derochie-amd: I can implement as a function, and detect the programming error. However, a lock isn't necessary here because this flag is only accessed from a single rank. Thoughts?

wenkaidu: I think the lock is to prevent race condition between read/write from different threads. BTW, there is an inherent bug from current MSCCL API: even in one process per GPU mode, each GPU can be associated with multiple communicators and has different rank numbers. mscclLoadAlgo(... int rank) was designed incorrectly. Maybe we should deprecate this API as well, and always use comm as mapping key

corey-derochie-amd: Only one thread should ever be accessing that flag at one time.

We should chat about the MSCCL bug. It might be out of scope for this work, or maybe we can easily deal with it here.

src/misc/msccl/msccl_status.cc
```diff
@@ -6,17 +6,66 @@
...
+  }
+  return *threadRankState;
+}
+
+bool mscclInitialized(int rank) {
+  return mscclGetRankState(rank).initialized;
+}
+
+void mscclSetInitialized(int rank, bool initialized) {
+  mscclGetRankState(rank).initialized = initialized;
```
wenkaidu: Let's assert a failure here if it is initialized already. This way we can capture programming error or unsupported op mode

src/misc/msccl/msccl_lifecycle.cc
wenkaidu: sure. at least the patch will not make it worse than what we have today. We can address the original issue later

src/misc/msccl/msccl_status.cc
corey-derochie-amd: Would you like an actual ASSERT or to log the error?

wenkaidu: I would like to see application to fail instantly with clear message with the reason. Otherwise we may get strange segfault or memory fault which takes really long time to figure out

wenkaidu (approved):
LGTM. Although I have not tested myself. We need to update UT to enable testing

test/common/TestBedChild.cpp
```diff
@@ -34,20 +38,30 @@
...
+
 namespace RcclUnitTesting
 {
-  TestBedChild::TestBedChild(int const childId, bool const verbose, int const printValues)
+  TestBedChild::TestBedChild(int const childId, bool const verbose, int const printValues, bool const useRankThreading)
   {
     this->childId = childId;
     this->verbose = verbose;
     this->printValues = printValues;
+    this->useRankThreading = useRankThreading;
```
gilbertlee-amd, line 73: If this is only going to be used to determine numThreadsToUse, I'd recommend just storing numThreadsToUse here instead.

test/common/TestBedChild.cpp
```diff
@@ -473,16 +487,30 @@ namespace RcclUnitTesting
...
     // Loop over all collectives to be executed in group call
     for (int collId = 0; collId < this->numCollectivesInGroup[groupId]; ++collId)
     {
       // Loop over all local ranks
+      ErrCode errCode = TEST_SUCCESS;
+      #ifdef ENABLE_OPENMP
+      if (this->verbose) INFO("Group %d collective %d running %d threads\n", groupId, collId, numThreadsToUse);
+      auto& errCodeVal = reinterpret_cast<int&>(errCode);
+      #pragma omp parallel for num_threads(numThreadsToUse) reduction(max : errCodeVal)
+      #endif
```
gilbertlee-amd: I don't think you actually need to worry about #ifdef-ing this out.
The pragma will just get ignored if we didn't compile for OpenMP.

test/common/TestBedChild.cpp
corey-derochie-amd: `numThreadsToUse` gets set to `numRanksToExecute`, which isn't known until it is read from the pipe.

test/common/TestBedChild.cpp
corey-derochie-amd: I'd prefer the `INFO` not to print, but I can just key that off of `this->useRankThreading`?

test/common/TestBedChild.cpp
gilbertlee-amd: Ah - right!

test/common/TestBedChild.cpp
gilbertlee-amd: Yes - I think so.

wenkaidu (approved):

PR #1208, opened 2024-06-10
use maxChannel 64 for multi-node inside kernel

nusislam:
For multi-node, the maxChannel value is set to 64. This value should be used inside the kernel also.

## Details

**What were the changes?**  
RCCL was using pre-defined MAXCHANNELS value inside the kernel. But the maxChannel is set to 64 for multi-node. So, inside the kernel also, it should use 64 as the maxChannel.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/common.h
+++ b/src/device/common.h
@@ -231,7 +231,13 @@ __forceinline__ __device__ void ncclKernelMain(struct ncclDevComm* comm, struct 
   const int tid = threadIdx.x;
   int x = tid;
   int total = 0, y;
-  int num = MAXCHANNELS/64 > 0 ? MAXCHANNELS/64 : 1;
+  int num;
+
+  if (comm->maxChannels == 64) {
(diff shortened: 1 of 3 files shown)


wenkaidu:
Why is this necessary? Are there some junks in channel mask array?

nusislam:
> Why is this necessary? Are there some junks in channel mask array?

If there were junks, I would expect tests to fail. All multi-node tests have passed so far.

I added this because this is more precise and the loop inside the kernel does not need to run for two iterations (single iteration should work for maxChannel=64).

wenkaidu:
If there is no functional issue without it, I would rather not make this change: it looks like a hack and is not worth it, especially for RCCL kernels which sub-microseconds latency is not an issue

nusislam:
Closing this PR as it does not seem necessary.

PR #1211, opened 2024-06-12
enable UT to test with channels greater than 64

saurabhAMD:
## Details
Enabling UT to test with channels greater than 64

**What were the changes?**  
Enabled UT to test with channels 56, 84 and 112 for allreduce if the architecture is gfx94 (MI300)

There was no UT to test configuration greater than 64 channels on MI300X previously.

1) Enable this feature if the architecture is MI300 (gfx94).
2) Set number of channels as 56, 84 and 112 for allreduce test.

**Additional Documentation:**  
1) Only 3 values of number of channels and 1 type of collective algorithm was chosen for testing to not increase the total time consumption of test by a lot.
2) Had to create a separate child process to get the architecture information as we cannot use HIP call prior to launching unless it is inside another child process.

## Approval Checklist

### DIFF
--- a/test/AllReduceTests.cpp
+++ b/test/AllReduceTests.cpp
@@ -102,6 +102,30 @@ namespace RcclUnitTesting
     testBed.Finalize();
   }
 
+  TEST(AllReduce, Channels)
+  {
+    TestBed testBed;
+    if(testBed.ev.isGfx94) {
+      // Configuration
+      std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAllReduce};
+      std::vector<ncclDataType_t> const dataTypes       = {ncclBfloat16, ncclHalf};
+      std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+      std::vector<int>            const roots           = {0};
+      std::vector<int>            const numElements     = {64 * 1024 * 1024, 1024};
+      std::vector<bool>           const inPlaceList     = {false};
+      std::vector<bool>           const managedMemList  = {false};
+      std::vector<bool>           const useHipGraphList = {false, true};
+      std::vector<char *>         const channelList     = {"56", "84", "112"}; 
+      for (auto channel : channelList) {
+        setenv("NCCL_MIN_NCHANNELS", channel, 1);
+        testBed.RunSimpleSweep(funcTypes, dataTypes, redOps, roots, numElements,
+                              inPlaceList, managedMemList, useHipGraphList);
+        testBed.Finalize();
+        unsetenv("NCCL_MIN_NCHANNELS");
+      }
+    }
+  }
+
   TEST(AllReduce, ManagedMemGraph)
   {
     TestBed testBed;
--- a/test/AllToAllTests.cpp
+++ b/test/AllToAllTests.cpp
@@ -85,4 +85,28 @@ namespace RcclUnitTesting
                            inPlaceList, managedMemList, useHipGraphList);
     testBed.Finalize();
   }
+
+    TEST(AllToAll, Channels)
+  {
+    TestBed testBed;
+    if(testBed.ev.isGfx94) {
+      // Configuration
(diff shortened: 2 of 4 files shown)


wenkaidu:
Can you add alltoall as well? sendrecv type of collectives have their own channels. Need to ensure they are not broken.

saurabhAMD:
> Can you add alltoall as well? sendrecv type of collectives have their own channels. Need to ensure they are not broken.

Okay makes sense, will add that. Based on our discussion with Gilbert, I earlier thought I am supposed to add it only for allreduce so that UT takes reasonable amount of time.

wenkaidu:
alltoall should be a fast test as there is only 1 data type and 1 op

test/AllReduceTests.cpp
```diff
@@ -102,6 +102,30 @@ namespace RcclUnitTesting
     testBed.Finalize();
   }
 
+  TEST(AllReduce, Channels)
+  {
+    TestBed testBed;
+    if(testBed.ev.isGfx94) {
+      // Configuration
+      std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAllReduce};
+      std::vector<ncclDataType_t> const dataTypes       = {ncclUint8};
```
wenkaidu: please use half and bfloat16, which are most likely being used in real world apps

test/AllReduceTests.cpp
```diff
@@ -102,6 +102,30 @@ namespace RcclUnitTesting
...
+      std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+      std::vector<int>            const roots           = {0};
+      std::vector<int>            const numElements     = {2500};
```
wenkaidu: Please at least use two numElements to cover both MSCCL and RCCL

test/AllReduceTests.cpp
```diff
@@ -102,6 +102,30 @@ namespace RcclUnitTesting
...
+      std::vector<bool>           const inPlaceList     = {false};
+      std::vector<bool>           const managedMemList  = {false};
+      std::vector<bool>           const useHipGraphList = {false};
```
wenkaidu: Both graph and non-graph need to be tested

test/AllReduceTests.cpp
```diff
@@ -102,6 +102,30 @@ namespace RcclUnitTesting
...
+  TEST(AllReduce, Channels)
+  {
+    TestBed testBed;
+    if(testBed.ev.isGfx94) {
+      // Configuration
+      std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAllReduce};
+      std::vector<ncclDataType_t> const dataTypes       = {ncclBfloat16, ncclHalf};
+      std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+      std::vector<int>            const roots           = {0};
+      std::vector<int>            const numElements     = {2500, 1024};
```
wenkaidu: 1024 will trigger MSCCL, which will be fine.
2500 is too small to guarantee triggering RCCL. Let's use at least 64M elements here.

test/AllToAllTests.cpp
```diff
@@ -85,4 +85,28 @@ namespace RcclUnitTesting
...
+    TEST(AllToAll, Channels)
+  {
+    TestBed testBed;
+    if(testBed.ev.isGfx94) {
+      // Configuration
+      std::vector<ncclFunc_t>     const funcTypes       = {ncclCollAllToAll};
+      std::vector<ncclDataType_t> const dataTypes       = {ncclBfloat16, ncclHalf};
+      std::vector<ncclRedOp_t>    const redOps          = {ncclSum};
+      std::vector<int>            const roots           = {0};
+      std::vector<int>            const numElements     = {2500, 1024};
```
wenkaidu: same here. use 64M and 1024

wenkaidu (approved):

PR #1212, opened 2024-06-13
Adding core binding in info

AtlantaPepsi:
## Details

**Work item:** Internal

**What were the changes?**  
Added core binding info after communicator init complete

For future debugging

## Approval Checklist

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -2004,7 +2004,7 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   }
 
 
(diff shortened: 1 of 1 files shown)


wenkaidu (approved):

edgargabriel (commented):
just to clarify, sched_getcpu() provides the core that the executable is running on, not the binding info. A process can be bound to multiple cores, but it can still only run on a single core at any given point in time. Retrieving the binding info would have to use sched_getaffinity(), which returns in a cpu_set of all cores that this process has been bound to.

Not sure what sched_getcpu() reports for multi-threaded  applications where threads might use different cores. Also note, that sched_getcpu() can also return an error instead of the info if the kernel doesn't support this feature.

PR #1222, opened 2024-06-24
Unit Tests for testing channels

saurabhAMD:
## Details
Enabling UT to test with channels greater than 64 with keeping reasonable run time.

**What were the changes?**  
Enabled UT to test with channels 84 and 112 for allreduce and 112 for allToAll if the architecture is gfx94 (MI300)

The run time for UT was high, had to reduce it.

1) Removing 56 channels from UT
2) Not allowing to set more than 64 channels if number of GPUs is less than 8. This change was made in RCCL behavior in general. Also added a warning message so that the users are made aware of it.
3) Not running the UT when less than 8 GPUs.

**Additional Documentation:**  
None

## Approval Checklist

### DIFF
--- a/src/graph/connect.cc
+++ b/src/graph/connect.cc
@@ -674,7 +674,11 @@ ncclResult_t ncclTopoPostset(struct ncclComm* comm, int* firstRanks, int* treePa
 
   int minNchannels = ncclMinNchannels();
   if (comm->nNodes > 1) {
-     minNchannels = std::min(64, maxChannels);
+    minNchannels = std::min(64, maxChannels);
+  }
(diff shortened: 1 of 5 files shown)


wenkaidu (approved):

corey-derochie-amd (approved):
LGTM. However, I wonder if there's a way to leave the existing tests but have them only run in a certain mode, but the default has a more reasonable runtime?

wenkaidu:
> LGTM. However, I wonder if there's a way to leave the existing tests but have them only run in a certain mode, but the default has a more reasonable runtime?

The issue is we enabled RCCL to run 112 channels and recommended to end users. Without these new UT, these use cases do not have any testing.

PR #1225, opened 2024-06-25
Improved versioning info, added hostname and librccl path in NCCL_DEBUG=VERSION

rahulvaidya20:
## Details

**Work item:** Internal

**What were the changes?**  
To improve HIP and ROCm versioning info (includes build number, patch and githash). Added support to display hostname and path to librccl.so.

Displaying complete version info helps in understanding the rocm/hip build type (mainline, staging etc). Enabling support for displaying hostname can retrieve (dell, banff, etc if specified in hostname). Path to librccl.so is important to understand user's librccl path loaded by the dynamic linker at runtime.

**Changes to init.cc**: Added a new macro, VERSION_STRING_EXTENDED to display hip and rocm version. Added support to display hostname using **gethostname** function.
**New files**: 
1. **hip_rocm_version_info.h**: Checks rocm version and sets the macro HIP_BUILD_INFO, ROCM_BUILD_INFO. 
2. **rccl_libpath_info.h** and **rccl_libpath_info.cc**: Added to new function to display librccl.so path used by dynamic linker at runtime. Used dynamic linker api to achieve this.

**Additional Documentation:**  

Sample output:

image

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -371,6 +371,7 @@ set(SRC_FILES
   src/include/git_version.h
   src/include/graph.h
   src/include/group.h
+  src/include/hip_rocm_version_info.h
   src/include/ibvcore.h
   src/include/ibvsymbols.h
   src/include/ibvwrap.h
@@ -391,6 +392,7 @@ set(SRC_FILES
   src/include/rccl_vars.h
   src/include/register.h
   src/include/rccl_float8.h
+  src/include/rccl_libpath_info.h
   src/include/rocm_smi_wrap.h
   src/include/rocmwrap.h
   src/include/roctx.h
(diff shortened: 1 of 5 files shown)


wenkaidu:
We want to get the path of currently loaded librccl.so. It could be loaded through LD_LIBRARY_PATH or LD_PRELOAD. Maybe this solution is better?
https://stackoverflow.com/questions/43409167/find-location-of-loaded-shared-library-from-in-that-shared-library

rahulvaidya20:
> We want to get the path of currently loaded librccl.so. It could be loaded through LD_LIBRARY_PATH or LD_PRELOAD. Maybe this solution is better? https://stackoverflow.com/questions/43409167/find-location-of-loaded-shared-library-from-in-that-shared-library
@wenkaidu, this is a neat solution. Will update my PR with these changes shortly. Thanks!

wenkaidu:
To make patch simpler, can we avoid adding new files, like rccl_libpath_info.h and rccl_libpath_info.cc? There are plenty of symbols can be used. API functions, like ncclCommInitRank, are very stable and maybe better candidates.

rahulvaidya20:
That makes sense. I'll go ahead and update my changes.

rahulvaidya20:
Created a new PR for this: https://github.com/ROCm/rccl/pull/1232

PR #1228, opened 2024-06-26
[GRAPH] Use channel shuffling only for IB systems

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
- Re-introduce channel shuffling for MI300x systems with Infiniband interconnect.
- Fix channels=48 for RoCE systems 

Channel shuffling improves performance on IB systems, but adversely affects RoCE systems.

`NCCL_IB_GID_INDEX` needs to be defined as **3** for RoCE systems. If this is the case, do not use channel shuffling rings and use channels=48, else, use channel shuffling and channels=64.

## Approval Checklist


gilbertlee-amd (approved):

src/graph/rome_models.cc
```diff
@@ -642,8 +889,9 @@ static struct rcclRomeModel rome_model_81 = {
                 "N6 6 2 0 7 5 3 1 4 N4|"
                 "N7 7 3 2 6 0 4 1 5 N5|"
                 "N1 1 0 2 4 3 5 7 6 N6|"
-                "N0 0 1 2 3 4 5 6 7 N7",
-  .options    = "noCpuCheck=1,tuning=5,disableNumaMatching=1",
+		"N0 0 1 2 3 4 5 6 7 N7",
+
+  .options    = "noCpuCheck=1,tuning=5,disableNumaMatching=1,gid3=1",
```
gilbertlee-amd: For code clarity, I think we should rename this option to isRoCE (or however the capitalization should be).

gilbertlee-amd: Alternatively, set this to gidIndex=3 and check that we match that value if the option is set.

gilbertlee-amd (approved):

wenkaidu (approved):

src/graph/rome_models.cc
```diff
@@ -1367,12 +1616,23 @@ ncclResult_t parseRome4P2H(struct ncclTopoSystem* system, struct ncclTopoGraph*
...
   for (i = 0; i < sizeof(romeTopoModels)/sizeof(romeTopoModels[0]); i++) {
     bool ignore_cpu = checkOption(romeTopoModels[i].options, "noCpuCheck");
     if (!ignore_cpu && (arch != NCCL_TOPO_CPU_ARCH_X86 || vendor != NCCL_TOPO_CPU_VENDOR_AMD || model != NCCL_TOPO_CPU_TYPE_ROME))
       continue;
+
     bool ignore_numa = checkOption(romeTopoModels[i].options, "disableNumaMatching");
     if (!ignore_numa && romeTopo.nCpus != romeTopoModels[i].nCpus) continue;
+
+    // check if isRoCE=1 is defined in model struct
+    bool optionsIsRoCE = checkOption(romeTopoModels[i].options, "isRoCE");
```
wenkaidu: what happen with models doesn't have "isRoCE" option?

nileshnegi: `checkOption` returns false if "isRoCE" is not found in options, so `optionsIsRoCE` will be 0.\
variable `isRoCE` in line 1621 is initialized as false, so line 1634: `if (optionsIsRoCE != isRoCE) continue;` will not come into effect.

verified with model match for single node run, which shows "Rome model index 38 with GPU mapping: 0 1 2 3 4 5 6 7" i.e. `rome_model_79` which does not have "isRoCE" in options

wenkaidu: if the purpose is to skip "if (optionsIsRoCE != isRoCE) continue" in this case, optionsIsRoCE needs to have 3 values: not present, true and false.

wenkaidu (approved):

src/graph/connect.cc
```diff
@@ -627,8 +627,19 @@ ncclResult_t ncclTopoPostset(struct ncclComm* comm, int* firstRanks, int* treePa
   // Only use full MAXCHANNELS for gfx94x
   int maxChannels = IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx94") ? MAXCHANNELS : 2*CHANNEL_LIMIT;
 
+  // Define channels=64 for multi-node systems
+  int multiNodeNchannels = 64;
```
wenkaidu: We should only use more than 32 channels on MI300

wenkaidu (approved):

wenkaidu (approved):

PR #1296, opened 2024-08-14
Mustafa/tune and fix validation allreduce 8n

mustafabar:
## Details
This PR addresses an issue that results in an all-reduce validation error in the RCCL-MSCCL multithreaded integration path with 8 GPUs and small message sizes. Also, it shaves off some of the degradations for small messages at that scale reported internally

**Work item:** _Internal_

**What were the changes?**  
Changes are primarily to the generated XML files for 8 GPUs to address the bug in LL protocol 

This is to fix an accidental finding while addressing a regression issue for multithreaded MSCCL usage for allreduce. Upon running with `-t1 -g8` for small message range, wrong values were reported by RCCL's allreduce-perf test

I have tuned the msccl-algorithms using different Python scripts that are based on a DSL developed by MSFT. At the runtime interpreter level, there's an issue with LL protocol going out of bounds. As a temporary fix, I have replaced it with Simple and also reduced the number of channels created to compensate for the latency incurred by Simple. The overall performance is equal to or better than the baseline

**Additional Documentation:**  
To compare and contrast the results, change the `MSCCL_ALGO_DIR` env variable to include an absolute path to the older version. 
- To reproduce validation error on the MI300 nodes, run allreduce with `-t1 -g8` 
- To compare performance run allreduce with `-t8 -g1`


tools/msccl-algorithms/allreduce-allpairs-8n-Simple-1pass-op.xml
```diff
@@ -0,0 +1,2178 @@
+<algo name="allreduce_pairs" proto="Simple" nchannels="2" nchunksperloop="16" ngpus="8" coll="allreduce" inplace="0" outofplace="1" minBytes="0" maxBytes="25599">
```
nusislam, line 1: We are supposed to use LL protocol for such small message sizes. Does the new xml provider better performance for all configs (1 GPU per process, etc.)?

mustafabar: The performance is unchanged (more or less the same) compared to LL for 1 GPU per thread. I can test it out with 8 MPI processes

mustafabar: Comparison is in this comment https://ontrack-internal.amd.com/browse/LWPCOMMLIBS-422#:~:text=I-,resolved,-the%20noticeable%20degradation

wenkaidu: I think we should try to understand where the regression coming from before using workaround. Switching LL to Simple seems to be dramatic measure.

mustafabar: If you run it a few times, LL also results in a validation error with `-g 8 -t 1 -b4 -e 1M -f 2`. This has to be investigated as a defect.

PR #1314, opened 2024-08-29
Include mscclpp as a git submodule

corey-derochie-amd:
## Details

**What were the changes?**  
Include the mscclpp source code as a git submodule rather than using `download_project` in the `cmake` step.

Some environments don't have internet access at compile time, so `download_project` won't work. Instead, we should checkout all code together when we do have access.

We are still technically using `download_project` to maintain the same semantics, but we skip the actual checkout step and just use the build step.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "ext-src/mscclpp"]
+	path = ext-src/mscclpp
+	url = https://github.com/microsoft/mscclpp.git
+	ignore = dirty
--- a/README.md
+++ b/README.md
@@ -66,12 +66,17 @@ By default, RCCL builds for all GPU targets defined in `DEFAULT_GPUS` in `CMakeL
 ### To build the library using CMake:
 
 ```shell
-$ git clone https://github.com/ROCm/rccl.git
+$ git clone https://github.com/ROCm/rccl.git --recurse-submodules
 $ cd rccl
 $ mkdir build
 $ cd build
 $ cmake ..
 $ make -j 16      # Or some other suitable number of parallel jobs
+```
+If you have already cloned, you can checkout the `mscclpp` submodule manually.
+```shell
+$ cd ext-src/mscclpp
+$ git submodule update --init --recursive
 ```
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "ext-src/mscclpp"]
+	path = ext-src/mscclpp
+	url = https://github.com/microsoft/mscclpp.git
```
nusislam, line 3: How are you specifying the commit hash?

ext-src/mscclpp
corey-derochie-amd, line 1: @nusislam This specifies the commit hash.

cmake/MSCCLPP.cmake
```diff
@@ -63,15 +73,16 @@ if(ENABLE_MSCCLPP)
         mscclpp_cmake_arg(HIP_COMPILER)
     
         download_project(PROJ                mscclpp_nccl
-                         GIT_REPOSITORY      https://github.com/microsoft/mscclpp.git
-                         GIT_TAG             1e82dd444fc1ed8b7add354eebaab8a94e67d5fc
+                         # GIT_REPOSITORY      https://github.com/microsoft/mscclpp.git
+                         # GIT_TAG             1e82dd444fc1ed8b7add354eebaab8a94e67d5fc
```
corey-derochie-amd, line 77: I left this in as a comment to act as documentation of the commit hash, since it's more visible. But there's also the risk of them getting out of sync if the maintainer forgets to update it. Should it be removed?

.gitmodules
corey-derochie-amd: Identified in the other comment.

README.md
```diff
@@ -66,7 +66,7 @@ By default, RCCL builds for all GPU targets defined in `DEFAULT_GPUS` in `CMakeL
 ### To build the library using CMake:
 
 ```shell
-$ git clone https://github.com/ROCm/rccl.git
```
corey-derochie-amd, line 69: Add a comment about how to do submodule after the clone.

PR #1319, opened 2024-09-04
[ON HOLD] Unit Tests for CPX mode

saurabhAMD:
## Details

**Work item:** "Internal"

**What were the changes?**  
Enabling Unit Tests to run for CPX mode.

Currently UT tests all possible numbers of GPUs up to 16. If same scheme is used to test CPX mode, UT will take forever to complete.

1) Detect if system is in CPX mode.
2) If yes, only test pow2 number of GPUs.
3) Fit GPUs into minimal number of sockets. Unique ID used to detect the OAM for the gpu. Implementation logic: GPUs mapped to sockets associated with a greater number of GPUs are given higher preference for assignment.

**Additional Documentation:**  
None.

## Approval Checklist


test/common/EnvVars.cpp
```diff
@@ -88,6 +90,117 @@ namespace RcclUnitTesting
...
+      int status;
+      if (read(pipefd[0], cpxMode, sizeof(*cpxMode)) != sizeof(*cpxMode)) return TEST_FAIL;
+      waitpid(pid, &status, 0);
+      assert(!status);
+      close(pipefd[0]);
+      close(pipefd[1]);
+    }
+    return TEST_SUCCESS;
+    return 0;
+  }
```
corey-derochie-amd, line 201: I'm curious why `rocm-smi` is run in a fork process?

saurabhAMD: Because we cannot use HIP call prior to launching unless it is inside another child process. You can refer to line 207 for this comment. The same procedure is used for other features such as finding the architecture of the system as in line 211.

corey-derochie-amd: In your case, I think `popen` already forks and creates a pipe internally, so you may not need your own fork. If so, `getDeviceMode` and `getDevicePriority` can be simplified.

test/common/EnvVars.cpp
```diff
@@ -115,6 +228,18 @@ namespace RcclUnitTesting
     // Total number of reduction ops
     int numOps = ncclNumOps;
 
+    gpuPriorityOrder.resize(numDetectedGpus);
+    for(int i=0;i<numDetectedGpus;i++){
+      gpuPriorityOrder[i] = i;
+    }
+    if(isGfx94) {
+      bool cpxMode = false;
+      getDeviceMode(&cpxMode);
```
corey-derochie-amd, line 237: The return value of `getDeviceMode` is never used, so if it can be simplified as I've suggested, you could rename it to `isDeviceCpxMode` and return `bool`.

corey-derochie-amd:
This PR is made redundant by #1324 and #1403

PR #1324, opened 2024-09-06
Enabling Unit Tests for CPX mode

saurabhAMD:
## Details

**Work item:** "Internal"

**What were the changes?** 
Enabling Unit Tests to run for CPX mode. 

Currently UT tests all possible numbers of GPUs up to 16. If same scheme is used to test CPX mode, UT will take forever to complete. 

1) Detect if system is in CPX mode.
2) If yes, only test pow2 number of GPUs.

**Additional Documentation:**  
None.

## Approval Checklist

### DIFF
--- a/test/common/EnvVars.cpp
+++ b/test/common/EnvVars.cpp
@@ -139,7 +139,6 @@ namespace RcclUnitTesting
     showNames      = GetEnvVar("UT_SHOW_NAMES"  , 1);
     minGpus        = GetEnvVar("UT_MIN_GPUS"    , 2);
     maxGpus        = GetEnvVar("UT_MAX_GPUS"    , numDetectedGpus);
-    onlyPow2Gpus   = GetEnvVar("UT_POW2_GPUS"   , false);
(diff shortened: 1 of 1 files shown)


gilbertlee-amd (changes requested):
We shouldn't be overriding UT_POW2_GPUs.  
Instead, you should be setting the default based on whether or not we are in CPX mode, rather than  hard-assigning.  
This way, if a user wants to run non-power of 2s, they still can (by setting UT_POW2_GPUS=0).

gilbertlee-amd (approved):

test/common/EnvVars.cpp
```diff
@@ -116,6 +151,13 @@ namespace RcclUnitTesting
     // Total number of reduction ops
     int numOps = ncclNumOps;
 
+    bool cpxMode = false;
```
gilbertlee-amd, line 154: For consistency, please rename this bool as isCpxMode

saurabhAMD: I have raised the pull request with the requested change: https://github.com/ROCm/rccl/pull/1327/files

PR #1337, opened 2024-09-18
Fix regression in multi-thread mode even while MSCCL is disabled

corey-derochie-amd:
## Details

**Work item:** Internal

**What were the changes?**  
When MSCCL is enabled, only adjust `minNchannels` if `mscclCompatible` is true.

Even if MSCCL is enabled, we do a further check for compatibility and set the `mscclCompatible` flag. MSCCL will only be used in this case, so we shouldn't do any further adjustments if it's not.

**Additional Documentation:**  
This was causing a significant performance regression.

## Approval Checklist

### DIFF
--- a/src/graph/connect.cc
+++ b/src/graph/connect.cc
@@ -690,7 +690,9 @@ ncclResult_t ncclTopoPostset(struct ncclComm* comm, int* firstRanks, int* treePa
   if (mscclEnabled() && (comm->topo->mscclEnabled || mscclForceEnabled())) {
     int mscclNumChannelsRequired = maxNchannels;
     mscclSchedulerInit(comm, &mscclNumChannelsRequired);
-    minNchannels = std::max(minNchannels, mscclNumChannelsRequired);
+    if (comm->mscclCompatible) {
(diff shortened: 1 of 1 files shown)


corey-derochie-amd:
@nusislam @wenkaidu This addresses a performance regression when MSCCL is disabled for single-process (i.e. single-thread or multi-thread). Does it look valid?

wenkaidu:
@nileshnegi I think we decided that MSCCL should observe NCCL_MAX_NCHANNELS set by user

corey-derochie-amd:
> @nileshnegi I think we decided that MSCCL should observe NCCL_MAX_NCHANNELS set by user

Would you suggest, then, that we do not set `minNchannels` at all, rather than conditionally setting it?

wenkaidu (approved):
sorry, I see what you mean. The PR looks good to me

nileshnegi:
> @nileshnegi I think we decided that MSCCL should observe NCCL_MAX_NCHANNELS set by user

this is enforced in the call to `mscclSchedulerInit()`
```
int mscclNumChannelsRequired = maxNchannels;
mscclSchedulerInit(comm, &mscclNumChannelsRequired);
```

Issue #1254, opened 2024-07-14
[Issue]: Unexpected fallthrough?


r-barnes:
### Problem Description

There is an undocumented fallthrough [here](https://github.com/ROCm/rccl/blob/9cbb3da224de813bf1f2356b4431f7b375ee6d35/src/misc/msccl/msccl_lifecycle.cc#L520):
```
      threadLocalStatus.groupStatus = mscclGroupUnsupportedOp;
      NCCLCHECK(mscclFallBackSavedParams());
    case mscclGroupUnsupportedOp:
      NCCLCHECK(mscclFallBackSavedParams());
      break;
```
This looks like a bug. Is the fallthrough intentional (mark with `[[fallthrough]]`) or should it be a `break`?

### Operating System

N/A

### CPU

N/A

### GPU

AMD Instinct MI300X

### ROCm Version

ROCm 6.1.0, ROCm 6.0.0, ROCm 5.7.1, ROCm 5.7.0, ROCm 5.6.0, ROCm 5.5.1, ROCm 5.5.0

### Steps to Reproduce

Code issue

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

N/A

corey-derochie-amd:
This fix will be present in ROCm 6.4.

PR #1342, opened 2024-09-19
Remove unhandled switch case fall-through in MSCCL

corey-derochie-amd:
## Details

**Work item:** #1254

**What were the changes?**  
Added missing `break` in switch case.

Switch case was falling through unhandled.

**Additional Documentation:**  
Need to determine if this fall-through was intentional.

## Approval Checklist

### DIFF
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -595,6 +595,7 @@ ncclResult_t mscclEnqueueCheck(
         }
       threadLocalStatus.groupStatus = mscclGroupUnsupportedOp;
       NCCLCHECK(mscclFallBackSavedParams());
+      break;
     case mscclGroupUnsupportedOp:
       NCCLCHECK(mscclFallBackSavedParams());
       break;


mustafabar (approved):

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -592,6 +592,7 @@ ncclResult_t mscclEnqueueCheck(
         }
       threadLocalStatus.groupStatus = mscclGroupUnsupportedOp;
       NCCLCHECK(mscclFallBackSavedParams());
+      break;
     case mscclGroupUnsupportedOp:
```
corey-derochie-amd, line 596: Alternatively, is this fall-through intentional? If so, it double calls `mscclFallBackSavedParams`. Is that also intentional?

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -592,6 +592,7 @@ ncclResult_t mscclEnqueueCheck(
...
       NCCLCHECK(mscclFallBackSavedParams());
```
corey-derochie-amd, line 600: Alternatively, is this fall-through intentional? If so, it double calls `mscclFallBackSavedParams`. Is that also intentional?

mustafabar: The call to `threadLocalStatus.savedSchedulerParams.clear();` at the end of the fallback function tells me it is unintentional. No one tries to refill the saveSchedulerParams in between as it is thread_local.

PR #1347, opened 2024-09-24
MSCCL Multithreaded regression root cause fix

mustafabar:
## Details

**Work item:** _Internal_

**What were the changes?**  
_Address MT MSCCL issue and reenable MSCCL in MT mode through enabling MSCCL single process mode_

_Severe regression when multiple devices are used per thread in rccl-tests allreduce._

_Set the device id correctly so that scratch memory and sync buffers are allocated on the right device_

## Approval Checklist

### DIFF
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -28,7 +28,7 @@
 
 RCCL_PARAM(MscclEnabled, "MSCCL_ENABLE", 1);
 RCCL_PARAM(MscclForceEnabled, "MSCCL_FORCE_ENABLE", 0);
-RCCL_PARAM(MscclEnableSingleProcess, "MSCCL_ENABLE_SINGLE_PROCESS", 0);
+RCCL_PARAM(MscclEnableSingleProcess, "MSCCL_ENABLE_SINGLE_PROCESS", 1);
 static const char* mscclAlgoFilePathEnv = "MSCCL_ALGO_FILE_PATH";
 
 bool mscclEnabled() {
--- a/src/misc/msccl/msccl_status.cc
+++ b/src/misc/msccl/msccl_status.cc
@@ -27,9 +27,9 @@ static mutex rankStatesMutex;
 static unordered_map<int, shared_ptr<mscclRankState>> rankStates;
 
 static inline mscclRankState& mscclGetRankState(int rank) {
-  static thread_local shared_ptr<mscclRankState> threadRankState = make_shared<mscclRankState>();
-
+  // In the unlikely case of negative rank, return a per-thread state
   if (rank < 0) {
+    static thread_local shared_ptr<mscclRankState> threadRankState(new mscclRankState());
     return *threadRankState;
   }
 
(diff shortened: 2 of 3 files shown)


gilbertlee-amd (approved):

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -28,7 +28,7 @@
 
 RCCL_PARAM(MscclEnabled, "MSCCL_ENABLE", 1);
 RCCL_PARAM(MscclForceEnabled, "MSCCL_FORCE_ENABLE", 0);
-RCCL_PARAM(MscclEnableSingleProcess, "MSCCL_ENABLE_SINGLE_PROCESS", 0);
+RCCL_PARAM(MscclEnableSingleProcess, "MSCCL_ENABLE_SINGLE_PROCESS", 1);
```
corey-derochie-amd, line 31: I'm not sure we want to enable by default, this will enable MSCCL for the single-threaded case as well. @gilbertlee-amd what is your opinion?

src/misc/msccl/msccl_status.cc
```diff
@@ -27,18 +27,20 @@ static mutex rankStatesMutex;
 static unordered_map<int, shared_ptr<mscclRankState>> rankStates;
 
 static inline mscclRankState& mscclGetRankState(int rank) {
-  static thread_local shared_ptr<mscclRankState> threadRankState = make_shared<mscclRankState>();
-
+  // In the unlikely case of negative rank, return a per-thread state
   if (rank < 0) {
+    static thread_local shared_ptr<mscclRankState> threadRankState(new mscclRankState());
```
corey-derochie-amd, line 32: I don't think the negative `rank` is an unlikely case, although this might still be fine.

(`rank` has a default value of `-1` and early in the code this function will be called without a rank.)

src/misc/msccl/msccl_status.cc
```diff
@@ -27,18 +27,20 @@ static mutex rankStatesMutex;
...
   lock_guard<mutex> lock(rankStatesMutex);
 
   auto rankStateIt = rankStates.find(rank);
   if (rankStateIt == rankStates.end()) {
-    rankStateIt = rankStates.insert(make_pair(rank, make_shared<mscclRankState>(*threadRankState))).first;
-    rankStateIt->second->rank = rank;
+    // Create a per rank threadRankState rather than per thread
+    shared_ptr<mscclRankState> newthreadRankState(new mscclRankState());
+    newthreadRankState->rank = rank;
+    rankStateIt = rankStates.insert(make_pair(rank, newthreadRankState)).first;
```
corey-derochie-amd, line 43: The original code already made a per-rank state rather than a per-thread state. It was copying from the existing thread state in case changes had been made to it earlier in the code path. This is the whole reason a negative rank is not unlikely above. It is by design.

corey-derochie-amd:
I don't think CI was ever run in a configuration where this path was tested.

src/misc/msccl/msccl_status.cc
mustafabar: It created a per-rank state, but the previous `static thread_local` indicates that if multiple ranks are on the same thread, they will share the state pointer. That's my understanding of `static thread_local`.

corey-derochie-amd: They will, but the original code was making a copy to preserve any state information that was set before the rank ID was known.

PR #1365, opened 2024-10-07
disabled wbinvl1 for gfx9x on ll128

akolliasAMD:
## Details
Built in is no longer valid on gfx9 so it creates compilations issues. ll128 will need rework to be reenabled at gfx9x

## Approval Checklist

### DIFF
--- a/src/device/prims_ll128.h
+++ b/src/device/prims_ll128.h
@@ -13,7 +13,7 @@
 #define NCCL_LL128_FLAGTHREAD (NCCL_LL128_LINEELEMS-1)
 
 #ifndef RCCL_USE_WBINVL1_VOL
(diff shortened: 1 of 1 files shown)


src/device/prims_ll128.h
```diff
@@ -13,7 +13,7 @@
 #define NCCL_LL128_FLAGTHREAD (NCCL_LL128_LINEELEMS-1)
 
 #ifndef RCCL_USE_WBINVL1_VOL
-#if defined(__GFX8__) || defined(__GFX9__)
+#if defined(__GFX8__)
```
wenkaidu: we should only disable on gfx94x, not all gfx9

wenkaidu (approved):

nileshnegi (approved):

PR #1367, opened 2024-10-09
[BUILD] Require use of Python3 interpreter

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Use the Python3 interpreter in the CMake build process to generate host/device tables.

`src/device/generate.py` fails with syntax errors if the default Python interpreter is Python2.

Replace `find_package(PythonInterp REQUIRED)` with `find_package(Python3 COMPONENTS Interpreter)` in CMakeLists.txt.

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -591,15 +591,24 @@ endforeach()
 
 # Generate device/host tables and all the collective functions that are going to be in librccl.so
 #==================================================================================================
-find_package(PythonInterp REQUIRED)
+find_package(Python3 COMPONENTS Interpreter REQUIRED)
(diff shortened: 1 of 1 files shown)


corey-derochie-amd:
This won't prevent the build from failing, right? It will just fail with a useful message?

nileshnegi:
> This won't prevent the build from failing, right? It will just fail with a useful message?

Yes, it will specifically search for Python3 (as systems can have both Python2 and Python3 installed), and fail if Python3 is not pre-installed.

corey-derochie-amd (approved):

wenkaidu (approved):

PR #1429, opened 2024-11-22
Memory leak fix when numIBDevices = 0

PJAvinash:
## Details
These changes address Memory leak bug when number of IB devices on a system = 0. 

Right now we are invoking wrap_ibv_free_device_list() only when number of IB devices on a system are Non Zero. But ibv_free_device_list() must be called everytime there is an invocation to ibv_get_device_list(). 

**Work item:** "Internal"
**What were the changes?**  
- Modifying / enhancing failure path in ncclIBInit().
- pthread_detach()  is called on same thread twice , making the effect idempotent. Hence the second call is removed.

- Code improvements
- To fix memory leaks on MI210 when number of IB devices = 0;

- Analyzing the code with debugger
**Additional Documentation:**  
- pthread_detach() call is removed as it is a side effect of earlier nccl sync

## Approval Checklist


thananon (approved):
This looks good to me.

src/transport/net_ib.cc
```diff
@@ -505,9 +510,6 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction) {
           ncclIbMergedDevs[mergedDev].speed += ncclIbDevs[ncclNIbDevs].speed;
           ncclNIbDevs++;
           nPorts++;
-          // [RCCL]
-          pthread_detach(ncclIbAsyncThread);
-          // [/RCCL]
```
wenkaidu, line 547: Why removing pthread_detach? This was added to fix a memory leak issue.

src/transport/net_ib.cc
PJAvinash, line 1: pthread_detach() is being called twice with same parameter, first time at line 486, second call will  simply return -1 and it is not being checked, and has no effect.

src/transport/net_ib.cc
PJAvinash: pthread_detach() is being called twice with same parameter, first time at line 486, second call will simply return -1 and it is not being checked, and has no effect.

wenkaidu (approved):

corey-derochie-amd (changes requested):
Please provide a more descriptive title, as this will become the commit message.

src/transport/net_ib.cc
```diff
@@ -457,8 +457,8 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction) {
       if (ncclSuccess != wrap_ibv_get_device_list(&devices, &nIbDevs)) { ret = ncclInternalError; goto fail; }
 
       // Should NCCL merge multi-port devices into one?
-      int mergeNics;
-      mergeNics = ncclParamIbMergeNics();
+      int mergeNics = ncclParamIbMergeNics();
+
 build_ib_list:
```
corey-derochie-amd, line 462: This whole block is removed in NCCL 2.24, so won't matter for long.

src/transport/net_ib.cc
```diff
@@ -562,8 +563,7 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction) {
           }
         }
       }
-
-      if (nIbDevs && (ncclSuccess != wrap_ibv_free_device_list(devices))) { ret = ncclInternalError; goto fail; };
+      if (ncclSuccess != wrap_ibv_free_device_list(devices)) { ret = ncclInternalError; goto fail;}
```
corey-derochie-amd, line 566: The original behaviour would avoid returning an error if there were no devices. To maintain that behaviour, you could do this.
```suggestion
      if ((ncclSuccess != wrap_ibv_free_device_list(devices)) && nIbDevs) { ret = ncclInternalError; goto fail; }
```

corey-derochie-amd: Doing so would also skip your `WARN` below, so you would have to decide whether that's necessary in that case.

corey-derochie-amd (approved):
LGTM, I've left a few suggestions to consider.

src/transport/net_ib.cc
PJAvinash: Ideally we would like to detect if wrap_ibv_free_device_list() is failing/not behaving as expected for some reason ?

corey-derochie-amd: I'm honestly not sure if it matters for an empty list, but in general I agree we would like to know if this call is failing, as I think it should never fail.

PR #1431, opened 2024-11-22
Adding RCCL_MODEL_REVERSAL_DISABLE env var to disable model reversal

gilbertlee-amd:
## Details
Adding a RCCL_MODEL_REVERSAL_DISABLE environment variable that disables the normal reversal of even/odd nodes for rings.
This is to enable experimenting the whether or not model reversal provides benefits.

None

**What were the changes?**  
Adding env var that forces reversal to false when parsing.

To enable flipping on/off "rail-optimized" rings.

See above.

**Additional Documentation:**  
none

## Approval Checklist

### DIFF
--- a/src/graph/rome_models.cc
+++ b/src/graph/rome_models.cc
@@ -1137,7 +1137,7 @@ ncclResult_t parseGraph(const char* str, struct ncclTopoSystem* system, struct n
   int ngpus = system->nodes[GPU].count;
   int nnets = system->nodes[NET].count;
 
-  if (rcclParamModelReversalDisable) reverse = 0;
+  if (rcclParamModelReversalDisable()) reverse = 0;
 
   do {
     if (str[offset] == 'N') {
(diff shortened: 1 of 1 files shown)


src/graph/rome_models.cc
```diff
@@ -1133,6 +1136,9 @@ ncclResult_t parseGraph(const char* str, struct ncclTopoSystem* system, struct n
   int net_offset = 0, net_count = 0;
   int ngpus = system->nodes[GPU].count;
   int nnets = system->nodes[NET].count;
+
+  if (rcclParamModelReversalDisable) reverse = 0;
```
nileshnegi: Isn't this supposed to be a function call like `rcclParamModelReversalDisable()`?

wenkaidu: > Isn't this supposed to be a function call like `rcclParamModelReversalDisable()`?

agree

gilbertlee-amd: Good catch - I didn't notice it because it compiles because test for function always returns true.

wenkaidu (approved):

nileshnegi (approved):

PR #1434, opened 2024-11-26
Add HDP flush for gfx940

wenkaidu:
## Details

Internal

**What were the changes?**  
Add HDP flush for gfx940

Avoid potential data corruption

Add HDP flush for gfx940

**Additional Documentation:**  

## Approval Checklist


thananon (approved):
If this is what I think it is doing, then it looks good to me.

General question, why was it done the other way before this PR?

wenkaidu:
> If this is what I think it is doing, then it looks good to me.
> 
> General question, why was it done the other way before this PR?

It was all about saving a micro-second when possible. I need to convert it back to draft as testing is still ongoing

src/device/prims_simple.h
```diff
@@ -52,7 +52,7 @@ class Primitives<
   uint64_t connStepCache; // Cache last seen value of (*connStepPtr)
   int      connStepSize; // Connection step size
   uint64_t* barriers;
-  uint64_t* barrier_next;
+  uint64_t barrier_next;
```
cenzhaometa: these are per thread-local variables residing on registers, right?

wenkaidu: yes

src/device/prims_simple.h
```diff
@@ -674,8 +674,9 @@ class Primitives<
 
     // For send operations, we need an extra warp to overlap the threadfence and the copy
     if (tid == 0) ncclShmem.groups[group].warpStart = threadIdx.x/WARP_SIZE;
-    barriers = ncclShmem.groups[group].barrier;
-    barrier_next = ncclShmem.groups[group].barrier_next;
+    barriers = ncclShmem.barrier;
+    if (tid%WARP_SIZE == 0) barriers[threadIdx.x/WARP_SIZE] = 0;
```
cenzhaometa: - 256 threads per CU
- 64 threads per warp
- 2 warps per group 
- 1 group maps to 1 send or 1 recv

barrier was declared as an array of GROUP size (2 in our case), but here we access them based on WARP_ID (4 in our case), won't this lead to out of bounds?
```
barrier[NCCL_MAX_GROUPS];
```

wenkaidu: #define NCCL_MAX_GROUPS (NCCL_MAX_NTHREADS/WARP_SIZE)

src/device/prims_simple.h
```diff
@@ -674,8 +674,9 @@ class Primitives<
 
     // For send operations, we need an extra warp to overlap the threadfence and the copy
     if (tid == 0) ncclShmem.groups[group].warpStart = threadIdx.x/WARP_SIZE;
-    barriers = ncclShmem.groups[group].barrier;
-    barrier_next = ncclShmem.groups[group].barrier_next;
+    barriers = ncclShmem.barrier;
```
cenzhaometa, line 677: prefer naming them as `groupBarriers` to be clear it's group-level barriers?

src/device/prims_simple.h
wenkaidu: NCCL_MAX_GROUPS is max allowed groups, not necessary groups actually used

src/device/prims_simple.h
wenkaidu: sure. can make this change once it is confirmed working

src/device/prims_simple.h
cenzhaometa: never mind, I think `NCCL_MAX_GROUPS` actually means number of warps, we are good.

src/device/primitives.h
```diff
@@ -22,13 +22,13 @@
     const int w = threadIdx.x/WARP_SIZE; \
     const int wid = threadIdx.x%WARP_SIZE; \
     if (wid == 0) { \
-      barrier_next[w] += 1; \
+      barrier_next += 1; \
       barriers[w] += 1; \
```
cenzhaometa, line 27: could you elaborate why we don't need `atomicAdd` here?

is this because we only have 1 writer here? (and 2 readers in our case)

src/device/primitives.h
```diff
@@ -22,13 +22,13 @@
...
       int spins = 0; \
       int rate_limit = 50; \
       for (int i = 0; i < nthreads/WARP_SIZE; i++) { \
         uint8_t warp = ncclShmem.groups[group].warpStart + i; \
-        while (atomicAdd((unsigned long long *)&barriers[warp], 0) < barrier_next[w]) { \
+        while (atomicAdd((unsigned long long *)&barriers[warp], 0) < barrier_next) { \
```
cenzhaometa: shall we cast to uin64_t to be consistent with its declaration?

src/device/primitives.h
cenzhaometa: why we don't need `atomicAdd(&barriers[w], 1)` here? is it because we only have 1 writer? (and 2 readers)

src/device/prims_simple.h
```diff
@@ -668,8 +673,10 @@ class Primitives<
     stepSize(stepSize_ == 0 ? ncclShmem.comm.buffSizes[NCCL_PROTO_SIMPLE]/NCCL_STEPS/sizeof(T) : stepSize_) {
 
     // For send operations, we need an extra warp to overlap the threadfence and the copy
-    barriers = &ncclShmem.groups[group].barrier;
-    barrier_next = ncclShmem.groups[group].barrier_next;
+    if (tid == 0) ncclShmem.groups[group].warpStart = threadIdx.x/WARP_SIZE;
```
cenzhaometa: `threadIdx.x/WARP_SIZE` is the warp_id (0, 1, 2, 3), not the warpStart (0, 2), right? 

e.g thread c0: warp-id = 3, wrap_start = 2.

src/device/prims_ll.h
```diff
@@ -618,7 +618,7 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p>:
     auto *channel = &ncclShmem.channel;
     if (tid == 0) ncclShmem.groups[group].warpStart = threadIdx.x/WARP_SIZE;
     barriers = ncclShmem.barrier;
-    if (tid%WARP_SIZE == 0) barriers[threadIdx.x/WARP_SIZE] = 0;
+    if (tid%WARP_SIZE == 0 && threadIdx.x/WARP_SIZE < NCCL_MAX_GROUPS) barriers[threadIdx.x/WARP_SIZE] = 0;
```
cenzhaometa: 1)
```
barrier[NCCL_MAX_GROUPS];
```
won't original code just lead to out of bound error if `threadIdx.x/WARP_SIZE >= MAX_GOURPS`? 

2)
don't we need atomic_store here too? for setting barriers = 0

PR #1461, opened 2024-12-16
Remove unused highestTransportType

mustafabar:
## Details
___Value of `highestTransportType` was set and passed to `ncclTransportP2pSetup` where the value is output only for the latter function. `highestTransportType` is not used later by the caller___

**Work item:** _"Internal"_

**What were the changes?**  
_Remove the unused variable to improve readability._

_Easier maintenance. highestType is already set inside `ncclTransportP2pSetup`. Setting it outside distorts semantics_

_Passing NULL instead of an unused variable. If the value is to be used in the future, a non-NULL can be passed instead._

**Additional Documentation:**  
_None_

## Approval Checklist


wenkaidu (approved):

corey-derochie-amd:
Is this in the RCCL-only code? If not, is it a necessary change? It could add to future merge conflicts with NCCL.

PR #1476, opened 2025-01-02
Adds API for non-caching-store and load

ranapratap55:
RCCL provides "low latency" protocols for communication between agents, where the entire message consisting of data and flags is packed into a single L2 cache line. This is usually accomplished using atomic relaxed instructions in LLVM. But the 128-byte version of this protocol (LL-128) requires 128-bit load or store instructions that bypass the cache and are not broken up into multiple instructions. The nontemporal builtin is not always suitable for this use case.

The proposed approach is to provide a C++ function template that encapsulates an inline assembly call. This asm is intended to use the appropriate load/store parameters for each combination of data size and architecture.

### DIFF
--- a/src/device/non_caching_load.h
+++ b/src/device/non_caching_load.h
@@ -1,0 +1,77 @@
+#ifndef NON_CACHING_LOAD_H_
+#define NON_CACHING_LOAD_H_
+
+template<typename T>
+inline
+__attribute__((always_inline))
+__host__ __device__ T __non_caching_load(const T* p)
+{
+    #if !defined(__GFX11__) && !defined(GFX12)
+        #define LD "global_load_ubyte"
+        #define LD2 "global_load_ushort"
+        #define LD3 "global_load_dword"
+        #define LD4 "global_load_dwordx2"
+        #define LD5 "global_load_dwordx4"
+        #if defined(__gfx940__) || defined(__gfx941__) || defined(__gfx942__)
+            #define BITS "sc0 sc1 nt"
+        #elif defined(__GFX9__) || defined(__gfx1010__) || defined(__gfx1011__) || defined(__gfx1012__) || defined(__gfx1013__)
+            #define BITS "glc slc"
+        #else
+            #define BITS "glc slc dlc"
+        #endif
+        #define WAIT ((0 << 14) | (0x3f << 8) | (0x7) << 4)
+    #else
+        #define LD "global_load_u8"
+        #define LD2 "global_load_u16"
+        #define LD3 "global_load_b32"
+        #define LD4 "global_load_b64"
+        #define LD5 "global_load_b128"
+        #define BITS "glc slc dlc"
+        #define WAIT ((0 << 10) | (0x3f << 4) | 0x7)
+    #endif
+    #define LOAD LD " %0 %1 off " BITS
+    #define LOAD2 LD2 " %0 %1 off " BITS
+    #define LOAD3 LD3 " %0 %1 off " BITS
+    #define LOAD4 LD4 " %0 %1 off " BITS
+    #define LOAD5 LD5 " %0 %1 off " BITS
+    
+    T r{};
+
+    switch (sizeof(T)) {
+    case 1:
+        asm volatile(LOAD : "={v0}"(r) : "v"(p));
+        break;
+    case 2:
+        asm volatile(LOAD2 : "={v0}"(r) : "v"(p));
+        break;
+    case 4:
+        asm volatile(LOAD3 : "={v0}"(r) : "v"(p));
+        break;
+    case 8:
+        asm volatile(LOAD4 : "={v[0:1]}"(r) : "v"(p));
+        break;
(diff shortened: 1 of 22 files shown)


ranapratap55:
created a new PR with both load and store. Added test cases for byte, 2-byte, 4, 8, 16.

closed the old pr #1289

wenkaidu:
@ranapratap55 can you clarify if the non-caching load and store can work over coarse grained device memory, which is the memory type used in your tests?

ranapratap55:
> @ranapratap55 can you clarify if the non-caching load and store can work over coarse grained device memory, which is the memory type used in your tests?

yes, it works on coarse-grained memory. I have used the hipMalloc() which allocates coarse-grained memory.

wenkaidu:
> > @ranapratap55 can you clarify if the non-caching load and store can work over coarse grained device memory, which is the memory type used in your tests?
> 
> yes, it works on coarse-grained memory. I have used the hipMalloc() which allocates coarse-grained memory.

Can you port and test https://github.com/ROCm/rccl/tree/develop/tools/p2p-latency-test to use the new API, and add the commit to this PR? This is more accurate testing for RCCL use cases. Please try changing memory from "uncached" to coarse-grained memory in these tests.

src/device/non_caching_load.h
```diff
@@ -0,0 +1,77 @@
...
+        #define LD4 "global_load_dwordx2"
+        #define LD5 "global_load_dwordx4"
+        #if defined(__gfx940__) || defined(__gfx941__) || defined(__gfx942__)
+            #define BITS "sc0 sc1 nt"
+        #elif defined(__GFX9__) || defined(__gfx1010__) || defined(__gfx1011__) || defined(__gfx1012__) || defined(__gfx1013__)
+            #define BITS "glc slc"
+        #else
+            #define BITS "glc slc dlc"
+        #endif
+        #define WAIT ((0 << 14) | (0x3f << 8) | (0x7) << 4)
```
mustafabar, line 22: please add a comment where you got the WAIT values from

src/device/non_caching_load.h
```diff
@@ -0,0 +1,77 @@
+#ifndef NON_CACHING_LOAD_H_
+#define NON_CACHING_LOAD_H_
+
+template<typename T>
+inline
+__attribute__((always_inline))
+__host__ __device__ T __non_caching_load(const T* p)
+{
+    #if !defined(__GFX11__) && !defined(GFX12)
```
mustafabar, line 9: There are two families of instructions (e.g `*_u8` vs `*_ubyte`). These need to have some coverage in the tests pointed out by @wenkaidu by targeting 1 arch from each.

tools/non-caching-store/non-caching-store.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    T* h_o1 = (T*)malloc(size);
+    hipMemcpy(h_o1, out1, size, hipMemcpyDeviceToHost);
+
+    hipFree(out1);
+    hipFree(h_o1);
+    return;
+}
+
+int main(int argc, char **argv)
```
mustafabar, line 70: Can you add validation cpp? 
For example, cover memory consistency cases of read-after-write, write-after-write, read-after-read and so on

ranapratap55:
> > > @ranapratap55 can you clarify if the non-caching load and store can work over coarse grained device memory, which is the memory type used in your tests?
> > 
> > 
> > yes, it works on coarse-grained memory. I have used the hipMalloc() which allocates coarse-grained memory.
> 
> Can you port and test https://github.com/ROCm/rccl/tree/develop/tools/p2p-latency-test to use the new API, and add the commit to this PR? This is more accurate testing for RCCL use cases. Please try changing memory from "uncached" to coarse-grained memory in these tests.

the new __non_caching_load template that we provided takes input arguments as 8, 16, 32, 64, and 128byte pointer. It will generate global_load_b128 asm when we provide 128b input.

But https://github.com/ROCm/rccl/tree/develop/tools/p2p-latency-test is using two 64b __builtin_nontemporal_loads() as below

[    i4.v[0] = __builtin_nontemporal_load(src->v);
    i4.v[1] = __builtin_nontemporal_load(src->v+1);](https://github.com/ROCm/rccl/blob/dc75209dd7ad35379f7738417db8a31261c1a281/tools/p2p-latency-test/ll_latency_test.cpp#L68C1-L69C52) 
    
  and converting them to one global_load_b128 which I am unaware of how to port using new template.

wenkaidu:
@ranapratap55 The test code you added appears to be used in kernel launches. CP firmware will flush/invalidate caches around kernel boundary which defeat the purpose of non-caching load/store. You will need to use persistent kernel to test this correctly, such as https://github.com/ROCm/rccl/tree/develop/tools/p2p-latency-test. In these tests, both load and store are issues from persistent kernels which do not launch and exit during the test.

ammallya:
Imported to ROCm/rocm-systems

PR #1477, opened 2025-01-02
Add MSCCLPP user buffer registration APIs and integrate with RCCL

nusislam:
## Details
This PR adds user buffer registration APIs in MSCCLPP and integrates them with the corresponding RCCL APIs.

**What were the changes?**  
A patch has been added in ext-src. This patch will be applied to MSCCLPP during RCCL build. Also MSCCLPP related changes have been added to the RCCL user buffer registration code.

These changes are required to use MSCCLPP in non-HIP-graph mode.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/cmake/MSCCLPP.cmake
+++ b/cmake/MSCCLPP.cmake
@@ -68,20 +68,20 @@ if(ENABLE_MSCCLPP)
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
 	execute_process(
-	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
-	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-	)
-  
-   execute_process(
+            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
+
+        execute_process(
             COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
             WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
-        
-	  execute_process(
-	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
-	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-	  )
-  
+
+        execute_process(
+            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
+
         message(STATUS "Building mscclpp only for gfx942.")
 
         mscclpp_cmake_arg(CMAKE_PREFIX_PATH)
@@ -110,21 +110,21 @@ if(ENABLE_MSCCLPP)
 	execute_process(
            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/cpx.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-  )
-	execute_process(
-	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
-	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-	)
-  
-  execute_process(
+        )
+        execute_process(
+            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
+
+        execute_process(
             COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
             WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-   )
-        
-	execute_process(
-	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
-	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-	)
+        )
+
+        execute_process(
+            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
     endif()
 
     execute_process(COMMAND objcopy
--- a/cmake/MSCCLPP.cmake
+++ b/cmake/MSCCLPP.cmake
@@ -68,9 +68,13 @@ if(ENABLE_MSCCLPP)
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
 	execute_process(
-           COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
-           WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-        )
+	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
+	execute_process(
+	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
 
         message(STATUS "Building mscclpp only for gfx942.")
 
@@ -98,13 +102,18 @@ if(ENABLE_MSCCLPP)
 
         find_package(mscclpp_nccl REQUIRED)
 	execute_process(
-    		COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/cpx.patch
-        	WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-    	)
-	execute_process(
-           COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+           COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/cpx.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
+	execute_process(
+	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
+	execute_process(
+	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
+
     endif()
 
     execute_process(COMMAND objcopy
--- a/ext-src/mem-reg.patch
+++ b/ext-src/mem-reg.patch
@@ -11,7 +11,7 @@ index 7f50792..b8b146d 100644
 + */
 +ncclResult_t ncclCommRegister(ncclComm_t comm, void* buff, size_t size, void** handle);
 +ncclResult_t ncclCommDeregister(ncclComm_t comm, void* handle);
-+
++ncclResult_t ncclBuffIsRegistered(ncclComm_t comm, const void* buff, size_t count, bool* registered);
  /*
   * Send
   *
@@ -39,7 +39,7 @@ index a697be2..1d4af61 100644
  static size_t ncclTypeSize(ncclDataType_t type) {
    switch (type) {
      case ncclInt8:
-@@ -577,6 +583,87 @@ NCCL_API ncclResult_t ncclAllGather(const void* sendbuff, void* recvbuff, size_t
+@@ -577,6 +583,96 @@ NCCL_API ncclResult_t ncclAllGather(const void* sendbuff, void* recvbuff, size_t
    return ncclSuccess;
  }
  
@@ -124,6 +124,15 @@ index a697be2..1d4af61 100644
 +  return ncclSuccess;
 +}
 +
++NCCL_API ncclResult_t ncclBuffIsRegistered(ncclComm_t comm, const void* buff, size_t count, bool* registered){
++  size_t buffBytes;
++  CUdeviceptr buffBasePtr;
++  MSCCLPP_CUTHROW(cuMemGetAddressRange(&buffBasePtr, &buffBytes, (CUdeviceptr)buff));
++  channelKey buffKey{(void*)buffBasePtr, buffBytes};
++  auto buffIt = comm->channelScratchInfos.find(buffKey);
++  *registered =  buffIt != comm->channelScratchInfos.end();
++  return ncclSuccess;
++}
  NCCL_API ncclResult_t ncclSend(const void*, size_t, ncclDataType_t, int, ncclComm_t, cudaStream_t) {
    // TODO: implement this function
    return ncclInternalError;
--- a/ext-src/mem-reg.patch
+++ b/ext-src/mem-reg.patch
@@ -11,15 +11,35 @@ index 7f50792..b8b146d 100644
 + */
 +ncclResult_t ncclCommRegister(ncclComm_t comm, void* buff, size_t size, void** handle);
 +ncclResult_t ncclCommDeregister(ncclComm_t comm, void* handle);
-+ncclResult_t ncclBuffIsRegistered(ncclComm_t comm, const void* buff, size_t count, bool* registered);
++
  /*
   * Send
   *
 diff --git a/apps/nccl/src/nccl.cu b/apps/nccl/src/nccl.cu
-index a697be2..d8497e7 100644
+index a697be2..1d4af61 100644
 --- a/apps/nccl/src/nccl.cu
 +++ b/apps/nccl/src/nccl.cu
-@@ -577,6 +577,67 @@ NCCL_API ncclResult_t ncclAllGather(const void* sendbuff, void* recvbuff, size_t
+@@ -65,6 +65,7 @@ struct ncclComm {
+   std::unordered_map<channelKey, ChannelInfo> channelInInfos;
+   std::unordered_map<channelKey, ChannelInfo> channelOutInfos;
+   std::unordered_map<channelKey, ChannelInfo> channelScratchInfos;
++  std::unordered_map<void*, channelKey> handleKeys;
+   std::shared_ptr<char> scratchBuff;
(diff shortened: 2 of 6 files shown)


ext-src/mem-reg.patch
```diff
@@ -0,0 +1,89 @@
...
++      ChannelInfo channelInfo{outChannels, outChannels, setupSmChannelDeviceHandles(outChannels), setupSmChannelDeviceHandles(outChannels)};
++      recvIt = comm->channelOutInfos.emplace(buffKey, channelInfo).first;
++  }
++  *handle = (void*) buffBasePtr;
++
++  return ncclSuccess;
++}
++
++NCCL_API ncclResult_t ncclCommDeregister(ncclComm_t comm, void* handle) {
++        if (comm && handle) {
```
mustafabar: Shouldn't we free any resources associated with `handle` here?
Example:
```
auto it = comm->channelScratchInfos.find(buffKey);
if (it != comm->channelScratchInfos.end()) {
  comm->channelScratchInfos.erase(it);
}
```
Same thing can be done for channelOut and channelIn

ext-src/mem-reg.patch
```diff
@@ -0,0 +1,89 @@
...
++                handle = nullptr;
++        }
```
mustafabar: The implementation nullifies the `handle` inside the function, but since it's passed as void*, the change doesn't reflect in the caller's context. This can lead to dangling pointers. I suggest passing handle as a void** to resolve this issue, if API change is possible

```suggestion
+NCCL_API ncclResult_t ncclCommDeregister(ncclComm_t comm, void** handle) {
+        if (comm && handle && *handle) {
+                *handle = nullptr;
+        }
```

ext-src/mem-reg.patch
```diff
@@ -0,0 +1,89 @@
...
++  CUdeviceptr buffBasePtr;
++  MSCCLPP_CUTHROW(cuMemGetAddressRange(&buffBasePtr, &buffBytes, (CUdeviceptr)buff));
++
++  int rank = comm->comm->bootstrap()->getRank();
++  channelKey buffKey{(void*)buffBasePtr, buffBytes};
++
++  std::vector<mscclpp::RegisteredMemory> remoteMemories;
++
++  // Creating the channels
++  auto buffIt = comm->channelScratchInfos.find(buffKey);
```
mustafabar, line 58: If there's no guarantee this will be called from a single thread, we probably need to lock access to channelScratchInfos and the rest of the maps. I suggest using one lock if this function doesn't fall on the critical path.

nusislam: This map is supposed to be checked by the host process and MSCCLPP is invoked only when using one GPU per process case.

ext-src/mem-reg.patch
nusislam: This is a NCCL API and cannot be changed. I am still working on some of these fixes.

ext-src/mem-reg.patch
```diff
@@ -11,15 +11,15 @@ index 7f50792..b8b146d 100644
...
++ncclResult_t ncclBuffIsRegistered(ncclComm_t comm, const void* buff, size_t count, bool* registered);
  /*
   * Send
   *
 diff --git a/apps/nccl/src/nccl.cu b/apps/nccl/src/nccl.cu
 index a697be2..d8497e7 100644
 --- a/apps/nccl/src/nccl.cu
 +++ b/apps/nccl/src/nccl.cu
-@@ -577,6 +577,58 @@ NCCL_API ncclResult_t ncclAllGather(const void* sendbuff, void* recvbuff, size_t
+@@ -577,6 +577,67 @@ NCCL_API ncclResult_t ncclAllGather(const void* sendbuff, void* recvbuff, size_t
```
nusislam: This change does not seem to be necessary

src/register.cc
```diff
@@ -155,32 +156,59 @@ NCCL_API(ncclResult_t, ncclCommRegister, const ncclComm_t comm, void* buff, size
...
-  }
-  if (reg->state & COLLNET_REG_COMPLETE) {
-    NCCLCHECK(ncclCollnetDeregBuffer(comm, reg->proxyconn, reg->collnetHandle));
-  }
-  free(reg);
-  memmove(cache->slots+slot, cache->slots+slot+1, (cache->population-slot-1)*sizeof(struct ncclReg*));
-  cache->population -= 1;
+
+  #ifdef ENABLE_MSCCLPP
+    if (comm->mscclCompatible){
```
nusislam: Here you should do the same checks that you did before registering the buffer.
`if (comm->mscclCompatible && size > 0 && (size & 31) == 0 && size <= comm->mscclpp_threshold){`

ext-src/mem-reg.patch
```diff
@@ -0,0 +1,147 @@
...
+ ncclResult_t pncclAllGather(const void* sendbuff, void* recvbuff, size_t sendcount, ncclDataType_t datatype,
+                             ncclComm_t comm, cudaStream_t stream);
+ 
++/*
++ * Register/Deregister
++ */
++ncclResult_t ncclCommRegister(ncclComm_t comm, void* buff, size_t size, void** handle);
++ncclResult_t ncclCommDeregister(ncclComm_t comm, void* handle);
++ncclResult_t ncclBuffIsRegistered(ncclComm_t comm, const void* buff, size_t count, bool* registered);
++size_t ncclBufferSize(ncclComm_t comm, void* handle);
```
corey-derochie-amd: Since you're inserting functions into mscclpp, you can just give them names such as `mscclppCommRegister`, and then there's no need to rename via mscclpp_nccl_syms.txt.

nusislam: `ncclCommRegister()` and `ncclCommDeRegister()` are NCCL APIs. We need to keep these names so that a workload can also be run with the mscclpp backend.

dsidler: I am curious why `ncclCommRegister` and `ncclCommDeRegister` are not implement in mscclpp? The stub functions already exist: https://github.com/microsoft/mscclpp/blob/6d26b92665383b60ffae7e4dd9ab4cc047d34721/apps/nccl/src/nccl.cu#L809C1-L817C2

nusislam:
@corey-derochie-amd - Any more reviews?

isaki001:
There is a typo I need to fix. Will push commit shortly

src/register.cc
```diff
@@ -155,32 +158,60 @@ NCCL_API(ncclResult_t, ncclCommRegister, const ncclComm_t comm, void* buff, size
...
+    if (comm->mscclCompatible && size > 0 && (size & 31) == 0 && size <= comm->mscclpp_threshold){
+      bool isManagedBuffer = false; 
+      CUDACHECK(hipPointerGetAttribute(&isManagedBuffer, HIP_POINTER_ATTRIBUTE_IS_MANAGED, const_cast<void*>(buff)));
+      if(!isManagedBuffer){
+        INFO(NCCL_INIT, "MSCCL++: ncclCommRegister");
+        NCCLCHECK(mscclpp_ncclCommRegister(comm->mscclpp_comm, buff, size, handle));
+      }
+      else{
+        WARN("MSCCL++: Cannot register user-buffers on managed memory");
+      }
```
corey-derochie-amd: This will do nothing, issue a warning, then fall through and return `ncclSuccess`. Is that the right behaviour? Should it not return an error?

isaki001: I think the expectation is that if no registration happens in MSCCL++ due to the buffer being in managed memory, no MSCCL++ or RCCL registration will happen and execution will fall back to RCCL kernel without UBR. 

I actually need to do a small adjustment to not cause a RCCL deregistration error when MSCCL++ detects no registered buffer because of this. I just tested a simple fix for this.

But since we always fallback to RCCL kernel when MSCCL++ is not enabled, I think a warning without return an error is OK. Also, I haven't seen any errors returned in any UBR related routines, except when RCCL cannot find the registration handle during deregistration. Of course, this definitely makes it harder to see if anything goes wrong.

ext-src/mem-reg.patch
corey-derochie-amd: @nusislam Should these changes be submitted as a mscclpp PR?

nusislam: Eventually, yes. But at the moment, will use this patch.

src/register.cc
```diff
@@ -155,32 +158,60 @@ NCCL_API(ncclResult_t, ncclCommRegister, const ncclComm_t comm, void* buff, size
...
+        NCCLCHECK(ncclNvlsDeregBuffer(&reg->mcHandle, reg->regAddr, reg->dev, reg->regSize));
+        reg->regAddr = (CUdeviceptr)NULL;
+      }
+      if (reg->state & COLLNET_REG_COMPLETE) {
+        NCCLCHECK(ncclCollnetDeregBuffer(comm, reg->proxyconn, reg->collnetHandle));
+      }
+      free(reg);
+      memmove(cache->slots+slot, cache->slots+slot+1, (cache->population-slot-1)*sizeof(struct ncclReg*));
+      cache->population -= 1;
+    }
```
corey-derochie-amd:
```suggestion

  #ifdef ENABLE_MSCCLPP
    const size_t size = mscclpp_ncclBufferSize(comm->mscclpp_comm, handle);
    if (comm->mscclCompatible  && size > 0 && (size & 31) == 0 && size <= comm->mscclpp_threshold){
      NCCLCHECK(mscclpp_ncclCommDeregister(comm->mscclpp_comm, handle));
      return ncclSuccess;
    }
  #endif
  
  NCCLCHECK(CommCheck(comm, "ncclCommRegister", "comm"));
  struct ncclReg* reg = (struct ncclReg*)handle;
  struct ncclRegCache* cache = &comm->regCache;
  int slot;
  for (slot=0; slot<cache->population && cache->slots[slot] != reg; slot++);
  if (slot == cache->population) {
    WARN("Deregister: Could not find handle");
    return ncclInvalidUsage;
  }
  if (--reg->refs) return ncclSuccess;
  NCCLCHECK(ncclNetDeregister(comm, reg));
  if (reg->state & NVLS_REG_COMPLETE) {
    NCCLCHECK(ncclNvlsDeregBuffer(&reg->mcHandle, reg->regAddr, reg->dev, reg->regSize));
    reg->regAddr = (CUdeviceptr)NULL;
  }
  if (reg->state & COLLNET_REG_COMPLETE) {
    NCCLCHECK(ncclCollnetDeregBuffer(comm, reg->proxyconn, reg->collnetHandle));
  }
  free(reg);
  memmove(cache->slots+slot, cache->slots+slot+1, (cache->population-slot-1)*sizeof(struct ncclReg*));
  cache->population -= 1;
```
Proposing this to minimize the diff and avoid later merge conflicts.

nusislam:
@isaki001 - I think the helper routines do not have to be named as NCCL APIs; the return type can also be changed. Other than this, the PR looks good to me.

src/register.cc
```diff
@@ -155,32 +158,60 @@ NCCL_API(ncclResult_t, ncclCommRegister, const ncclComm_t comm, void* buff, size
...
+      else{
+        WARN("MSCCL++: Cannot register user-buffers on managed memory");
+      }
+    }
+    else
+    #endif
+    {
+       INFO(NCCL_INIT, "RCCL: ncclCommRegister");
+      NCCLCHECK(ncclRegister(comm, buff, size, handle));
+    }
```
corey-derochie-amd:
```suggestion
  #ifdef ENABLE_MSCCLPP
    if (comm->mscclCompatible && size > 0 && (size & 31) == 0 && size <= comm->mscclpp_threshold){
      bool isManagedBuffer = false; 
      CUDACHECK(hipPointerGetAttribute(&isManagedBuffer, HIP_POINTER_ATTRIBUTE_IS_MANAGED, const_cast<void*>(buff)));
      if(!isManagedBuffer){
        INFO(NCCL_INIT, "MSCCL++: ncclCommRegister");
        NCCLCHECK(mscclpp_ncclCommRegister(comm->mscclpp_comm, buff, size, handle));
      }
      else{
        WARN("MSCCL++: Cannot register user-buffers on managed memory");
      }
      return ncclSuccess;
    }
  #endif
  INFO(NCCL_INIT, "RCCL: ncclCommRegister");
  NCCLCHECK(ncclRegister(comm, buff, size, handle));
```
Proposing this to reduce the diff and avoid future merge conflicts.

cmake/MSCCLPP.cmake
```diff
@@ -68,14 +68,20 @@ if(ENABLE_MSCCLPP)
...
 	execute_process(
-           COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
-           WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-        )
-        execute_process(
+	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
+  
+   execute_process(
```
corey-derochie-amd:
```suggestion
            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
        )

        execute_process(
```
Whitespace suggestions

cmake/MSCCLPP.cmake
```diff
@@ -68,14 +68,20 @@ if(ENABLE_MSCCLPP)
...
             COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
             WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
-
+        
+	  execute_process(
+	   COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	  )
+
```
corey-derochie-amd:
```suggestion

        execute_process(
            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
        )

```
Whitespace suggestions

cmake/MSCCLPP.cmake
```diff
@@ -102,17 +108,23 @@ if(ENABLE_MSCCLPP)
...
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-        )
-        execute_process(
+  )
+	execute_process(
+	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
+  
+  execute_process(
```
corey-derochie-amd:
```suggestion
        )
        execute_process(
            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
        )

        execute_process(
```
Whitespace

cmake/MSCCLPP.cmake
```diff
@@ -102,17 +108,23 @@ if(ENABLE_MSCCLPP)
...
             COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
             WORKING_DIRECTORY ${MSCCLPP_SOURCE}
-        )
+   )
+        
+	execute_process(
+	   COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
+	   WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	)
```
corey-derochie-amd:
```suggestion
        )

        execute_process(
            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mem-reg.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
        )
```

corey-derochie-amd (approved):
Left some suggestions.

src/register.cc
```diff
@@ -155,32 +158,62 @@ NCCL_API(ncclResult_t, ncclCommRegister, const ncclComm_t comm, void* buff, size
...
+        NCCLCHECK(ncclNvlsDeregBuffer(&reg->mcHandle, reg->regAddr, reg->dev, reg->regSize));
+        reg->regAddr = (CUdeviceptr)NULL;
+      }
+      if (reg->state & COLLNET_REG_COMPLETE) {
+        NCCLCHECK(ncclCollnetDeregBuffer(comm, reg->proxyconn, reg->collnetHandle));
+      }
+      free(reg);
+      memmove(cache->slots+slot, cache->slots+slot+1, (cache->population-slot-1)*sizeof(struct ncclReg*));
+      cache->population -= 1;
+    }
```
corey-derochie-amd:
```suggestion

  #ifdef ENABLE_MSCCLPP
    if (comm->mscclCompatible) {
      //if size is zero, registration failed, we don't want RCCL deregistration code to trigger and cause failure
      const size_t size = mscclpp_ncclBufferSize(comm->mscclpp_comm, handle);
      if (size > 0 && (size & 31) == 0 && size <= comm->mscclpp_threshold) {
        NCCLCHECK(mscclpp_ncclCommDeregister(comm->mscclpp_comm, handle));
        return ncclSuccess;
      }
    }
  #endif

  NCCLCHECK(CommCheck(comm, "ncclCommRegister", "comm"));
  struct ncclReg* reg = (struct ncclReg*)handle;
  struct ncclRegCache* cache = &comm->regCache;
  int slot;
  for (slot=0; slot<cache->population && cache->slots[slot] != reg; slot++);
  if (slot == cache->population) {
    WARN("Deregister: Could not find handle");
    return ncclInvalidUsage;
  }
  if (--reg->refs) return ncclSuccess;
  NCCLCHECK(ncclNetDeregister(comm, reg));
  if (reg->state & NVLS_REG_COMPLETE) {
    NCCLCHECK(ncclNvlsDeregBuffer(&reg->mcHandle, reg->regAddr, reg->dev, reg->regSize));
    reg->regAddr = (CUdeviceptr)NULL;
  }
  if (reg->state & COLLNET_REG_COMPLETE) {
    NCCLCHECK(ncclCollnetDeregBuffer(comm, reg->proxyconn, reg->collnetHandle));
  }
  free(reg);
  memmove(cache->slots+slot, cache->slots+slot+1, (cache->population-slot-1)*sizeof(struct ncclReg*));
  cache->population -= 1;
```

isaki001:
Thanks @corey-derochie-amd I'll add your suggestions and I'll change the names as @nusislam indicated.

src/register.cc
```diff
@@ -165,11 +165,11 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
       if(!isManagedBuffer){
         INFO(NCCL_INIT, "MSCCL++: ncclCommRegister");
         NCCLCHECK(mscclpp_ncclCommRegister(comm->mscclpp_comm, buff, size, handle));
+        return ncclSuccess;
```
corey-derochie-amd, line 168: I'm debating whether this is the right behaviour... Are these buffers reused in ways that might call MSCCL++ sometimes and RCCL other times? Are buffers always matched to message size, or could they use a buffer that's larger than the given message size? If so, the safest thing to do might be to register for BOTH RCCL and MSCCL++. Thoughts, @nusislam ?

nusislam: The cases when `mscclppRegister` should be used vs. `rcclRegister` should be used are mutually exclusive.

PR #1483, opened 2025-01-07
[MSCCLPP] IBVerbs: Check if IBV_ACCESS_RELAXED_ORDERING exists

nileshnegi:
## Details

**Work item:**
SWDEV-506051

**What were the changes?**  
When building MSCCLPP, check if `IBV_ACCESS_RELAXED_ORDERING` is defined in `infiniband/verbs.h`. If not, disable symbol use in `ext-src/mscclpp/src/ib.cc`

Some environments (like Debian 10) inherently support older versions of `libibverbs` (e.g. 22.1-1) that do not have `IBV_ACCESS_RELAXED_ORDERING` defined. This can cause a compile-time error when building RCCL with MSCCL++ support.

Patching MSCCLPP's CMakeLists.txt to check if `IBV_ACCESS_RELAXED_ORDERING` is defined in `infiniband/verbs.h`. If not, disable the use of this symbol in MSCCLPP's `src/ib.cc`

## Approval Checklist

### DIFF
--- a/cmake/MSCCLPP.cmake
+++ b/cmake/MSCCLPP.cmake
@@ -71,6 +71,10 @@ if(ENABLE_MSCCLPP)
            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
+        execute_process(
+            COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
 
         message(STATUS "Building mscclpp only for gfx942.")
 
@@ -104,6 +108,10 @@ if(ENABLE_MSCCLPP)
 	execute_process(
            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/read-allred.patch
            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+        )
+        execute_process(
+            COMMAND git apply --reverse ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/mscclpp_ibv_access_relaxed_ordering.patch
+            WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
     endif()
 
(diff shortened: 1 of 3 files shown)


ext-src/mscclpp_ibv_access_relaxed_ordering.patch
```diff
@@ -0,0 +1,33 @@
...
++++ b/CMakeLists.txt
+@@ -96,6 +96,28 @@ include(${PROJECT_SOURCE_DIR}/cmake/AddFormatTargets.cmake)
+ 
+ # Find ibverbs and libnuma
+ find_package(IBVerbs)
++
++# Check if IBV_ACCESS_RELAXED_ORDERING exists in infiniband/verbs.h
++# Use internal IBVerbs definitions if it does not exist
++if(IBVERBS_FOUND)
++  try_compile(HAS_IBV_ACCESS_RELAXED_ORDERING
```
nusislam, line 13: This is a performance parameter. Disabling IB does not seem appropriate if this is not there.

nileshnegi: then would you suggest removing this parameter from [mscclpp/src/ib.cc](https://github.com/microsoft/mscclpp/blob/1e82dd444fc1ed8b7add354eebaab8a94e67d5fc/src/ib.cc#L53C1-L54C1)?

wenkaidu (approved):

cmake/MSCCLPP.cmake
```diff
@@ -63,13 +63,17 @@ if(ENABLE_MSCCLPP)
                 WORKING_DIRECTORY ${MSCCLPP_SOURCE}
             )
         endif()
-        execute_process(
-           COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/cpx.patch
-           WORKING_DIRECTORY ${MSCCLPP_SOURCE}
+	execute_process(
+            COMMAND git apply -v --recount ${CMAKE_CURRENT_SOURCE_DIR}/ext-src/cpx.patch
```
nusislam: Can you remove these changes? It might cause conflict with the other mscclpp that I am working on.

ext-src/mscclpp_ibv_access_relaxed_ordering.patch
```diff
@@ -0,0 +1,51 @@
...
++++ b/CMakeLists.txt
+@@ -96,6 +96,24 @@ include(${PROJECT_SOURCE_DIR}/cmake/AddFormatTargets.cmake)
+ 
+ # Find ibverbs and libnuma
+ find_package(IBVerbs)
++
++# Check if IBV_ACCESS_RELAXED_ORDERING exists in infiniband/verbs.h
++# Disable use of this symbol in mscclpp/src/ib.cc if it does not exist
++if(IBVERBS_FOUND)
++  try_compile(HAS_IBV_ACCESS_RELAXED_ORDERING
```
nusislam, line 13: Do require a new compile flag now?

nusislam (approved):

PR #1503, opened 2025-01-23
non-hipGraph MSCCL++ tests for allReduce and allGather

isaki001:
## Details

**What were the changes?**  
Added functional test for allGather and allReduce when utilizing MSCCL++ kernels in non-hipGraph mode, with/without managed memory.

No test for non-hipGraph mode user-buffer registration.

TestBed infrastructure was encountering a hang. As such, I added a simple routine that creates 8 process through fork(), and calls allReduce/allGather. 

**Additional Documentation:**  

## Approval Checklist


test/AllReduceTests.cpp
```diff
@@ -242,4 +242,36 @@ namespace RcclUnitTesting
...
+    for (int i = 0; i < count; ++i){
+        sendBuff[i] = i;
+        expected[i] = i * nranks;
+    }
+    fork_and_launch_rccl(nranks, ncclCollAllReduce, sendBuff, recvBuff, expected);
+  }
+
+  TEST(AllReduce, ManagedMemUserBufferRegistration)
+  {          
+    setenv("UT_PROCESS_MASK", "2", 1);
```
corey-derochie-amd: I'm a bit uncomfortable setting the process mask in the test, as this overrides the intentions of the test runner. I think if this test only succeeds for multi-process, we should let it fail for single process and the runner should know not to run it for single process. Or you could check the process mask and skip the test unless it's multi process. Thoughts?

src/register.cc
```diff
@@ -172,6 +172,7 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
       else{
         WARN("MSCCL++: Cannot register user-buffers on managed memory. RCCL user-buffer registration will occur.");
       }
+      return ncclSuccess;
```
corey-derochie-amd: We were originally back and forth on whether to exit the function here. Why is this needed now?

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -434,6 +434,7 @@ static ncclResult_t mscclFallBackSavedParams() {
           param.p.root, param.comm, param.stream));
         break;
       case mscclFuncAllReduce:
+          INFO(NCCL_INIT, "fallback ncclReduce");
```
corey-derochie-amd: The ncclAllReduce function already logs itself to the INFO COLL log, so these are not needed.

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -442,6 +443,7 @@ static ncclResult_t mscclFallBackSavedParams() {
           param.p.op, param.comm, param.stream));
         break;
       case mscclFuncAllGather:
+        INFO(NCCL_INIT, "fallback ncclAllGather");
```
corey-derochie-amd: Not needed.

corey-derochie-amd (dismissed):
Please give the PR a more descriptive title, as this will become the commit message.

src/register.cc
isaki001: It's not needed here. I was testing and didn't realize this made it in the commit.

test/AllReduceTests.cpp
isaki001: I was originally enabling this through an environment variable, however I thought this was not required as it complicates the launching of the tests. I thought that this test should be run as any other, without needing explicit variables. I also thought this could have implications if the CI used it, where we wouldn't want failures. But if the CI executes the test appropriately, I don't see any problem with remove this.

corey-derochie-amd: Upon further investigation, I think setting this does nothing because this env is parsed at the beginning of the test process and not read again, and the parsed result is only used in TestBed, so setting it here will have no effect.

These should be removed.

corey-derochie-amd (changes requested):

test/common/StandaloneUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+        call_rccl(id, collID, r, nranks, sendBuff, recvBuff, use_managed_mem);
+        for(int i = 0; i < recvBuff.size(); ++i)
+          ASSERT_EQ(recvBuff[i], expected[i]);
+        break;
+      }
+    }
+
+    getAndDistributeNCCLid(nranks);
+    for(int r = 0; r < nranks; ++r)
+      wait(NULL); // Wait for all children
```
corey-derochie-amd: Will this cause all child processes to `getAndDistributeNCCLid` and `wait` after they are done with `call_rccl`?

test/AllGatherTests.cpp
```diff
@@ -120,4 +121,36 @@ namespace RcclUnitTesting
...
+    size_t count = 2048;
+    std::vector<int> sendBuff(count, 0);
+    std::vector<int> recvBuff(count, 0);
+    std::vector<int> expected(count, 0);
+
+    for (int i = 0; i < count; ++i){
+        sendBuff[i] = i;
+        expected[i] = i * nranks;
+    }
+    fork_and_launch_rccl(nranks, ncclCollAllGather, sendBuff, recvBuff, expected);
```
corey-derochie-amd: What is the root cause of the hang? That's not expected so we should really solve that instead of introducing a new test pattern that forks.

test/common/StandaloneUtils.cpp
isaki001: Yeah, I think you are right. The intention was for the parent process to do that. But the child process would get there too.

test/AllGatherTests.cpp
isaki001: We dont' know what was causing it. It seemed like the hang was in ncclCommRegister. The plan was to solve the issue with the TestBed test, and to push a simple working test before that.

test/AllGatherTests.cpp
```diff
@@ -120,4 +121,42 @@ namespace RcclUnitTesting
...
+    for (int i = 0; i < count; ++i){
+        sendBuff[i] = i;
+    }
+
+    for(int r = 0; r < nranks; ++r)
+      for (int i = 0; i < count; ++i)
+        expected[r*count + i] = sendBuff[i];
+
+    fork_and_launch_rccl(nranks, ncclCollAllGather, sendBuff, recvBuff, expected, use_managed_mem);
+  }
```
corey-derochie-amd: If these tests will fail for single-process, there should be a comment documenting such and logic to skip the test in that condition, I think?

test/AllReduceTests.cpp
```diff
@@ -242,4 +242,34 @@ namespace RcclUnitTesting
...
+    std::vector<int> sendBuff(count, 0);
+    std::vector<int> recvBuff(count, 0);
+    std::vector<int> expected(count, 0);
+    const bool use_managed_mem = true;
+    for (int i = 0; i < count; ++i){
+        sendBuff[i] = i;
+        expected[i] = i * nranks;
+    }
+    fork_and_launch_rccl(nranks, ncclCollAllReduce, sendBuff, recvBuff, expected, use_managed_mem);
+  }
```
corey-derochie-amd: If these tests will fail for single-process, there should be a comment documenting such and logic to skip the test in that condition.

test/AllGatherTests.cpp
isaki001: Could I detect based on reading UT_PROCESS_MASK?

corey-derochie-amd: Nevermind, I forgot the previous work, this isn't necessary.

corey-derochie-amd (dismissed):

test/common/TestBed.cpp
```diff
@@ -47,6 +47,9 @@
 
 namespace RcclUnitTesting
 {
+
+   
+
```
corey-derochie-amd:
```suggestion
```

test/common/TestBed.hpp
```diff
@@ -13,6 +13,26 @@
...
+
+#define NCCLCHECK(cmd) do {                                     \
+    ncclResult_t res = cmd;                                     \
+    if (res != ncclSuccess) {                                   \
+         printf("NCCL failure %s:%d '%s'\n",                    \
+            __FILE__,__LINE__,ncclGetErrorString(res));         \
+    }                                                           \
+} while(0)
+
+  //void call_RCCL(ncclUniqueId id, int collID, int rank, int nranks, std::vector<int>& send, std::vector<int>& recv, bool managed = false);
```
corey-derochie-amd: None of this is needed.
```suggestion
```

test/common/CallCollectiveForked.cpp
```diff
@@ -0,0 +1,164 @@
...
+    if (res != ncclSuccess) {                                   \
+         printf("NCCL failure %s:%d '%s'\n",                    \
+            __FILE__,__LINE__,ncclGetErrorString(res));         \
+    }                                                           \
+} while(0)
+
+namespace RcclUnitTesting
+{
+
+void call_rccl(ncclUniqueId id, int collID, int rank, int nranks, const std::vector<int>& send, std::vector<int>& recv, bool use_managed_mem){
```
corey-derochie-amd:
```suggestion
void callCollective(ncclUniqueId id, int collID, int rank, int nranks, const std::vector<int>& send, std::vector<int>& recv, bool use_managed_mem){
```

test/common/CallCollectiveForked.cpp
```diff
@@ -0,0 +1,164 @@
...
+        if(ngpus != nranks){
+          exit(0);
+        }
+        //child processes
+        if(r == 0)
+          createNCCLid(r);
+        else
+          getNCCLidFromParent(r);
+
+        call_rccl(id, collID, r, nranks, sendBuff, recvBuff, use_managed_mem);
```
corey-derochie-amd:
```suggestion
        callCollective(id, collID, r, nranks, sendBuff, recvBuff, use_managed_mem);
```

test/common/CallCollectiveForked.cpp
```diff
@@ -0,0 +1,164 @@
...
+{
+
+void call_rccl(ncclUniqueId id, int collID, int rank, int nranks, const std::vector<int>& send, std::vector<int>& recv, bool use_managed_mem){
+    switch(collID){
+        case ncclCollAllReduce:
+        case ncclCollAllGather:
+        break;
+
+        default:
+            ERROR("This collective is not implemented for call_RCCL routine");
```
corey-derochie-amd:
```suggestion
            ERROR("This collective is not implemented for callCollective routine");
```

corey-derochie-amd (approved):
@isaki001 I'd suggest renaming `call_rccl` but other than that this is good to go. The failed CI is not related to your changes.

PR #1505, opened 2025-01-24
Disabled MSCCL++ feature except when building on Ubuntu or CentOS host systems

corey-derochie-amd:
## Details
___Do not mention proprietary info or link to internal work items in this PR.___

**Work item:** _"Internal", or link to GitHub issue (if applicable)._

**What were the changes?**  
Added condition for MSCCL++ to only build on an Ubuntu or CentOS host system. Sets `ENABLE_MSCCLPP` flag to `OFF`.

**Why were the changes made?**  
Microsoft has said mscclpp is only supported on Ubuntu, so this removes unsupported configurations. CentOS has also been tested, so we will support it as well.

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #1507, opened 2025-01-25
[MSCCLPP] Use boost/filesystem on SLES15

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Use `boost/filesystem` instead of C++ 17 `filesystem` in MSCCLPP on SLES 15.

SLES 15.5 does not support C++ 17, so we cannot use `filesystem` methods. As an alternative, `boost/filesystem` can be used, but some modifications to MSCCLPP src are required.

Apply `sles-cpp-filesystem.patch` to MSCCLPP src if CMake determines that the OS is SLES.

## Approval Checklist

### DIFF
--- a/cmake/MSCCLPP.cmake
+++ b/cmake/MSCCLPP.cmake
@@ -85,6 +85,14 @@ if(ENABLE_MSCCLPP)
             WORKING_DIRECTORY ${MSCCLPP_SOURCE}
         )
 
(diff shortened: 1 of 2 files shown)


edgargabriel:
I don't think the statement that SLES 15.5 does not support std::filesystem is correct, this is probably more a property of the compiler (version) used.

nileshnegi:
Opting for https://github.com/ROCm/rccl/pull/1505

PR #1516, opened 2025-01-30
replacing rccl_float8 with hip_fp8

mberenjk:
## Details

**Work item**: [LWPCOMMLIBS-576]
**What were the changes?** 
move to HIP FP8 header file

RCCL needs to switch over to HIP FP8 header file which has become availiable in recent ROCm releases. 

RCCL-UT and rccl-tests have been tested and passed with datatype=FP8.
**Additional Documentation:**  
We still want to keep current RCCL internal header file in case RCCL is built on older ROCm releases which has no HIP FP8 header.


wenkaidu (approved):

ext-src/mscclpp
corey-derochie-amd, line 1: Why was mscclpp updated?

mberenjk: The mscclpp submodule is pointing to an older commit. There are no changes to mscclpp in this PR.

corey-derochie-amd: The commit is fixed at an older commit on purpose, was updating necessary for these changes?

corey-derochie-amd (changes requested):
I think there are necessary changes missing in nccl.h.in, please stand by while I find them.

mberenjk:
Creating a new pull request to add compatibility with the gfx942 architecture.

PR #1517, opened 2025-01-31
Make rccl version detection robust.

rahulc-gh:
* Accept an EXPLICIT_ROCM_VERSION and use that vs inspecting the environment if provided.
* Use CMake's built in file reading support vs execute_process (without error checking) to avoid silent but deadly later failures.
* Properly quote some comparisons to avoid syntax errors if they happen to have an empty string.
* Guard against ROCM_PATH being an empty string, avoiding stray path extensions to root directories, etc.

## Details

**What were the changes?**  

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,10 +135,12 @@ endif()
 set(CMAKE_INSTALL_PREFIX "${ROCM_PATH}" CACHE PATH "")
 set(CMAKE_CXX_STANDARD   14)   # We use C++14 features, this will add compile option: -std=c++14
 set(CMAKE_CXX_EXTENSIONS OFF)  # Without this line, it will add -std=gnu++14 instead, which has some issues.
-list(APPEND CMAKE_PREFIX_PATH  # Add ROCM_PATH to CMake search paths (for finding HIP / HSA
-            ${ROCM_PATH}
-            ${ROCM_PATH}/hip
-            ${ROCM_PATH}/llvm)
+if(ROCM_PATH)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -135,10 +135,12 @@ endif()
 set(CMAKE_INSTALL_PREFIX "${ROCM_PATH}" CACHE PATH "")
 set(CMAKE_CXX_STANDARD   14)   # We use C++14 features, this will add compile option: -std=c++14
 set(CMAKE_CXX_EXTENSIONS OFF)  # Without this line, it will add -std=gnu++14 instead, which has some issues.
-list(APPEND CMAKE_PREFIX_PATH  # Add ROCM_PATH to CMake search paths (for finding HIP / HSA
-            ${ROCM_PATH}
-            ${ROCM_PATH}/hip
-            ${ROCM_PATH}/llvm)
+if(ROCM_PATH)
```
nileshnegi, line 138: we define/check for `ROCM_PATH` in [toolchain-linux.cmake](https://github.com/ROCm/rccl/blob/develop/toolchain-linux.cmake). Is this check still needed?

stellaraccident: Looks like you are defaulting the cache variable there (different topic, but we need to remove all of this shaky dependence on /opt/rocm codebase wide at some point -- it is really a bad idea). But it is still valid for it to be configured as an empty string (which disables lookups into a full SDK). So yes, we should be conditioning anything that is setting paths based on ROCM_PATH to only do it if not empty.

corey-derochie-amd (approved):
The valgrind test isn't working right now, ignoring the test failure.

Issue #1509, opened 2025-01-28
set of GPU_TARGETS as CACHE variable silently latches prior config


stellaraccident:
Line 92 in CMakeLists.txt: `set(GPU_TARGETS "${SUPPORTED_GPUS}" CACHE STRING "GPU targets to compile for.")` causes `GPU_TARGETS` to silently latch whatever the first `SUPPORTED_GPUS` is (which is derived from the "public" option `AMDGPU_TARGETS`.

Repro:

1. Build rccl with `-DAMDGPU_TARGETS=gfx942`
2. Reconfigure with `-DAMDGPU_TARGETS=gfx1100`
3. Observe that the "Compiling for" status message does not change.

nileshnegi:
Is there a reason for not removing existing CMakeCache.txt and other files in the build directory if you are reconfiguring?

stellaraccident:
It is very common to be able to fiddle with CMake options in the same build dir and I generally consider it a P0 that the project a) should work for "reasonable" config changes, and b) [even worse and what is happening here] not latch and do the wrong thing silently. Call it baseline expectations of a CMake project. When I run across something that is doing what this is, I generally consider that the project isn't very mature.

stellaraccident:
(in this case, this cost me 15 minutes because the outer build system had a bug that passed an incorrect list of targets and the fault in this project meant the cache was poisoned. Generally, when I run into a footgun like this, I at least let someone know -- which is what I am doing here... since one person wasting time and finding the solution often means that others are just wondering around aimlessly in the woods)

sohaibnd:
@stellaraccident This should be fixed now with https://github.com/ROCm/rccl/pull/1533, let me know if you have any follow-up concerns otherwise I can close this issue.

stellaraccident:
Thanks. When can I expect this on mainline?

sohaibnd:
It should be in mainline soon but I can't provide more information publicly.

PR #1533, opened 2025-02-05
Remove COMPILING_TARGETS from CMakeLists.txt

sohaibnd:
COMPILING_TARGETS is not actually used for --offload-arch option, instead GPU_TARGETS is being used implicitly when we call
find_package(hip REQUIRED) (See hip-config-amd.cmake)

## Details

**Work item:** https://github.com/ROCm/rccl/issues/1509

**What were the changes?**  
Remove COMPILING_TARGETS from CMakeLists.txt

COMPILING_TARGETS is not actually used for --offload-arch option, instead GPU_TARGETS is being used implicitly when we call
find_package(hip REQUIRED) (See hip-config-amd.cmake)

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ if (BUILD_ADDRESS_SANITIZER)
       endif()
     endif()
   endforeach()
-  SET(GPU_TARGETS "${amdgpu_targets}" CACHE STRING "Modified GPU list for Address-Sanitizer enabled build." FORCE)
+  SET(GPU_TARGETS "${amdgpu_targets}")
 endif()
 
 # Check if clang compiler can offload to GPU_TARGETS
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -83,7 +85,7 @@ if (BUILD_ADDRESS_SANITIZER)
       endif()
     endif()
   endforeach()
-  SET(GPU_TARGETS "${amdgpu_targets}" CACHE STRING "Modified GPU list for Address-Sanitizer enabled build." FORCE)
+  SET(GPU_TARGETS "${amdgpu_targets}")
```
nileshnegi, line 86: we need FORCE here to override the default list of GPU_TARGETS with `xnack+` appended values if not specified

sohaibnd: I've changed it to a normal variable instead of a cache variable (CACHE keyword removed) which does't need FORCE (reference: https://cmake.org/cmake/help/latest/command/set.html). Another advantage is that since normal variables won't be cached, if the user specifies BUILD_ADDRESS_SANITIZER OFF after previously specifying ON, the xnack+ will no longer be appended which is what you expect.

nileshnegi: does not work as intended

sohaibnd: There is a set(GPU_TARGETS...CACHE) which gets called in hip-config-amd.cmake and leads to the normal variable being unset because of CMP0126, so I changed line 71 back to a set(CACHE), it should work now. I also got rid of COMPILING_TARGETS since it is not actually used to set --offload-arch option.

nileshnegi (approved):

PR #1537, opened 2025-02-06
Create internal thread to dump proxy information at given interval.

thananon:
## Details
This PR creates a helper thread to dump out proxy operations at every given interval. This is useful to debug hangs in multi-node scenario. The thread is not created by default. It is controlled via param `RCCL_PROXY_MONITOR_INTERVAL`. If set to > 0, the thread will attempt to dump proxy info at that interval (seconds).

This should have minimal/no impact to the default mode of operation.

image

**What were the changes?**  

More debugging tool for developers.

Create a helper thread that dump out information.

## Approval Checklist

### DIFF
--- a/src/include/proxy.h
+++ b/src/include/proxy.h
@@ -74,8 +74,7 @@ struct ncclProxyOp {
   uint8_t* sendbuff;
   uint8_t* recvbuff;
 
-  int nextRank;
-  int prevRank;
+  int peer;
 
   union ncclProxyOpSpecifics specifics;
 
@@ -149,8 +148,7 @@ struct ncclProxyArgs {
 
   union ncclProxyOpSpecifics specifics;
 
-  int prevRank;
-  int nextRank;
+  int peer;
   int send;
   int retry_total;
 };
--- a/src/include/proxy.h
+++ b/src/include/proxy.h
@@ -76,6 +76,8 @@ struct ncclProxyOp {
 
   int peer;
   int rank;
+  uint64_t tail;
+  uint64_t gputail;
 
   union ncclProxyOpSpecifics specifics;
 
@@ -152,6 +154,8 @@ struct ncclProxyArgs {
   int peer;
   int rank;
   int send;
+  uint64_t tail;
+  uint64_t gputail;
   int retry_total;
 };
 #define NCCL_MAX_NETDEVS 128
--- a/src/include/proxy.h
+++ b/src/include/proxy.h
@@ -75,6 +75,7 @@ struct ncclProxyOp {
   uint8_t* recvbuff;
 
   int peer;
+  int rank;
 
   union ncclProxyOpSpecifics specifics;
 
@@ -149,6 +150,7 @@ struct ncclProxyArgs {
   union ncclProxyOpSpecifics specifics;
 
   int peer;
+  int rank;
   int send;
   int retry_total;
 };
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -21,6 +21,10 @@
 #include <unistd.h>
 #include <sys/time.h>
 
+static ncclProxyProgressState* ncclLastProxyState;
+static pthread_t proxyMonitorThread;
+int proxyMonitorInit = 0;
+
 static bool NeedProxy(int type, int pattern, int root, struct ncclRing* ring, int nranks) {
   if (pattern == ncclPatternRing || pattern == ncclPatternRingTwice) return true;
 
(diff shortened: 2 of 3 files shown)


src/proxy.cc
```diff
@@ -269,7 +269,7 @@ ncclResult_t printProxyOp(struct ncclProxyArgs* op, int poolIndex, int opIndex)
 	// Send or recv within a collective. Dump raw state data.
 	fprintf(stderr, " nb:%zd ns:%d p:%lu t:%lu r:%lu, d:%lu ",sub->nbytes,sub->nsteps, sub->posted, sub->transmitted, sub->received, sub->done);
       }
-      fprintf(stderr, "%c peer:%d chan:%d ", status, peer, sub->channelId);
+      fprintf(stderr, "%c peer:%d chan:%d myrank:%d", status, peer, sub->channelId, op->rank);
```
ryanhankins: It would be clearer to me to stick myrank before peer rather than at the end.

src/proxy.cc
```diff
@@ -916,6 +919,38 @@ ncclResult_t ncclProxyProgressDestroy(struct ncclProxyState* proxyState) {
...
+void *ncclProxyMonitor(void *args) {
+  int interval = *(int*) args;
+  int elapsed = 0;
+  while (proxyMonitorInit) {
+    if (elapsed > interval) {
+      dumpProxyState(ncclLastProxyState);
+      elapsed = 0;
+    }
+    sleep(3);
+    elapsed += 3;
```
ryanhankins: I'd make this env-configurable instead of 3.

src/proxy.cc
```diff
@@ -935,7 +935,7 @@ void *ncclProxyMonitor(void *args) {
 
 static ncclResult_t ncclProxyMonitorCreate() {
   int *interval = (int*) malloc(sizeof(int));
-  *interval = 1;
+  *interval = 60;
```
ryanhankins: Would help to make this configurable.

src/proxy.cc
```diff
@@ -919,32 +924,38 @@ ncclResult_t ncclProxyProgressDestroy(struct ncclProxyState* proxyState) {
   return ncclSuccess;
 }
 
-void *ncclProxyMonitor(void *args) {
+void *rcclProxyMonitor(void *args) {
   int interval = *(int*) args;
   int elapsed = 0;
+
```
ryanhankins, line 930: NIT: extra line.

src/proxy.cc
```diff
@@ -21,9 +21,12 @@
 #include <unistd.h>
 #include <sys/time.h>
 
+/* RCCL Proxy Monitor */
+RCCL_PARAM(ProxyMonitorInterval, "PROXY_MONITOR_INTERVAL", 0);
 static ncclProxyProgressState* ncclLastProxyState;
 static pthread_t proxyMonitorThread;
 int proxyMonitorInit = 0;
+/**/
```
ryanhankins, line 29: Extra comment.

src/include/proxy.h
```diff
@@ -76,6 +76,8 @@ struct ncclProxyOp {
 
   int peer;
   int rank;
+  uint64_t tail;
```
ryanhankins, line 79: I'd add a comment here to say that these are for proxy dump.  Otherwise I'm afraid they'll get hijacked for some other purpose and misused.  Same in ncclProxy Args.

ryanhankins (approved):

src/proxy.cc
thananon: I think you are looking at the older commit? This is fully configurable at the current one.

src/proxy.cc
```diff
@@ -21,6 +21,13 @@
 #include <unistd.h>
 #include <sys/time.h>
 
+/* RCCL Proxy Monitor */
+RCCL_PARAM(ProxyMonitorInterval, "PROXY_MONITOR_INTERVAL", 0);
+static ncclProxyProgressState* ncclLastProxyState;
+static pthread_t proxyMonitorThread;
+int proxyMonitorInit = 0;
```
mustafabar, line 28: How about making it `static` if not used outside `proxy.cc`?

src/proxy.cc
```diff
@@ -919,6 +924,44 @@ ncclResult_t ncclProxyProgressDestroy(struct ncclProxyState* proxyState) {
...
+  return NULL;
+}
+
+static ncclResult_t rcclProxyMonitorCreate() {
+  if(rcclParamProxyMonitorInterval() <= 0) return ncclSuccess;
+  if(proxyMonitorInit == 1) return ncclSuccess;
+
+  int *interval = (int*) malloc(sizeof(int));
+
+  *interval = rcclParamProxyMonitorInterval();
```
mustafabar, line 950: can you just read `rcclParamProxyMonitorInterval` as a local var inside `proxyMonitorThread`? I didn't see the need to pass it as arg

PR #1538, opened 2025-02-07
replacing rccl_float8 with hip_fp8 and address compatibility issue

mberenjk:
## Details

**Work item:** _ [LWPCOMMLIBS-576]

**What were the changes?**  
move to HIP FP8 header file 

RCCL needs to switch over to HIP FP8 header file which has become available in recent ROCm releases.

RCCL-UT and rccl-tests have been tested and passed with datatype=FP8.

**Additional Documentation:**  
For gfx942, we need to support FP8_FNUZ, as the OCP type is currently unsupported. Additionally, we should retain the current RCCL internal header file to ensure compatibility with older ROCm releases that lack the HIP FP8 header. 
## Approval Checklist

### DIFF
--- a/src/include/rccl_float8.h
+++ b/src/include/rccl_float8.h
@@ -38,7 +38,9 @@ typedef struct
     uint8_t data;
 } rccl_bfloat8;
 
-#elif ROCM_VERSION >= 60300 // __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
+// __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
+#elif ROCM_VERSION >= 60300
+
 #include <hip/hip_fp8.h>
 
 #if HIP_FP8_TYPE_OCP
@@ -78,9 +80,9 @@ inline __host__ __device__ float operator*(rccl_bfloat8 a, float b)
 {
     return float(a) * float(b);
 }
+
 // For older versions of ROCm that do not include hip_fp8.h,
 // we provide a local version of the header file as a fallback.
-
 #else
 
 #define HIP_HOST_DEVICE __host__ __device__
@@ -1059,6 +1061,6 @@ inline __host__ __device__ T explicit_downcast(Ta a, uint32_t rng)
 
 // =================================================================================================
 
-#endif // __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
+#endif
 
 #endif // ROCBLAS_FLOAT8_H
--- a/src/include/rccl_float8.h
+++ b/src/include/rccl_float8.h
@@ -25,7 +25,7 @@
 
 #include <stdint.h>
 
-#if __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
+#if __cplusplus < 201103L || !defined(__HIP__)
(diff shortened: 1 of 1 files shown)


src/include/rccl_float8.h
```diff
@@ -38,7 +38,50 @@ typedef struct
     uint8_t data;
 } rccl_bfloat8;
 
-#else // __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
+#elif ROCM_VERSION >= 60300 // __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
```
corey-derochie-amd: Looks like `__hip_fp8_*_fnuz` exists in ROCm 6.2.x (no OCP types). Perhaps we should add a condition for that here as well.

src/include/rccl_float8.h
```diff
@@ -25,7 +25,7 @@
 
 #include <stdint.h>
 
-#if __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
+#if __cplusplus < 201103L || !defined(__HIP__)
```
corey-derochie-amd: I don't think this change is actually necessary, but I also think that this header was one place where we missed swapping these HIP defines... is `__HCC__` and `__HIPCC__` still valid. @nileshnegi might know that.

wenkaidu (approved):

src/include/rccl_float8.h
```diff
@@ -1016,6 +1059,6 @@ inline __host__ __device__ T explicit_downcast(Ta a, uint32_t rng)
 
 // =================================================================================================
 
-#endif // __cplusplus < 201103L || (!defined(__HCC__) && !defined(__HIPCC__))
+#endif
```
corey-derochie-amd, line 1064: Let's ditch this and change it if we change the logic above.
```suggestion
#endif // __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
```

src/include/rccl_float8.h
```diff
@@ -1016,6 +1059,6 @@ inline __host__ __device__ T explicit_downcast(Ta a, uint32_t rng)
...
 
-#endif // ROCBLAS_FLOAT8_H
+#endif  // ROCBLAS_FLOAT8_H
```
corey-derochie-amd:
```suggestion
#endif // ROCBLAS_FLOAT8_H
```

src/include/rccl_float8.h
nileshnegi: we use `(!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))` in RCCL

`defined(__HIP_PLATFORM_AMD__)` is the norm now, but `defined(__HIPCC__)` is still needed for backward compatibility... we can remove `defined(__HCC__)`

corey-derochie-amd:
```suggestion
#if __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
```

src/include/rccl_float8.h
mberenjk: Tested hip_fp8 from ROCm 6.2 on MI300 and MI250; confirmed it works fine. Also verified compatibility with ROCm 6.1 to 6.4 without build failures.

PR #1541, opened 2025-02-10
enable 224 channels

isaki001:
## Details

**What were the changes?**  
Enabled 256 channels

RCCL cannot go past 128 channels

Due to NCCL 2.22 sync, it was not enough to change MAXCHANNELS, as division by zero erros occured during enqueue; I had to set nMaxChannels in struct ncclTaskColl to use 16 bits instead of 8. Then, I had to also adjust the max-allowed kernel sizes since budget restrictions introduced in NCCL 2.22 were causing a hang when utilizing 256 channels.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/include/comm.h
+++ b/src/include/comm.h
@@ -222,7 +222,7 @@ struct ncclTaskColl {
   int chunkSteps, sliceSteps;
   // Computed later:
   size_t trafficBytes;
-  int32_t nMaxChannels:8;
+  int32_t nMaxChannels:16;
   int32_t nWarps:8;
   int32_t algorithm:8, protocol:8;
   uint32_t isCollnet:1, isNvls:1;
--- a/src/include/comm.h
+++ b/src/include/comm.h
@@ -222,7 +222,7 @@ struct ncclTaskColl {
   int chunkSteps, sliceSteps;
   // Computed later:
   size_t trafficBytes;
-  int32_t nMaxChannels:16;
+  int32_t nMaxChannels:8;
   int32_t nWarps:8;
   int32_t algorithm:8, protocol:8;
   uint32_t isCollnet:1, isNvls:1;
--- a/src/include/device.h
+++ b/src/include/device.h
@@ -68,7 +68,7 @@ union ncclLLFifoLine {
 };
 
 #define WARP_SIZE warpSize
-#define MAXCHANNELS 256
+#define MAXCHANNELS 224
 #define CHANNEL_LIMIT 16
 #define NCCL_MAX_LOCAL_RANKS 64
 #define NCCL_MAX_NTHREADS 256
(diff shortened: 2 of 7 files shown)


wenkaidu:
Can you find out what caused the kern arg size to go over 4K? I would rather keep 4K which is exactly one page. "5K" may have unintended consequences. Please check with @BertanDogancay  if there is an alternative method to reduce kern arg size.

isaki001:
> Can you find out what caused the kern arg size to go over 4K? I would rather keep 4K which is exactly one page. "5K" may have unintended consequences. Please check with @BertanDogancay if there is an alternative method to reduce kern arg size.

It happens in the testBudget routine in enqueue.cc
https://github.com/ROCm/rccl/blob/develop/src/enqueue.cc#L436

```cpp
 ssize_t batchBytes = nWorkBatches*sizeof(struct ncclDevWorkBatch);
````


nWorkBatches is 256 when using 256 channels, which leads to batchBytes being 4096. 
Because budget->inArgsBytes is 4032 the budget check failed, causing `cheduleCollTasksToPlan` to return ncclSuccess without ever reaching the point where planner->nTasksColl is decremented. 
This caused an infinite loop in ` ncclLaunchPrepare` because a do-while loop condition was never reaching zero `while (planner->nTasksColl + planner->nTasksP2p != 0);`

I will check if I can fix without increasing the budget threshold.

**Comment by isaki001** [comment]
@wenkaidu There doesn't seem to be a way to decrease the kernel args further and the batchBytes will be more than 4K with 256 channels. 
Do we want to strictly stay at 1 page? @BertanDogancay suggested two pages if it doesn't have any negative impact.

**Comment by wenkaidu** [comment]
Do we really need 256 channels? @gilbertlee-amd should we only need multiple of 56, i.e. 56*4=224?

File: src/include/comm.h [hunk 1]
```diff
@@ -222,7 +222,7 @@ struct ncclTaskColl {
   int chunkSteps, sliceSteps;
   // Computed later:
   size_t trafficBytes;
-  int32_t nMaxChannels:8;
+  int32_t nMaxChannels:16;
```
wenkaidu: This can be kept as 8 bits

src/include/device.h
```diff
@@ -532,7 +532,7 @@ struct alignas(16) ncclDevKernelArgs {
 
 __host__ __device__ constexpr int ncclMaxKernelArgsSize(/*int cudaDriver, */int cudaArch=NCCL_CUDA_ARCH) {
   //return (cudaArch < 700 || cudaDriver < 12010) ? 4<<10 : (32<<10)-4;
-  return 4<<10;
+  return 5<<10;
```
wenkaidu: This can be kept as 4K

src/include/device.h
```diff
@@ -544,6 +544,7 @@ struct alignas(16) ncclDevKernelArgsStorage {
 };
 
 typedef ncclDevKernelArgsStorage<(4<<10)> ncclDevKernelArgs4K;
+typedef ncclDevKernelArgsStorage<(5<<10)> ncclDevKernelArgs5K;
```
wenkaidu: This line can be removed

wenkaidu (approved):

thananon:
Do we still need this? If not, can we close?

PR #1550, opened 2025-02-14
Enable GDRCopy only on gfx94x

wenkaidu:
## Details

Internal

**What were the changes?**  
Enable GDRCopy only on gfx94x

It appears to be unsafe to enable this feature on older GPU architectures, i.e. gfx90a

Check GPU architecture and decide

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/include/gdrwrap.h
+++ b/src/include/gdrwrap.h
@@ -164,7 +164,11 @@ static gdr_t ncclGdrInit() {
   cudaDeviceProp devProp;
   char gcnArchNameSubstr[128];
   cudaError_t err = cudaGetDeviceProperties(&devProp, 0);
-  if (err != cudaSuccess) return NULL;
+  if (err != cudaSuccess) {
+    WARN("Failed to GetDeviceProperties for device");
+    return NULL;
+  }
+ return NULL;
   GcnArchNameFormat(devProp.gcnArchName, gcnArchNameSubstr);
   if (IsArchMatch(gcnArchNameSubstr, "gfx94")) {
     INFO(NCCL_INIT, "Enabled GDRCopy equivalent memory allocation on %s", gcnArchNameSubstr);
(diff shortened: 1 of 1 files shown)


src/include/gdrwrap.h
```diff
@@ -160,8 +161,18 @@ typedef struct gdr_mem_desc {
 
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
 static gdr_t ncclGdrInit() {
-  INFO(NCCL_INIT, "Enabled GDRCopy equivalent memory allocation");
-  return (gdr_t)0x12345678L;
+  cudaDeviceProp devProp;
```
mustafabar, line 164: Since this is guaranteed to be in the HIP block, why not just use HIP? In the same file, there are direct hip usages.

wenkaidu: In general, if the CUDA API can be hipified, then using CUDA names has the benefit of direct compilation on CUDA.

wenkaidu: hipFree was missed when we switch to hipify path a few years ago. I will change it as well

thananon (approved):

src/include/gdrwrap.h
```diff
@@ -160,8 +161,18 @@ typedef struct gdr_mem_desc {
...
+  char gcnArchNameSubstr[128];
+  cudaError_t err = cudaGetDeviceProperties(&devProp, 0);
+  if (err != cudaSuccess) return NULL;
```
mustafabar:
```suggestion
  if (err != cudaSuccess) { 
    WARN("Failed to GetDeviceProperties for device");
    return NULL;
  }
```

wenkaidu: done

mustafabar (approved):

PR #1573, opened 2025-02-25
quick fix for MSCCLPP enabled ubr

AtlantaPepsi:
## Details
User Buffer Registration is not guarded by MSCCLPP enable in runtime
now with RCCL_MSCCLPP_ENABLE=0 rccl should work with no error for UBR enabled

## Approval Checklist

### DIFF
--- a/src/register.cc
+++ b/src/register.cc
@@ -161,6 +161,7 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
   NCCLCHECK(CommCheck(comm, "ncclCommRegister", "comm"));
   if (comm->checkPointers) NCCLCHECK(CudaPtrCheck(buff, comm, "buff", "ncclCommRegister"));
   #ifdef ENABLE_MSCCLPP
+  if (comm->mscclppCompatible) {
     if (comm->mscclCompatible && size > 0){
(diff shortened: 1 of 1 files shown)


src/register.cc
```diff
@@ -161,6 +162,7 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
   NCCLCHECK(CommCheck(comm, "ncclCommRegister", "comm"));
   if (comm->checkPointers) NCCLCHECK(CudaPtrCheck(buff, comm, "buff", "ncclCommRegister"));
   #ifdef ENABLE_MSCCLPP
+  if (rcclParamMscclppEnabled()) {
```
corey-derochie-amd: Is `comm->mscclppCompatible` a valid indicator instead? I'm not sure, because it is set in `ncclCommInitRankFunc`, but it might be. Note that it also considers the gcnArch and the result of `mscclppCommCompatible()`.

corey-derochie-amd: I think the `comm->mscclCompatible` check below should stay, but this could be changed to `comm->mscclppCompatible`.
```suggestion
  if (comm->mscclppCompatible) {
```

nusislam (approved):

PR #1574, opened 2025-02-26
[BUILD] Enable multiple GPU targets in MSCCLPP

nileshnegi:
## Details

**Work item:** Internal

**What were the changes?**  
Add support for multiple GPU targets when building MSCCL++

Currently, MSCCL++ only builds for `gfx942`. This needs to change to add support for `gfx950`

Use `%` as the list separator instead of the default `;` in CMake's download_project

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -316,23 +316,29 @@ endif()
 ## Disable building MSCCL++ if the build environment is invalid
 ## Currently MSCCL++ is supported only on gfx942 and gfx950, and only on Ubuntu and CentOS
 set(MSCCLPP_SUPPORTED_ARCHS "gfx942" "gfx942:xnack-" "gfx942:xnack+" "gfx950" "gfx950:xnack-" "gfx950:xnack+")
+
 # Check if any of the supported architectures are in GPU_TARGETS
 set(ARCH_MATCH_FOUND OFF)
-foreach(ARCH ${MSCCLPP_SUPPORTED_ARCHS})
-  if(ARCH IN_LIST GPU_TARGETS)
+set(MSCCLPP_GPU_TARGETS "")
+foreach(ARCH IN LISTS GPU_TARGETS)
+  if(ARCH IN_LIST MSCCLPP_SUPPORTED_ARCHS)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -316,23 +316,29 @@ endif()
 ## Disable building MSCCL++ if the build environment is invalid
 ## Currently MSCCL++ is supported only on gfx942 and gfx950, and only on Ubuntu and CentOS
 set(MSCCLPP_SUPPORTED_ARCHS "gfx942" "gfx942:xnack-" "gfx942:xnack+" "gfx950" "gfx950:xnack-" "gfx950:xnack+")
```
corey-derochie-amd, line 318: Suggestion: move this into MSCCLPP.cmake.

cmake/MSCCLPP.cmake
```diff
@@ -32,18 +32,6 @@
...
-
-    # Only add a cmake argument if it has a value
-    set(${NAME}_ARG "-D${NAME}=\"${ARG_VALUE}\"" PARENT_SCOPE)
-    if("${ARG_VALUE}" STREQUAL "")
-        set(${NAME}_ARG "" PARENT_SCOPE)
-    endif()
-endfunction()
-
-
 if(ENABLE_MSCCLPP)
```
corey-derochie-amd, line 35: Might want to wrap this in a function so we're not include-injecting it directly into CMakeLists.txt. Then we can define `MSCCLPP_SUPPORTED_ARCHS` directly in here and include at top of CMakeLists.txt.

corey-derochie-amd (approved):
Other notes are just suggestions, LGTM.

corey-derochie-amd (approved):

CMakeLists.txt
nileshnegi: Moving these checks straightforward is not the right approach as we add MSCCLPP files/headers to the list of source files  and targets before invoking `cmake/MSCCLPP.cmake`. If we then decide to disable MSCCLPP build (based on these checks), the build still tries to use find mscclpp symbols.
Need a better approach. To be fixed in a different PR.

PR #1576, opened 2025-02-26
misc/msccl: Read graph capture status for every collective call

nusislam:
## Details
If the same stream is used in non-graph mode followed by graph mode, the capture status was not getting updated. This PR adds a fix for this issue by reading the capture status in every collective call.

**What were the changes?**  

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/ext-src/reg-fix.patch
+++ b/ext-src/reg-fix.patch
@@ -1,0 +1,43 @@
+diff --git a/apps/nccl/src/nccl.cu b/apps/nccl/src/nccl.cu
+index 5c19dc6..5fb99ef 100644
+--- a/apps/nccl/src/nccl.cu
++++ b/apps/nccl/src/nccl.cu
+@@ -85,6 +85,7 @@ struct ncclComm {
+   std::unordered_map<channelKey, ChannelInfo> channelInInfos;
+   std::unordered_map<channelKey, ChannelInfo> channelOutInfos;
+   std::unordered_map<channelKey, ChannelInfo> channelScratchInfos;
++  std::unordered_map<channelKey, cudaIpcMemHandle_t> regHandles;
+   std::unordered_map<void*, channelKey> handleKeys;
+   std::shared_ptr<char> scratchBuff;
+   std::vector<mscclpp::RegisteredMemory> remoteScratchRegMemories;
+@@ -616,6 +617,11 @@ NCCL_API ncclResult_t ncclCommRegister(ncclComm_t comm, void* buff, size_t size,
+   p->ipcHandle = ipcHandle;
+   *handle = p;
+ 
++  auto regIt = comm->regHandles.find(buffKey);
++  if (regIt == comm->regHandles.end()) {
++        comm->regHandles[buffKey] = ipcHandle;
++  }
++
+   auto it = comm->handleKeys.find(*handle);
+   if (it == comm->handleKeys.end()) {
+      comm->handleKeys[*handle] = buffKey;
+@@ -642,6 +648,7 @@ NCCL_API ncclResult_t ncclCommDeregister(ncclComm_t comm, void* handle) {
+      if (outIt != comm->channelOutInfos.end()) {
+         comm->channelOutInfos.erase(outIt);
+      }
++     comm->regHandles.erase(buffKey);
+      comm->handleKeys.erase(handle);
+      free(handle);
+   }
+@@ -655,8 +662,8 @@ bool mscclpp_BuffIsRegistered(ncclComm_t comm, const void* buff){
+   CUdeviceptr buffBasePtr;
+   MSCCLPP_CUTHROW(cuMemGetAddressRange(&buffBasePtr, &buffBytes, (CUdeviceptr)buff));
+   channelKey buffKey{(void*)buffBasePtr, buffBytes};
+-  auto buffIt = comm->channelScratchInfos.find(buffKey);
+-  bool registered =  buffIt != comm->channelScratchInfos.end();
++  auto buffIt = comm->regHandles.find(buffKey);
++  bool registered =  buffIt != comm->regHandles.end();
+   return registered;
+ }
+ size_t
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -519,18 +519,16 @@ ncclResult_t mscclEnqueueCheck(
     case mscclNoGroup:
 #ifdef ENABLE_MSCCLPP
       if (comm->mscclppCompatible) {
-        //if (threadLocalStatus.captureStatus == mscclUnknownCaptureStatus) {
         INFO(NCCL_COLL, "MSCCL++: reading capture status");
         NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
-        //}
 
         const bool sendBuffRegistered = mscclpp_BuffIsRegistered(comm->mscclpp_comm, sendBuff); 
         const bool recvBuffRegistered = mscclpp_BuffIsRegistered(comm->mscclpp_comm, recvBuff);
         const bool graphMode = threadLocalStatus.captureStatus != mscclNoCapture;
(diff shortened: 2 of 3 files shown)


src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -519,15 +519,15 @@ ncclResult_t mscclEnqueueCheck(
...
 #ifdef ENABLE_MSCCLPP
       if (comm->mscclppCompatible) {
-        if (threadLocalStatus.captureStatus == mscclUnknownCaptureStatus) {
-          INFO(NCCL_COLL, "MSCCL++: reading capture status");
-          NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
-        }
+        //if (threadLocalStatus.captureStatus == mscclUnknownCaptureStatus) {
+        INFO(NCCL_COLL, "MSCCL++: reading capture status");
+        NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
+        //}
```
corey-derochie-amd: I'd also suggest that, if we read the capture status for every collective, no longer a need to log it here (or could convert to `TRACE`).
```suggestion
        INFO(NCCL_COLL, "MSCCL++: reading capture status");
        NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
```

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -565,15 +565,15 @@ ncclResult_t mscclEnqueueCheck(
...
```
corey-derochie-amd: Same as above.
```suggestion
        INFO(NCCL_COLL, "MSCCL++: reading capture status");
        NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
```

src/register.cc
```diff
@@ -167,6 +167,7 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
       if(!isManagedBuffer){
         INFO(NCCL_INIT, "MSCCL++: ncclCommRegister");
         NCCLCHECK(mscclpp_ncclCommRegister(comm->mscclpp_comm, buff, size, handle));
+	comm->ubr = true;
```
corey-derochie-amd: This seems a bit off... UBR is a property of the buffer, but this saves a flag on the communicator. That communicator could be used with other buffers, no?

src/register.cc
```diff
@@ -186,7 +187,8 @@ ncclResult_t ncclCommDeregister_impl(const ncclComm_t comm, void* handle) {
     const size_t size = mscclpp_BufferSize(comm->mscclpp_comm, handle);
     if (comm->mscclCompatible && size > 0) {
         NCCLCHECK(mscclpp_ncclCommDeregister(comm->mscclpp_comm, handle));
-      return ncclSuccess;
+	comm->ubr = false;
```
corey-derochie-amd: Yes, this would cause a bug with multiple registered buffers, because the flag will be set false while other registered buffers still exist.

src/register.cc
corey-derochie-amd: Is this actually just working around a bug in `mscclpp_BuffIsRegistered`?

nusislam: No, it is not working around a bug in `mscclpp_BuffIsRegistered`. We need to know that the buffer has been cached in mscclpp due to the user calling `ncclCommRegister()` not during the first collective call. Typically, registration will be called once/twice per communicator (with the buffers of the collective).

nusislam: Let me see if I can do it in a different way.

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -519,33 +519,31 @@ ncclResult_t mscclEnqueueCheck(
...
+              "mscclpp_ncclAllReduce", comm->opCount, sendBuff, recvBuff, count, dataType, op, root, comm, comm->nRanks, stream, graphMode);
             NCCLCHECK(mscclpp_ncclAllReduce(sendBuff, recvBuff, count, dataType, op, comm->mscclpp_comm, stream));
             threadLocalStatus.savedSchedulerParams.clear();
             break;
           }
           else if (func == mscclFuncAllGather && nBytes * comm->nRanks <= comm->mscclpp_threshold) {
-            INFO(NCCL_COLL,"%s: opCount %lx sendbuff %p recvbuff %p count %zi datatype %d op %d root %d comm %p [nranks=%d] stream %p",
-              "mscclpp_ncclAllGather", comm->opCount, sendBuff, recvBuff, count, dataType, op, root, comm, comm->nRanks, stream);
+            INFO(NCCL_COLL,"%s: opCount %lx sendbuff %p recvbuff %p count %zi datatype %d op %d root %d comm %p [nranks=%d] stream %p graphMode %d",
+              "mscclpp_ncclAllGather", comm->opCount, sendBuff, recvBuff, count, dataType, op, root, comm, comm->nRanks, stream, graphMode);
```
corey-derochie-amd, line 546: Should these be included in the final logging?

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -519,33 +519,31 @@ ncclResult_t mscclEnqueueCheck(
     case mscclNoGroup:
 #ifdef ENABLE_MSCCLPP
       if (comm->mscclppCompatible) {
-        if (threadLocalStatus.captureStatus == mscclUnknownCaptureStatus) {
-          INFO(NCCL_COLL, "MSCCL++: reading capture status");
-          NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
-        }
+        INFO(NCCL_COLL, "MSCCL++: reading capture status");
```
corey-derochie-amd, line 522: No longer need to log this?
```suggestion
```

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -565,33 +563,31 @@ ncclResult_t mscclEnqueueCheck(
     case mscclGroupSupportedOp:
 #ifdef ENABLE_MSCCLPP
       if (comm->mscclppCompatible) {
-        if (threadLocalStatus.captureStatus == mscclUnknownCaptureStatus) {
-          INFO(NCCL_COLL, "MSCCL++: reading capture status");
-          NCCLCHECK(mscclGetCaptureStatus(comm->rank, stream));
-        }
+	INFO(NCCL_COLL, "MSCCL++: reading capture status");
```
corey-derochie-amd, line 566: No longer need to log this?
```suggestion
```

ext-src/reg-fix.patch
```diff
@@ -0,0 +1,43 @@
...
+   std::shared_ptr<char> scratchBuff;
+   std::vector<mscclpp::RegisteredMemory> remoteScratchRegMemories;
+@@ -616,6 +617,11 @@ NCCL_API ncclResult_t ncclCommRegister(ncclComm_t comm, void* buff, size_t size,
+   p->ipcHandle = ipcHandle;
+   *handle = p;
+ 
++  auto regIt = comm->regHandles.find(buffKey);
++  if (regIt == comm->regHandles.end()) {
++        comm->regHandles[buffKey] = ipcHandle;
++  }
```
corey-derochie-amd, line 20: Is it possible to register a different `ipcHandle` for the same key? If so this may have unexpected behaviour.

corey-derochie-amd (approved):
LGTM, only a few questions remain.

ext-src/reg-fix.patch
nusislam: That's not possible with UBR.

src/misc/msccl/msccl_lifecycle.cc
nusislam: It is helpful for debugging. So I added it here.

src/misc/msccl/msccl_lifecycle.cc
nusislam: This log is enabled for `COLL` subsystem only. I think it can still remain. Do you see any issues with that?

PR #1581, opened 2025-03-03
misc/msccl: force use of mscclpp

nusislam:
## Details
This PR adds an environment variable `RCCL_MSCCLPP_FORCE_ENABLE=1` to force the use of mscclpp for non-graph and non-ubr mode. Based on the workload characteristics, this feature may or may not be useful. So, the default value of this environment variable is `0`.

**What were the changes?**  

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -121,6 +121,7 @@ static constexpr int64_t defaultEnableMscclpp = 0;
 #endif
 
 RCCL_PARAM(MscclppEnabled, "MSCCLPP_ENABLE", defaultEnableMscclpp);
+RCCL_PARAM(MscclppForceEnabled, "MSCCLPP_FORCE_ENABLE", 0);
 
 // GDRCOPY support: Off by default
 NCCL_PARAM(GdrCopyEnable, "GDRCOPY_ENABLE", 1);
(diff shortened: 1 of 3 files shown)


isaki001 (commented):
Can we not use the already defined mscclForceEnabled() for these checks?

corey-derochie-amd:
> Can we not use the already defined mscclForceEnabled() for these checks?

No, `mscclForceEnable()` will only force MSCCL and will not bypass the MSCCL++ graph mode or UBR check.

isaki001:
> > Can we not use the already defined mscclForceEnabled() for these checks?
> 
> No, `mscclForceEnable()` will only force MSCCL and will not bypass the MSCCL++ graph mode or UBR check.

Oh I see. Makes sense.

nusislam:
@corey-derochie-amd , @isaki001 - Can you review the PR? The test failure is not related to this PR.

src/init.cc
```diff
@@ -1945,6 +1946,11 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
...
         mscclpp_commToUniqueIdMap[comm->mscclpp_comm] = mscclppUniqueId;
         ncclCommToUniqueIdMap[comm] = job->commId;
+	if (rcclParamMscclppForceEnabled()) {
+		comm->mscclppForceEnable = true;
+	} else {
+		comm->mscclppForceEnable = false;	
+	}
       } else {
         WARN("MSCCL++: Cannot enable MSCCL++ on %s architecture", devProp.gcnArchName);
       }
```
corey-derochie-amd, line 1956: I'd suggest looking around and adding warnings like this anywhere the user force enables MSCCL++ in an environment where it cannot be used (if there isn't already a warning). This will let them know their intention is being denied.

corey-derochie-amd (approved):

src/init.cc
nusislam: This force mscclpp use is not for architectures where mscclpp is not supported, rather using it on architectures that support mscclpp but cannot use because of graph Mode/UBR not being enabled.

PR #1593, opened 2025-03-07
Add fault injection of starting warps with random variations

wenkaidu:
This is done by inserting randomly delays after __syncthreads(). The feature can be turned off by FAULT_INJECTION=OFF in cmake.

## Details
Add fault injection functionality to RCCL

**What were the changes?**  
Add fault injection of starting warps with random variations

Allow more coverage of issues that are hard to reproduce in unit tests environment

Add fault injection of starting warps with random variations

**Additional Documentation:**  

## Approval Checklist


src/device/common.h
```diff
@@ -160,6 +165,28 @@ extern __shared__ ncclShmemData ncclShmem;
   extern __shared__ ulong2 ncclShmemPerWarp[ncclShmemScratchWarpSize()*(NCCL_MAX_NTHREADS/WARP_SIZE)/sizeof(ulong2)];
 #endif
 
+#ifdef ENABLE_FAULT_INJECTION
+__device__ inline void insert_random_delay_per_warp() {
+  if (ncclShmem.faults & RANDOM_DELAY_ON_WARP_START) {
+    switch ((wall_clock64()>>(threadIdx.x/WARP_SIZE*2))&0x3) {
+      case 0:
+        __builtin_amdgcn_s_sleep(0);
```
cenzhaometa, line 173: these 4 sleeps are still deterministic, shall we use pure random numbers or you think this is good enough?

wenkaidu: It kind of random by reading the wall clock which changes rapidly. each warp takes different 2 bits from clock value which then determine the sleep cycles

cmake/scripts/add_faults.sh
```diff
@@ -0,0 +1,27 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+HIP_FILE=$1
+
+if [[ "$HIP_FILE" =~ .*/src/device/.*\.h ]]; then
+  sed -i "s/__syncthreads()/__syncthreads(); insert_random_delay_per_warp()/" "$HIP_FILE"
```
cenzhaometa, line 24: for Meta internal builds (using buck), we don't use cmake. so for any changes on cmake, we have to duplicate the effort in buck, can we avoid ad-hoc scripts and move them into code directly? 

e.g if #define FALUT_INEJCT insert_random_delay()..

wenkaidu: The purpose is to be able to automatically inject variations at all possible opportunities automatically with future code changes. It is just "sed" command, not really tied to cmake. Can buck execute sed? If not, we can try to use python script for this

cenzhaometa (commented):
can we add UT to demonstrate such it can reproduce the failure case w/o our previous fix, and pass now?

wenkaidu:
> can we add UT to demonstrate such it can reproduce the failure case w/o our previous fix, and pass now?

yes, I can give you a demo

cmake/scripts/add_faults.sh
wenkaidu: To reproduce the failure, we had to move __syncthreads() to the incorrect location.
https://github.com/ROCm/rccl/pull/1593/files#diff-ae625b0327a5d90bbe9414bc8f977b8dc59ba88c7a35c545d19aaa43e32577a5R576
This cannot be merged. So UT will not be able to reproduce failure on current develop without manually introduce bug in code.

dmwu: I was actually thinking that all the tests we add be placed in a separate test dir and files. Each test case can be built separately into a binary, e.g., using gTest. Following this, we should make very minimal change to the rccl src. Rather, all these changes to build file and code, e.g.,  insert_random_delay_per_warp, should be put into that test file, not the rccl src code.

src/device/common.h
```diff
@@ -44,22 +44,24 @@
     // TODO: switch to atomicInc after llvm crash is fixed
     // uint32_t pos = atomicInc(&ncclShmem.collTraceTail->tail, COLLTRACE_NUM_ITEMS)
 
-  #define traceKernelLaunch(launch_type) { \
+  #define traceKernelLaunch(launch_type, ix) { \
     INC_COLL_TRACE \
     collTrace->funcIndex = ncclShmem.funcId; \
     __trace_hwreg()\
+    collTrace->batchIx = ix; \
```
dmwu, line 51: What does this collTrace->batchIx and following p2pWork->sendRegistered, p2pWork->recvRegistered do?

cmake/scripts/add_faults.sh
wenkaidu: The goal is to introduce faults, in this case, race conditions to RCCL and it will impact all test cases, including future test cases. For example, when we inject random delays for warps and run full UT, we will be able to confirm all existing test cases are passing. We can also inject when running real workloads so failure can occur faster

src/device/common.h
wenkaidu: NCCL 2.22 added two __syncthreads() in sendrecv.h which wasn't there in 2.21. The result is the race condition no longer appears as hang in sub-barrier. Code got stuck in one of the  __syncthreads(). Adding batchIx can let us see the failure mode, i.e. mismatched batchIx. Since I was fixing traces, I just added registered bits trace as these will be useful in debugging user registered buffer.

src/device/common.h
gilbertlee-amd: There's actually variation already with this sleep function.  S_SLEEP 8 really means that it could wait between 449 to 512 clock cycles.

gilbertlee-amd (approved):

src/device/common.h
dmwu: @gilbertlee-amd Why the s_sleep function is indeterministic in the clock cycles? Is it by design to do so or a implementation bug?  this sounds a bit scary

gilbertlee-amd: I based that comment on what I found here:
www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/amd-instinct-mi300-cdna3-instruction-set-architecture.pdf

"S_SLEEP Cause a wave to sleep for up to ~8000 clocks. 14 The wave sleeps for (64*(SIMM16[6:0]-1) .. 64*SIMM16[6:0]) clocks. The exact amount of delay is approximate. Compare with S_NOP. When SIMM16[6:0] is zero then no sleep occurs."

I think this is simply a hardware detail.

PR #1603, opened 2025-03-14
RCCL Replayer update

AtlantaPepsi:
## Details
Revised RCCL Replayer, also adding structured logging.

**Work item:** Internal

**What were the changes?**  

-  adding replayer as a part of RCCL workload which **capture** RCCL API called, param used, and GPU context info (e.g. graph, device, timestamp, etc.)
-  support binary and json format output
-  some rudimentary modification to parsing in replayer, requires further change to **replay** logic
-  supports recording of collective calls, group and communicator operation except comm split

Current RCCL Replayer relies on NCCL DEBUG output, which may not be so easily parsable depending on workloads (given its many level and subsys). It also only supports collectives and don't accurately replay other RCCL operation and context. In general we needed more detailed logging option of RCCL beyond different levels of NCCL DEBUG log, as well as more powerful and efficient RCCL Replayer. 

Instead of relying on NCCL log we now have our own tailored replayer structures and objects as part of RCCL and captures required info as needed. Replaying part already existed inside `tools/rccl_replayer`, further work is required for parsing and replaying the new format of log


amd-jnovotny:
@AtlantaPepsi Are any docs updates required? I guess we will need a changelog entry too given that the changes will be apparent to customers.

src/collectives.cc
```diff
@@ -85,6 +85,14 @@ NCCL_API(ncclResult_t, ncclAllGather, const void* sendbuff, void* recvbuff, size
 
 ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sendcount,
     ncclDataType_t datatype, ncclComm_t comm, cudaStream_t stream) {
+
+  struct CollectiveCall coll = {-1, -1, -1, -1, 0, 0, 0, "AllGather" ,rrAllGather, comm->opCount, sendbuff, recvbuff, sendcount,
+    datatype, ncclSum, 0, comm, comm->nRanks, stream, comm->planner.nTasksP2p + comm->planner.nTasksColl,
+    comm->localRankToRank[comm->localRank]};
+
+  if (!mscclIsCaller()) // when msccl falls back to
+    NCCLCHECK(Replayer::rcclReplayerRecord(coll));
```
gilbertlee-amd: Would this logic for checking if MSCCL is falling back to RCCL be better off within Replayer::rcclReplayerRecord, instead of everywhere?

src/include/replayer.h
```diff
@@ -0,0 +1,85 @@
...
+  Replayer();
+  ~Replayer();
+
+ public:
+  static Replayer&      instance();
+  static ncclResult_t   rcclReplayerRecord(CollectiveCall& coll);
+  //static ncclResult_t   rcclReplayerWrite();
+};
+
+static __thread int     tid = -1;
```
gilbertlee-amd: suggest changing this variable name to rcclReplayThreadIdx

src/collectives.cc
```diff
@@ -85,6 +85,14 @@ NCCL_API(ncclResult_t, ncclAllGather, const void* sendbuff, void* recvbuff, size
 
 ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sendcount,
     ncclDataType_t datatype, ncclComm_t comm, cudaStream_t stream) {
+
+  struct CollectiveCall coll = {-1, -1, -1, -1, 0, 0, 0, "AllGather" ,rrAllGather, comm->opCount, sendbuff, recvbuff, sendcount,
+    datatype, ncclSum, 0, comm, comm->nRanks, stream, comm->planner.nTasksP2p + comm->planner.nTasksColl,
+    comm->localRankToRank[comm->localRank]};
```
gilbertlee-amd: This call looks prone to mistakes.  
I'll recommend having various helper constructor calls to CollectiveCall that match pretty much what the input parameters are.  The rest can be filled in by the helper function.

E.g.:  CollectiveCall(sendbuff, recvbuff, sendcount, datatype, comm, stream);

src/include/replayer.h
```diff
@@ -0,0 +1,85 @@
...
+  rrCommSplit = 24,
+  rrCommFinalize = 25,
+  rrCommDestroy = 26,
+  rrCommAbort = 27,
+  rrCommRegister = 30,
+  rrCommDeregister = 31
+} rcclCall_t;
+
+using namespace std::chrono;
+struct CollectiveCall {
```
gilbertlee-amd: We should rename this to be more clear that this is a rccl addition.
It might not even always be a collective call.

rcclReplayCall?

src/include/replayer.h
```diff
@@ -0,0 +1,85 @@
...
+  int                   tid;
+  int                   hipDev;
+  int                   groupDepth;
+  double                timestamp;
+  unsigned long long    graphID;
+  int                   graphCaptured;
+
+//explicit data from header
+  const char*           name; // for printing log, not binary parsable
+  rcclCall_t            coll; // in adjacent to op Name ^
```
gilbertlee-amd: coll might not be the best name.

src/init.cc
```diff
@@ -2200,6 +2200,17 @@ static ncclResult_t ncclCommInitRankDev(ncclComm_t* newcomm, int nranks, ncclUni
...
 exit:
+  // for loggin only, not ready for replaying
+  // !recording at sink
+  CollectiveCall initCall;
+  initCall.coll = rrCommInitRank;
+  initCall.comm = comm;
+  initCall.nRanks = nranks;
+  initCall.globalRank = myrank;
+  initCall.hipDev = cudaDev; // ...?
+  initCall.name = "CommInitRank";
```
gilbertlee-amd: Move this the helper CollectiveCall constructor function for now in Replayer.  Try not to overly modify original source code file / make sure that it's evident that this is a RCCL addition.

gilbertlee-amd (changes requested):

src/collectives.cc
corey-derochie-amd: Alternatively, you can use struct named-field initialization style to make it less error-prone. But it will also be verbose.

src/group.cc
```diff
@@ -91,6 +91,11 @@ ncclResult_t ncclAsyncJobComplete(struct ncclAsyncJob* job) {
 
 NCCL_API(ncclResult_t, ncclGroupStart);
 ncclResult_t ncclGroupStart_impl() {
+  struct CollectiveCall gc;
+  gc.coll = rrGroupStart;
+  gc.groupDepth = ncclGroupDepth;
+  gc.name = "GroupStart";
```
corey-derochie-amd: This case should allow for struct field initialization.

src/collectives.cc
corey-derochie-amd: `mscclIsCaller()` only makes sense in this context, and not everywhere we call `rcclReplayerRecord()` requires it. I'd suggest leaving as-is.

src/include/replayer.h
```diff
@@ -0,0 +1,85 @@
...
+  rrCommInitRank = 20,
+  rrCommSplit = 24,
+  rrCommFinalize = 25,
+  rrCommDestroy = 26,
+  rrCommAbort = 27,
+  rrCommRegister = 30,
+  rrCommDeregister = 31
+} rcclCall_t;
+
+using namespace std::chrono;
```
corey-derochie-amd: Recommend avoiding `using namespace` within header files, or else it will pollute the namespace in each source it is included from.

src/include/replayer.h
```diff
@@ -0,0 +1,85 @@
...
+  int                   root;
+  ncclComm_t            comm;
+  int                   nRanks;
+  cudaStream_t          stream;
+  int                   nTasks;
+  int                   globalRank;
+};
+
+class Replayer {
+ private:
```
corey-derochie-amd: FYI, default accessibility for a `class` is already `private`, so this isn't necessary.

src/include/replayer.h
corey-derochie-amd: This is a `static` global field declared in a header, which is generally not what you want. It should either be declared in the source file where it is used, or if it is used throughout the code, it should be marked `extern` and declared in one source file as non-`static`.

src/init.cc
```diff
@@ -2415,6 +2426,12 @@ static ncclResult_t commCleanup(ncclComm_t comm) {
 
 NCCL_API(ncclResult_t, ncclCommFinalize, ncclComm_t comm);
 ncclResult_t ncclCommFinalize_impl(ncclComm_t comm) {
+  CollectiveCall finalizeCall;
+  finalizeCall.coll = rrCommFinalize;
+  finalizeCall.comm = comm;
+  finalizeCall.name = "CommFinalize";
+  NCCLCHECK(Replayer::rcclReplayerRecord(finalizeCall));
```
corey-derochie-amd: This is just an example of what's valid. Others can decide if this is readable or desirable enough.
```suggestion
  NCCLCHECK(Replayer::rcclReplayerRecord({ .coll=rrCommFinalize, .comm=comm, .name="CommFinalize"}));
```

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+  {
+    output_name = std::string(filename).substr(0, dot);
+    output_extension = std::string(filename).substr(dot);
+    if (output_extension.compare(".json") == 0)
+    {
+      output_json = 1;
+    }
+  } else {
+    output_name = std::string(filename);
+  }
```
corey-derochie-amd: If we moved to C++17, we could use `std::filesystem` instead.

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+
+Replayer& Replayer::instance()
+{
+  static Replayer INSTANCE;
+  return INSTANCE;
+}
+
+void Replayer::rcclReplayerWrite(CollectiveCall &call)
+{
+  Replayer& replayer = Replayer::instance();
```
corey-derochie-amd: Since you're referencing the singleton anyway, this function can just be a non-`static` function.

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+  if (output_json) outputFile << "{\n";
+}
+
+Replayer& Replayer::instance()
+{
+  static Replayer INSTANCE;
+  return INSTANCE;
+}
+
+void Replayer::rcclReplayerWrite(CollectiveCall &call)
```
corey-derochie-amd: `call` is not modified, so it can be `const CollectiveCall&`.

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+  Replayer& replayer = Replayer::instance();
+  size_t collidx = (size_t) idx;
+  CollectiveCall call = replayer.colls[collidx];
+  Replayer::rcclReplayerRecord(call);
+}
+
+ncclResult_t Replayer::rcclReplayerRecord(CollectiveCall& call)
+{
+  call.timestamp = duration_cast<duration<double>>(high_resolution_clock::now().time_since_epoch()).count() * 1000;
+  Replayer& replayer = Replayer::instance();
```
corey-derochie-amd: This can be a non-`static` function too.

tools/RcclReplayer/rcclReplayer.cpp
```diff
@@ -180,31 +180,39 @@ void dataToCsv(GroupCall const& gc, std::ofstream &datafile, double runTime)
...
+  std::ifstream fp(logFilename, std::ios::binary);
   if (!fp) {
     printf("[ERROR] Unable to open file %s\n", logFilename);
     exit(-1);
   }
 
-  char line[2048];
-  LineItem li;
+  const size_t size = sizeof(LineItem) + 1;
+  char line[size]; //size of collectivecall struct
```
corey-derochie-amd: This may generate a variable-length-array warning (not sure). To avoid, I think it should be:
```suggestion
  constexpr size_t size = sizeof(LineItem) + 1;
  char line[size]; //size of collectivecall struct
```

tools/RcclReplayer/rcclReplayer.hpp
```diff
@@ -34,25 +35,30 @@
     }                                                           \
   } while(0)
 
+using namespace std::chrono;
```
corey-derochie-amd: Avoid `using namespace` in headers.

tools/RcclReplayer/rcclReplayer.cpp
```diff
@@ -180,31 +180,39 @@ void dataToCsv(GroupCall const& gc, std::ofstream &datafile, double runTime)
...
   int lineNum = 0;
 
-  while (fgets(line, 2048, fp)) {
-    ++lineNum;
+  while (fp.getline(line, size))
+  {
+    LineItem li = *((LineItem*)line);
```
corey-derochie-amd, line 199: I'm not sure, but do we need to worry about portability? When serializing the struct binary directly, you have to worry about endianness and alignment.

Also, I'm not sure how this plays with `ifstream::getline()`, which I think is meant to work with text not binary. If the struct contains a field with value equal to `'\n'`, would this work as expected?

corey-derochie-amd (dismissed):

tools/RcclReplayer/Makefile
```diff
@@ -0,0 +1,12 @@
+ROCM_DIR ?= /opt/rocm
+RCCL_DIR ?= ../../build/release
```
nileshnegi, line 2: can you add a check here for cases where someone does not build RCCL locally and plans to use RCCL Replayer with `$ROCM_DIR/lib/librccl.so`?

src/register.cc
```diff
@@ -166,21 +173,32 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
...
   INFO(NCCL_INIT, "RCCL: ncclCommRegister");
-  NCCLCHECK(ncclRegister(comm, buff, size, handle));
+  NCCLCHECKGOTO(ncclRegister(comm, buff, size, handle), ret, end);
+
+end:
+  registerCall.sendbuff = buff;
+  registerCall.recvbuff = *handle;
+  registerCall.name = "CommRegister";
+  NCCLCHECK(Replayer::rcclReplayerRecord(registerCall));
   return ncclSuccess;
```
corey-derochie-amd: I think this should return `ret`, is that correct?

src/register.cc
```diff
@@ -158,6 +158,13 @@ ncclResult_t ncclRegCleanup(struct ncclComm* comm) {
...
 ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t size, void** handle) {
+  // !recording at sink
+  CollectiveCall registerCall;
+  registerCall.coll = rrCommRegister;
+  registerCall.count = size;
+  registerCall.comm = comm;
+  ncclResult_t ret = ncclSuccess;
+
   NCCLCHECK(CommCheck(comm, "ncclCommRegister", "comm"));
   if (comm->checkPointers) NCCLCHECK(CudaPtrCheck(buff, comm, "buff", "ncclCommRegister"));
```
corey-derochie-amd, line 170: Do these need to be `NCCLCHECKGOTO`?

src/register.cc
```diff
@@ -166,21 +173,32 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
       CUDACHECK(hipPointerGetAttribute(&isManagedBuffer, HIP_POINTER_ATTRIBUTE_IS_MANAGED, const_cast<void*>(buff)));
```
corey-derochie-amd, line 175: Is there a `CUDACHECKGOTO`?

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+
+  ncclResult_t ret = ncclSuccess;
+  if (!replayer.filename)
+    return ret;
+
+  if (tid == -1)
+    tid = syscall(SYS_gettid);
```
corey-derochie-amd: Is the `gettid()` function a valid choice?

src/register.cc
```diff
@@ -166,21 +173,32 @@ ncclResult_t ncclCommRegister_impl(const ncclComm_t comm, void* buff, size_t siz
...
   #endif
   INFO(NCCL_INIT, "RCCL: ncclCommRegister");
-  NCCLCHECK(ncclRegister(comm, buff, size, handle));
+  NCCLCHECKGOTO(ncclRegister(comm, buff, size, handle), ret, end);
+
+end:
+  registerCall.sendbuff = buff;
+  registerCall.recvbuff = *handle;
+  registerCall.name = "CommRegister";
+  NCCLCHECK(Replayer::rcclReplayerRecord(registerCall));
```
corey-derochie-amd: If this `NCCLCHECK` fails, it will wipe out the result stored in `ret`. Is that intended?

tools/RcclReplayer/rcclReplayer.cpp
corey-derochie-amd: Checking the documentation confirms my worry. You probably want `ifstream::read()`.
https://en.cppreference.com/w/cpp/io/basic_istream/getline
https://en.cppreference.com/w/cpp/io/basic_istream/read

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+{
+  Replayer& replayer = Replayer::instance();
+
+  if (!replayer.filename)
+    return ;
+
+  std::unique_lock<std::mutex> lock(replayer.mtx);
+
+  static int depth = 0; // for indentation purpose
+  const char *ctxt_fmt = "time : %f, thread : %d, device : %d, captured : %d, graphID : %llu ]]\n"; // implicit context info
```
corey-derochie-amd: Recommend moving this down to where it's used to aid in readability. It's hard to see what's being written to the log at the place it's being written.

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+        const char *destroy_fmt = "%s : [comm : %p, size : %d, rank : %d, context : [";
+        len = snprintf(buffer, 4096, destroy_fmt, call.name, call.comm, call.nRanks, call.globalRank);
+      }
+    } else {
+      replayer.outputFile << "UNKNOWN operation\n";
+      return;
+    }
+    replayer.outputFile.write(buffer, len);
+    len = snprintf(buffer, 4096, ctxt_fmt, call.timestamp, call.tid, call.hipDev, call.graphCaptured, call.graphID);
+    replayer.outputFile.write(buffer, len);
```
corey-derochie-amd: This might work, but when writing as text the more idiomatic way to do this is to use the `<<` operator, which will handle localization, character sets, etc.

src/misc/replayer.cc
```diff
@@ -0,0 +1,193 @@
...
+    return ;
+  } else {
+    replayer.outputFile.write((char*)&call, sizeof(CollectiveCall));
+    replayer.outputFile << std::endl;
```
corey-derochie-amd: When writing as binary, I think the `endl` is unnecessary. You might need this because you were using `ifstream::getline()` to read it, but please see my other comment about that.

src/misc/replayer.cc
corey-derochie-amd: Also, in that case, it's more "proper" to use `endl` instead of `'\n'` directly in the strings, again due to various string-handling reasons, including things like Windows line-endings.

src/register.cc
AtlantaPepsi: User buffer register didnt have the goto clause, but rather just returns status directly with CUDA/NCCLCHECK. I added the goto clause for CommRegister, because replayer record functions depends on the completion of all msccl/rccl UBR function to return the handle, and I added a goto for all branches

src/register.cc
AtlantaPepsi: i think so, if `ret == success`, NCCLCHECK (hip graph) fails, NCCLCHECK will return failure
                if `ret` fails, NCCLCHECK also fails, if we had to choose one failure to return, I'd say NCCLCHECK (hip error) takes priority
                if NCCLCHECK also succeeds, we just return `ret` (nccl/msccl_Register status)

src/include/replayer.h
```diff
@@ -0,0 +1,91 @@
...
+  rrCommInitRank = 20,
+  rrCommSplit = 24,
+  rrCommFinalize = 25,
+  rrCommDestroy = 26,
+  rrCommAbort = 27,
+  rrCommRegister = 30,
+  rrCommDeregister = 31
+} rcclCall_t;
+
+struct rcclApiCall {
```
corey-derochie-amd: This is fine, but this struct doesn't actually represent a call, but rather a record of a call. Maybe consider naming it `rcclReplayRecord` or `rcclCallRecord` or some such.

src/include/replayer.h
```diff
@@ -0,0 +1,91 @@
...
+  ncclRedOp_t           op;
+  int                   root;
+  ncclComm_t            comm;
+  int                   nRanks;
+  cudaStream_t          stream;
+  int                   nTasks;
+  int                   globalRank;
+
+  rcclApiCall(){}
+  rcclApiCall(ncclInfo info, rcclCall_t type);
```
corey-derochie-amd: Recommend passing `info` as `const ncclInfo&`, as the current signature will copy the struct on constructor call.

src/misc/replayer.cc
```diff
@@ -0,0 +1,208 @@
+#include "bootstrap.h"
+#include "group.h"
+#include "utils.h"
+#include <string>
+#include <sys/syscall.h>
+ 
+using namespace std::chrono;
+int rcclReplayThreadIdx = -1;
```
corey-derochie-amd: I just checked, and I think this should be
```suggestion
__thread int Replayer::rcclReplayThreadIdx = -1;
```

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
+#include <fstream>
+#include <vector>
+#include <mutex>
+#include "debug.h"
+//#include "nccl.h"?
```
cenzhaometa: let's drop these commented out code

(same comment for other places)

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+
```
cenzhaometa: suggestion: add namespace for any newly introduced code to avoid naming conflicts (e.g CollectiveCall is too generic)

```
namespace replay {
struct CollectiveCall {
};

...
};
```

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+struct CollectiveCall {
+//  const char*		hostname; //<- i dont know what or how to do this in allgather
+  int                   pid;
```
cenzhaometa: nit: for primitive types (e.g int, bool etc), always add default initializer, initialize with invalid value.

`int pid{-1};`

(same comments for other places)

cenzhaometa: also for these fields, do we intent to change them once constructed, if not, let's mark them as const.

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+
+class Replayer {
+ private:
+  std::ofstream         outputFile;
+
+  char                  hostname[1024];
+  int                   pid = -1;
+  int                   numCall = 0; //for debugging only
+
+  std::mutex            mtx; //single lock or each in call?
```
cenzhaometa: should this mutex be static too?

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+  std::vector<CollectiveCall> colls;
+
+  char*                 filename;
```
cenzhaometa: let's use std::string, char* doesn't own anything, such raw pointer is not safe.

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+
```
cenzhaometa: let's also delete copy/move constructor, I assume this is a singleton class.
`
Replayer(const Replayer&) = delete;
Replayer& operator=(const Replayer&) = delete;
`

src/include/replayer.h
```diff
@@ -0,0 +1,49 @@
...
+  Replayer();
+  ~Replayer();
+
+ public:
+  static Replayer&      instance();
+  static void           rcclReplayerRecord(struct CollectiveCall& coll);
```
cenzhaometa: nit: for c++, we don't need `struct` keyword

do we intend to change the state of coll, if not:
1) let's pass right reference (&&) to enforce caller to do std::move(coll) to release its ownership. 
2) if you want to retain caller's ownership, pass const (CollectiveCall& ) instead.

prefer 1) tho

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+
+Replayer::Replayer()
+{
+  filename = getenv("RCCL_REPLAY_FILE");
+
+  if (!filename)
+    return;
+
+  pid = getpid();
+  outputFile.open(std::string(filename) + std::to_string(pid), std::ofstream::binary);
```
cenzhaometa: in addition to pid, can we encode `rank` as part of filename, it's easier to read.

also see if we can use fmt::format("{}-{}-{}", filename, rank, pid); it's more performant and readable than +.

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+  //int numCall = -1; //placeholder for total # of calls
+  //outputFile.write((char*)&numCall, sizeof(int));
+
+}
+
+Replayer& Replayer::instance()
+{
+  static Replayer INSTANCE;
```
cenzhaometa: for internal private member variables, use lower case naming prefix with _: e.g  `_instance`?

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+  return INSTANCE;
+}
+
+void Replayer::rcclReplayerRecord(struct CollectiveCall& coll)
+{
+  Replayer& replayer = Replayer::instance();  //?
+
+  if (!replayer.filename) //vs !replayer.outputFile.is_open()
+    return;
```
cenzhaometa: nit: for if/else, let's always add {}. 

I know NCCL upstream doesn't do this (this can lead to unexpected behavior), for our new code let's strive for best coding practice :)

(same comments for other places)

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+
+  if (tid == -1)
+    tid = syscall(SYS_gettid);
+
+  int hipDev;
+  hipGetDevice(&hipDev);
+
+  coll.pid = replayer.pid;
+  coll.tid = tid;
```
cenzhaometa: what's this tid used for?

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+  coll.hipDev = hipDev;
+
+  std::unique_lock<std::mutex> lock(replayer.mtx); //this->mtx vs replayer.mtx vs mtx
+  replayer.colls.push_back(coll); //move needed?
```
cenzhaometa: yes, let's do std::move(coll); here

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+    tid = syscall(SYS_gettid);
+
+  int hipDev;
+  hipGetDevice(&hipDev);
+
+  coll.pid = replayer.pid;
+  coll.tid = tid;
+  coll.hipDev = hipDev;
+
+  std::unique_lock<std::mutex> lock(replayer.mtx); //this->mtx vs replayer.mtx vs mtx
```
cenzhaometa: prefer to add a scope to make critical section clear
```
{
  lock(..);
  // any operation in critical section
}
```

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+  replayer.colls.push_back(coll); //move needed?
+  replayer.numCall++;
+}
+
+ncclResult_t Replayer::rcclReplayerWrite()
```
cenzhaometa: shall we add some protection this API is supposed to be called only once?

src/misc/replayer.cc
```diff
@@ -0,0 +1,88 @@
...
+    replayer.outputFile << std::endl;
+  }
+
+  //replayer.outputFile.write((char*)&colls[0], sizeof(CollectiveCall) * colls.size());
+  colls.clear();
+
+  return ret;
+}
+
+Replayer::~Replayer()
```
cenzhaometa: in program exit unexpectedly, shall we call `writeFile` in destructor to dump whatever is captured.

tools/RcclReplayer/rcclReplayer.cpp
```diff
@@ -180,31 +180,33 @@ void dataToCsv(GroupCall const& gc, std::ofstream &datafile, double runTime)
...
+  std::ifstream fp(logFilename, std::ios::binary);
   if (!fp) {
     printf("[ERROR] Unable to open file %s\n", logFilename);
     exit(-1);
   }
 
-  char line[2048];
-  LineItem li;
+  const size_t size = sizeof(LineItem) + 1;
+  char line[size]; //size of collectivecall struct
```
cenzhaometa, line 194: in c++, can we just use std::string, not to deal with fixed length char arrays

tools/RcclReplayer/rcclReplayer.hpp
```diff
@@ -36,23 +36,23 @@
...
-  int    datatype;
-  int    op;
-  int    root;
-  char   comm[32];
-  int    nRanks;
-  void*  stream;
-  int    task;
-  int    globalRank;
+  //char   hostname[MPI_MAX_PROCESSOR_NAME];
+  int           pid;
```
cenzhaometa, line 40: initialized primitive types.

cenzhaometa (changes requested):
thanks @AtlantaPepsi for carrying out structured logging to improve replay quality. 

major comments are:
1) let's improve code quality using more modern c++ paradigms (e.g namespace, std::string instead of char[] etc, more detailed comments inline)
2) for each commit, we should add unit test (UT) in the same commit. for example, this commit introduce Replayer class, we can add a multi-thread test case making sure a) each thread is accessing the same singleton class, b) those API are indeed threadsafe c) perform basic addRecord/Write to examine the output file. you can use google test 

I'll review the rest once these are addressed.

src/include/replayer.h
corey-derochie-amd: > also for these fields, do we intent to change them once constructed, if not, let's mark them as const.

Instead of making the whole struct immutable, would it not be preferable to use it with proper const-correctness? In this case the only concern might be how the timestamp is being set.

AtlantaPepsi: these implicit fields will either be never called for auxiliary RCCL calls like `ncclGetVersion`, or be changed and updated for every log entry

tools/RcclReplayer/rcclReplayer.cpp
AtlantaPepsi: this is legacy code, just minimal change to make it run, we will renovate the actual replaying part in later patces

gilbertlee-amd (approved):

src/collectives.cc
```diff
@@ -98,15 +98,21 @@ ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sen
   NvtxParamsAllGather payload{sendcount * ncclTypeSize(datatype), datatype};
   NVTX3_FUNC_WITH_PARAMS(AllGather, AllGatherSchema, payload)
 
+  struct ncclInfo info = { ncclFuncAllGather, "AllGather",
+    sendbuff, recvbuff, sendcount, datatype, ncclSum, 0, comm, stream, /* Args */
+    ALLGATHER_CHUNKSTEPS, ALLGATHER_SLICESTEPS };
+
+  rcclApiCall call(info, rrAllGather);
```
gilbertlee-amd: We should try to avoid unnecessary construction if possible.
Let's delay the construction of this rcclApiCall to inside the Record() function

src/collectives.cc
```diff
@@ -98,15 +98,21 @@ ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sen
...
+
+  if (!mscclIsCaller()) // when msccl falls back to
+    NCCLCHECK(Replayer::instance().record(call));
```
gilbertlee-amd: Let's rename this to Recorder instead of Replayer to better match what we are doing.
Replayer will be the tool that uses what the Recorder produces.

src/enqueue.cc
```diff
@@ -2288,12 +2288,19 @@ ncclResult_t ncclRedOpCreatePreMulSum_impl(ncclRedOp_t *op, void *scalar, ncclDa
...
+  rcclApiCall call({.opName = "RedOpCreatePreMulSum", .sendbuff = scalar, .datatype = datatype,
+		    .op = *op, .root = residence, .comm = comm}, rrRedOpCreatePreMulSum);
+  NCCLCHECK(Replayer::instance().record(call));
   TRACE_CALL("ncclRedOpCreatePreMulSum(%d,%p,%d,%d,%p)", *op, scalar, datatype, residence, comm);
   return ncclSuccess;
 }
 
 NCCL_API(ncclResult_t, ncclRedOpDestroy, ncclRedOp_t op, ncclComm_t comm);
 ncclResult_t ncclRedOpDestroy_impl(ncclRedOp_t op, ncclComm_t comm) {
+  rcclApiCall call({.opName = "RedOpDestroy", .op = op, .comm = comm}, rrRedOpDestroy);
```
gilbertlee-amd: Having both opName and rrRedOpDestroy is redundant.
Having a table that allows for conversion between the enumeration and string name would be better.

src/include/replayer.h
```diff
@@ -0,0 +1,109 @@
...
+  rrReduceScatter = 3,
+  rrAllReduce = 4,
+  rrSend = 6,
+  rrRecv = 7,
+  rrAllToAll = 8,
+  rrAllToAllv = 9 ,
+  rrGather = 5,
+  rrScatter = 13,
+  rrBcast,
+  rrGroupStart,
```
gilbertlee-amd: This numbering is confusing.

gilbertlee-amd (changes requested):

src/collectives.cc
corey-derochie-amd: I second this suggestion.

src/collectives.cc
```diff
@@ -98,15 +98,19 @@ ncclResult_t ncclAllGather_impl(const void* sendbuff, void* recvbuff, size_t sen
   NvtxParamsAllGather payload{sendcount * ncclTypeSize(datatype), datatype};
   NVTX3_FUNC_WITH_PARAMS(AllGather, AllGatherSchema, payload)
 
+  struct ncclInfo info = { ncclFuncAllGather, "AllGather",
+    sendbuff, recvbuff, sendcount, datatype, ncclSum, 0, comm, stream, /* Args */
+    ALLGATHER_CHUNKSTEPS, ALLGATHER_SLICESTEPS };
+
+  if (!mscclIsCaller()) // when msccl falls back to
+    NCCLCHECK(Recorder::instance().record(info, rrAllGather));
```
gilbertlee-amd: Please make this more consistent with the other record() function calls with respect to arguments.  record(rrAllGather, info);

src/include/recorder.h
```diff
@@ -0,0 +1,125 @@
...
+  rrCommRegister,
+  rrCommDeregister,
+  rrMemAlloc,
+  rrMemFree,
+  rrRedOpCreatePreMulSum,
+  rrRedOpDestroy,
+  rrOtherCall
+} rcclCall_t;
+
+extern const char* rcclCallStr[];
```
gilbertlee-amd: This should be defined within the header, so that Replayer can have access to it, without having to link in recorder.cc

src/misc/recorder.cc
```diff
@@ -0,0 +1,517 @@
...
+using namespace std::chrono;
+__thread int Recorder::rcclReplayThreadIdx = -1;
+int Recorder::depth = 0;
+
+static char buffer[4096];
+static rcclCall_t lastcall = rrGroupStart; // for trailing comma, need modify for multithread case for Bcast
+void indent(int depth, std::ofstream& o) {for (int i = 0; i < depth; i++) o << " ";}
+void newLine(std::ofstream& o) {if (lastcall != rrGroupStart) o << ","; o << std::endl;}
+
+const char* rcclCallStr[30]
```
gilbertlee-amd: Add proper count into enumeration instead of hard-coding 30.

gilbertlee-amd (changes requested):

src/misc/recorder.cc
corey-derochie-amd: It's valid to give no number when you provide an initializer list.

src/include/recorder.h
corey-derochie-amd: Anything shared between the Recorder and the Replayer should be in its own separate common header if that's the case.

src/include/replayer.h
AtlantaPepsi: to rename

src/misc/recorder.cc
```diff
@@ -0,0 +1,483 @@
...
+
+  int hipDev;
+  hipGetDevice(&hipDev); // need later change to copy from comm
+
+  call.pid = pid;
+  call.tid = rcclReplayThreadIdx;
+  call.hipDev = hipDev;
+  return;
+}
+
```
AtlantaPepsi, line 128: comment here for single process use

AtlantaPepsi: potentially implement an async option of logging in case of performance impact hit

src/include/recorder.h
```diff
@@ -0,0 +1,153 @@
...
+  std::vector<rcclApiCall> calls;
+
+  void                  captureGpuContext(rcclApiCall& call);
+  void                  write(const rcclApiCall &call);
+  static void           recordLater(void* idx);
+  Recorder();
+  Recorder(const Recorder&) = delete;
+  Recorder& operator=(const Recorder&) = delete;
+  ~Recorder();
+
```
AtlantaPepsi, line 134: const function for anything not changing state of recorder

src/misc/replayer.cc
gilbertlee-amd: @cenzhaometa - Logfile for a process could have multiple communicators therefore multiple ranks, and rank isn't even necessarily known at time of file creation.  I don't think this is going to be possible.

We also currently are unable to utilize the 3rd party fmt library.

src/include/replayer.h
AtlantaPepsi: now we still need those struct info in calling function so can't really use move here, plus they're mostly primitives at the moment

src/collectives.cc
```diff
@@ -154,6 +164,10 @@ NCCL_API(ncclResult_t, ncclAllToAll, const void* sendbuff, void* recvbuff, size_
 
 ncclResult_t ncclAllToAll_impl(const void* sendbuff, void* recvbuff, size_t count, ncclDataType_t datatype,
   ncclComm_t comm, hipStream_t stream) {
+
+  if (!mscclIsCaller()) // when msccl falls back to
```
cenzhaometa, line 172: as we discussed last time, for better coding practice, let's use {} even tho there's only one line for if/else

https://google.github.io/styleguide/cppguide.html?fbclid=IwZXh0bgNhZW0CMTEAAR4H2geECtbfFskhJ9-guh_jDAz4VuC-U56xkGMk2__InCS3ZnABE2st3CnBPw_aem_mRYsTiHAJk1k4ySfAfF_Ow#Formatting_Looping_Branching

please also consider adding these linters to the codebase

Same comments for other places.

test/_RecorderTests.cpp
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+
+#include "RcclMockFuncs.hpp"
+//#include "TestBed.hpp"
```
cenzhaometa, line 11: remove these unused commented out line?

test/common/RcclMockFuncs.hpp
```diff
@@ -0,0 +1,12 @@
+#include "info.h"
+#include "comm.h"
+//#define
```
cenzhaometa: same here and other places

tools/RcclReplayer/rcclReplayer.hpp
cenzhaometa: yes, please address comments. 

this can lead to namespace pollution

cenzhaometa (approved):

gilbertlee-amd (approved):

PR #1618, opened 2025-03-25
all_gather LL128 multi-node tuning and including LL cutoff points in tuning models

mustafabar:
## Details

**Work item:** _"Internal"_

**What were the changes?**  
- Make LL and LL128 params part of the tuning models to link it to arch detection.
- Tune all_gather for LL128 on MI300

More ring/tree-based collectives can be easily added. 

- Added new fields to `struct tuningModel`.
- Through study of the all_gather LL/LL128 cutoff points

**Additional Documentation:**  

## Approval Checklist


PedramAlizadeh (approved):

src/include/comm.h
```diff
@@ -486,6 +486,7 @@ struct ncclComm {
   float bandwidths[NCCL_NUM_FUNCTIONS][NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS];
   float ringbdw[NCCL_NUM_FUNCTIONS][NCCL_NUM_PROTOCOLS];
   int maxThreads[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS];
+  uint64_t minMaxLLRange[NCCL_NUM_FUNCTIONS][NCCL_NUM_PROTOCOLS - 1][2];
```
gilbertlee-amd, line 520: Is this something that really requires to be in every communicator?
Or can this be a static member of a RcclTuning class?

src/enqueue.cc
```diff
@@ -1727,19 +1719,38 @@ static ncclResult_t topoGetAlgoInfo(
...
+    } else if (IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx942")) {
+      // Warn that model detection for MI300 (or future others) did not work as expected
+      // Add supported archs to this condition as they come (e.g. gfx950)
+      // Also make sure the tuning_model and model detection are updated for new archs
+      static bool failedWarn = false;
+      if (!failedWarn) {
+        WARN("LL cutoff points not detected for a supported arch %s", comm->topo->nodes[GPU].nodes[0].gpu.gcn);
+        failedWarn = true;
+      }
+    }
```
gilbertlee-amd: RCCL specific code-modifications should be moved into a separate file, if possible

src/enqueue.cc
```diff
@@ -1727,19 +1719,38 @@ static ncclResult_t topoGetAlgoInfo(
...
+    auto llMin = comm->minMaxLLRange[info->func][NCCL_PROTO_LL][0];
+    auto llMax = comm->minMaxLLRange[info->func][NCCL_PROTO_LL][1];
+
+    auto ll128Min = comm->minMaxLLRange[info->func][NCCL_PROTO_LL128][0];
+    auto ll128Max = comm->minMaxLLRange[info->func][NCCL_PROTO_LL128][1];
+
+    // Only override model choices if min/max cutoff points are set in the tuning models
+    if((ll128Max > ll128Min) || (llMax > llMin)) {
+      // Keep it simple unless otherwise required
+      info->protocol = NCCL_PROTO_SIMPLE;
```
gilbertlee-amd: It's not immediately obvious why these conditions would result in the choice of simple.  Based on the choice of variable name, wouldn't the expectation be that Max > Min?

gilbertlee-amd (changes requested):

src/include/comm.h
mustafabar: It is not required as of now, but all static values are kept inside src/graph/tuning.cc and copied on a per-communicator from this array https://github.com/ROCm/rccl/blob/aace4e27f8b22c198cf5fc433c37810ab3aae802/src/graph/tuning.cc#L254

This is a NCCL design choice, and in the future there may be need to adjust per communicator based on nranks, nnodes etc.

src/enqueue.cc
mustafabar: This is updated now

src/enqueue.cc
mustafabar: Based on our discussion, we need to outline how to structure this for RCCL contributors to ensure effective future use. We can create a separate PR for this.

gilbertlee-amd (approved):

PR #1627, opened 2025-04-03
all_reduce LL/LL128 and Ring/Tree multi-node tuning for MI300

PedramAlizadeh:
## Details

**What were the changes?**  

- Adjust tree/ring tuning model parameters to address the AllReduce Tree->Ring switching point for large messages (above 64 MB) and get a performance boost up to 1.25x for the impacted message sizes. 

- The LL/LL128 cutoff range for AR was tuned and included in the MI300 rome model

- Move the tuning changes to a separate source file to maintain RCCL-only changes

Suboptimal tree usage for larger messages where ring works better
Modified tuning model parameters for MI300x and also, enabled LL128 on MI300 by default
**Additional Documentation:**  

## Approval Checklist


thananon (approved):
Looks good to me 👍

src/enqueue.cc
```diff
@@ -1946,27 +1946,37 @@ static ncclResult_t topoGetAlgoInfo(
...
     auto ll128Max = comm->minMaxLLRange[info->func][NCCL_PROTO_LL128][1];
 
     // Only override model choices if min/max cutoff points are set in the tuning models
-    if((ll128Max != RCCL_LL_LIMITS_UNDEFINED) || (llMax != RCCL_LL_LIMITS_UNDEFINED)) {
+    if ((ll128Max != RCCL_LL_LIMITS_UNDEFINED) || (llMax != RCCL_LL_LIMITS_UNDEFINED)) {
       // Keep it simple unless otherwise required
       info->protocol = NCCL_PROTO_SIMPLE;
-      // Normalize the comparison to sizePerRank as this is essentially what matters in determining protocol choice
+      // Normalize the comparison to sizePerRank as this is essentially what matters in determining protocol choice in RS and AG cases
       size_t sizePerRank = nBytes / comm->nRanks;
```
dmwu: Code can be simplified ?
`size_t sizePerRank = info->func == ncclFuncAllGather  || info->func == ncclFuncAllReduce ? nBytes / comm->nRanks : nBytes`
and do the same LL proto check for all functions

dmwu: For RS & AG, the send/recv msg size is nRanks * msg_size. What does sizePerRank mean here? I think it also assumes we are using the ring algo? So sizePerRank is essentially the bytes a rank send/recv in a ring step?

src/enqueue.cc
```diff
@@ -1946,27 +1946,37 @@ static ncclResult_t topoGetAlgoInfo(
...
+      else if (sizePerRank <= llMax && sizePerRank > llMin) {
         info->protocol = NCCL_PROTO_LL;
       }
 #if defined(ENABLE_LL128)
       // When applicable, LL128 RS performance is better than LL, so the next condition overrides the previous LL choice
-      if(comm->topo->ll128Enabled) {
-        if(sizePerRank <= ll128Max && sizePerRank > ll128Min) {
+      if (comm->topo->ll128Enabled) {
+        if (info->func == ncclFuncAllReduce) {
+          if (nBytes <= (ll128Max+(log2i(comm->nNodes)-1)*3145728) && nBytes > ll128Min) {
```
dmwu: Any reasoning behind this magic calculations?

src/enqueue.cc
mustafabar: I have inlined documentation on the meaning of sizePerRank. It maps to the send count for AG and recv count for RS, and send/recv cound for AR

src/enqueue.cc
mustafabar: The maximum size where LL64/128 benefits is logarithmic in the number of nodes used. The log factor is around 3MB. There is a chance of slight variance across MI300x clusters, which we can assess.

src/enqueue.cc
mustafabar: done

src/graph/rome_models.cc
```diff
@@ -814,7 +814,7 @@ static struct rcclRomeModel rome_model_81 = {
                 "N7 7 3 2 6 0 4 1 5 N5|"
                 "N1 1 0 2 4 3 5 7 6 N6|",
 
-  .options    = "noCpuCheck=1,tuning=5,disableNumaMatching=1",
+  .options    = "noCpuCheck=1,tuning=5,ll128Enabled=1,disableNumaMatching=1",
```
mustafabar, line 817: @PedramAlizadeh is this the only MI300x + 8NIC model configuration? Pls check.

src/include/comm.h
```diff
@@ -710,7 +710,7 @@ inline ncclResult_t ncclCommPollEventCallbacks(struct ncclComm *comm) {
     }
   }
 finish:
-  cudaThreadExchangeStreamCaptureMode(&mode);
+  CUDACHECK(cudaThreadExchangeStreamCaptureMode(&mode));
```
dmwu, line 714: Is this change related to AR perf tuning?

wenkaidu (approved):

src/enqueue.cc
```diff
@@ -1939,54 +1939,11 @@ static ncclResult_t topoGetAlgoInfo(
     time = backupTime;
   }
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
```
gilbertlee-amd: We should be even able to remove this #ifdef

src/include/comm.h
gilbertlee-amd: This minor change to use CUDACHECK should be fine.  It happens as part of NCCL v2.24.3-1 anyways.

src/include/rccl_common.h
```diff
@@ -0,0 +1,53 @@
...
+*/
+#ifndef RCCL_COMMON_H_
+#define RCCL_COMMON_H_
+#include "nccl_common.h"
+
+#define RCCL_TUNABLE_COLLS 4        // LL/LL64/LL128 tunable Collective count
+#define RCCL_RS_TUNABLE 0           // reduce_scatter index
+#define RCCL_AG_TUNABLE 1           // all_gather index
+#define RCCL_AR_TUNABLE 2           // all_reduce index
+#define RCCL_RE_TUNABLE 3           // reduce index
```
gilbertlee-amd: Would this be better as an enumeration rather than define?

src/include/rccl_common.h
```diff
@@ -0,0 +1,53 @@
...
+#define RCCL_LL_LIMITS_UNDEFINED 0
+#define RCCL_PROTOCOL_ENTRY_SIZE 3
+#define RCCL_PROTOCOL_MIN_IDX 0
+#define RCCL_PROTOCOL_MAX_IDX 1
+#define RCCL_PROTOCOL_FACTOR_IDX 2
```
gilbertlee-amd, line 39: Is there justification for the need for this type of indexing?
Rather than just introducing a struct with named fields to make code clear?

src/rccl_wrap.cc
```diff
@@ -0,0 +1,77 @@
...
+
+    // Only override model choices if min/max cutoff points are set in the tuning models
+    if ((ll128Max != RCCL_LL_LIMITS_UNDEFINED) || (llMax != RCCL_LL_LIMITS_UNDEFINED)) {
+      // Keep it simple unless otherwise required
+      info->protocol = NCCL_PROTO_SIMPLE;
+      // Normalize the comparison to sizePerRank as this is essentially what matters in determining protocol choice in RS and AG cases
+      // For AG, this is the send size per rank
+      // For RS, this is the recv size per rank
+      // For AR, this is the send/recv size per rank
+      size_t sizePerRank = (info->func == ncclFuncReduceScatter || info->func == ncclFuncAllGather)? nBytes / comm->nRanks : nBytes;
```
gilbertlee-amd: Recommend having a separate function for this

gilbertlee-amd (changes requested):

src/include/rccl_common.h
mustafabar: Good suggestion. I will have a separate PR for it when this one is merged, as it has side effects in several places. The initial reason was to fit in the arrays or arrays scheme of NCCL rather than an array of structs. But since we can control this part, named struct fields would make more sense

src/rccl_wrap.cc
mustafabar: done

src/include/rccl_common.h
mustafabar: done

src/graph/rome_models.cc
PedramAlizadeh: yes, this is true.

gilbertlee-amd (approved):

PR #1643, opened 2025-04-16
Enable `-Wdeprecated-copy-with-user-provided-copy`

r-barnes:
## Details
___Do not mention proprietary info or link to internal work items in this PR.___

**Work item:** _"Internal", or link to GitHub issue (if applicable)._

**What were the changes?**  
Added copy constructors. Enabled a new warning flag.

**Why were the changes made?**  
Having `operator=` without a copy constructor, or a copy constructor without `operator=` is deprecated in C++ and will cause code to fail to compile with newer standards. I've made appropriate fixes and enabled `-Wdeprecated-copy-with-user-provided-copy` to prevent regressions.

**How was the outcome achieved?**  
N/A

**Additional Documentation:**  
N/A

## Approval Checklist
___Do not approve until these items are satisfied.___
- [X] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #1647, opened 2025-04-17
Explicitly define a copy constructor for rccl_float8 and rccl_bfloat8

dmwu:
## Details

**What were the changes?** 
 
This is to pass LLVM build with -Wdeprecated-copy-with-user-provided-copy.
w/o the fix, we have following compile error:
`error: definition of implicit copy constructor for 'rccl_float8' is deprecated because it has a user-provided copy assignment operator [-Werror,-Wdeprecated-copy-with-user-provided-copy]
  495 |     inline __host__ __device__ rccl_float8& operator=(const rccl_float8& a)
rccl_float8.h:673:16: note: in implicit copy constructor for 'rccl_float8' first required here
  673 |         return rccl_float8(sinf(float(a)));`

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/include/rccl_float8.h
+++ b/src/include/rccl_float8.h
@@ -492,7 +492,7 @@ struct rccl_float8
     }
 
     // Explicitly define a copy constructor because copy assignment is defined
-    constexpr rccl_float8(const rccl_float8& o) : data(o.data) {}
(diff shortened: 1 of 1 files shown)


r-barnes:
#1643 might be a better choice since it enables a flag that would prevent regressions.

PR #1648, opened 2025-04-17
Cherry-pick (Add device synchronization before destroying proxy thread. (#1631))

corey-derochie-amd:
Cherry-pick of #1631 into ROCm 6.4.1

### DIFF
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -1566,6 +1566,7 @@ void* ncclProxyService(void* _args) {
   }
 
   // Wait for all operations to complete and stop progress thread before freeing any resource
+  hipDeviceSynchronize();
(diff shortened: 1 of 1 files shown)


nileshnegi (approved):

corey-derochie-amd:
@thananon It would be appropriate to add a note to the CHANGELOG for this cherry-pick.

thananon:
I guess there is a reason why we cherry-pick this into 6.4.1? We can add

> Add synchronization before destroying proxy thread to fix a rare hang caused by early termination.

amd-jnovotny:
@corey-derochie-amd and @thananon The changelog entry wasn't added. It would be best to add one. The following PR is still open for 6.4 and can maybe be added there: https://github.com/ROCm/rccl/pull/1654

PR #1650, opened 2025-04-18
[AllGather MSCCL] Multinode and single node support up to certain send count 

mustafabar:
## Details

**Work item:** _Internal_

**What were the changes?**  
- Generated in-place and out-of-place XMLs with LL128 MSCCL support and up to 32 GPUs (64 GPUs is WIP)
- Fixed the issue in single node XML that caused UT failure and therefore resulted in removing the AG msccl support

- MSCCL uses RCCL/NCCL primitives and is effective in latency ranges 
- MSCCL can be used to test out LogP algorithms

- Added Python support for out-of-place operation
- Debugged the issue causing the UT failure

**Additional Documentation:**  
- To use in multinode setting, `RCCL_MSCCL_FORCE_ENABLE=1` must be set
- MSCCL is used up to 49168 send count if we consider the fp32 case. **To generalize in terms of AG message size, this translates to 12292 * sizeof(datatype) * nGPUs bytes**
- Up to the mentioned range, 2 - 3x gains are observed

## Approval Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 * `RCCL_SOCKET_REUSEADDR` and `RCCL_SOCKET_LINGER` environment parameters
 * Setting `NCCL_DEBUG=TRACE NCCL_DEBUG_SUBSYS=VERBS` will generate traces for fifo and data ibv_post_sends
 * Added `--log-trace` flag to enable traces through the install.sh script (e.g. `./install.sh --log-trace`)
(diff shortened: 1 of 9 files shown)


wenkaidu (approved):

nusislam:
Which allgather algorithm did you use to generate the xmls? Direct or recursive doubling?

mustafabar:
> Which allgather algorithm did you use to generate the xmls? Direct or recursive doubling?

I used direct. In practice, RD didn't perform better. MSCCL RD implementation may need optimization.

amd-jnovotny:
@mustafabar I see the changelog entry shows this for 6.4, but it's too late for that. Is this intended for 6.4.1? If so, can you make the proper adjustment for the 6.4 release branch cherry-pick PR. I would like to have the chance to review at that time.

PR #1664, opened 2025-04-28
Added ENABLE_CODE_COVERAGE option to facilitate code coverage during builds.

atulkulk:
**Work item:** Internal

**What were the changes?**  
Added ENABLE_CODE_COVERAGE option to facilitate code coverage during builds.

The motivation behind this change is to expose internal library symbols for unit testing the RCCL library, avoiding the need to include source files directly in the RCCL unit test binary, which introduces maintenance overhead. Below is a table detailing the behavior of the options based on their configuration:

| BUILD_TESTS | ENABLE_CODE_COVERAGE | BUILDTYPE               | VISIBILITY            |
|-------------|-----------------------|-------------------------|-----------------------|
| ON          | ON                    | Debug | "-fvisibility=default" + Linker options |
| ON          | OFF                   | Release, RelWithDebInfo | "-fvisibility=hidden"|
| ON          | OFF                   | Debug | ROCM_VERSION >=60400 ? "-fvisibility=default" : "-fvisibility=hidden"|
| OFF         | ON                    | Debug | "-fvisibility=default" + Linker options |
| OFF         | OFF                   | Release, Debug, RelWithDebInfo | "-fvisibility=hidden" |

Integrated compilation and linker flags to enable code coverage metric generation during unit test execution.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(BUILD_LOCAL_GPU_TARGET_ONLY             "Build only for GPUs detected on 
 option(BUILD_SHARED_LIBS                       "Build as shared library"                       ON)
 option(BUILD_TESTS                             "Build unit test programs"                      OFF)
 option(COLLTRACE                               "Collective Trace Option"                       ON)
+option(ENABLE_CODE_COVERAGE                    "Enable code coverage"                          OFF)
 option(ENABLE_MSCCL_KERNEL                     "Enable MSCCL while compiling"                  ON)
 option(ENABLE_MSCCLPP                          "Enable MSCCL++"                                ON)
 option(ENABLE_MSCCLPP_CLIP                     "Enable MSCCL++"                                OFF)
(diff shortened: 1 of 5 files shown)


gilbertlee-amd (approved):

CMakeLists.txt
```diff
@@ -957,7 +957,7 @@ else()
   message(STATUS "Building shared RCCL library")
 endif()
 if (HAVE_KERNARG_PRELOAD)
-  target_link_options(rccl PRIVATE "SHELL:-Xoffload-linker -mllvm=-amdgpu-kernarg-preload-count=16")
+  target_link_options(rccl PRIVATE "SHELL:-Xoffload-linker -mllvm=-amdgpu-kernarg-preload-count=8")
```
nileshnegi: historically, we were told to use 16 for perf.

CMakeLists.txt
```diff
@@ -823,8 +824,40 @@ target_compile_options(rccl PRIVATE -Werror=uninitialized)
 target_compile_options(rccl PRIVATE -Werror=sometimes-uninitialized)
 target_compile_options(rccl PRIVATE -Wall)
 target_compile_options(rccl PRIVATE -Wno-format-nonliteral)
-target_compile_options(rccl PRIVATE -fgpu-rdc)               # Generate relocatable device code (required for extern __shared__)
-target_compile_options(rccl PRIVATE -fvisibility=hidden)     # Set symbol visibility to hidden by default
+target_compile_options(rccl PRIVATE -fgpu-rdc)
+
+## Set RCCL compile and linker options for code coverage build
+if(ENABLE_CODE_COVERAGE)
+  if(NOT CMAKE_BUILD_TYPE MATCHES "Debug|RelWithDebInfo")
```
nileshnegi: do we need to limit to these cmake build types? will the profile/coverage flags and `-g` not work with `Release` builds?

CMakeLists.txt
atulkulk: Agreed, but I was not able to reproduce the build issue locally. I am testing this change out on the CI to check whether it helps fix the issue. If it does, then we will have to take a call on this. Do you have any pointer to reproduce build issue locally? I used exactly the same arguments which the CI build uses.

CMakeLists.txt
atulkulk: The initial idea here was to limit it to debug build types only because ideally code coverage is run without any compiler optimizations. But we do not have to limit it those build types. I will remove that condition.

CMakeLists.txt
atulkulk: Never mind, I was able to reproduce the build issue locally. Reduction in preload count didn't help. I will investigate into it.

test/common/RcclMockFuncs.hpp
```diff
@@ -1,7 +1,15 @@
...
+#ifndef RCCL_EXPOSE_STATIC
+
 #include "info.h"
 #include "comm.h"
 
 void ncclDebugLog(ncclDebugLogLevel, unsigned long, char const*, int, char const*, ...) {};
 ncclResult_t getHostName(char* hostname, int maxlen, const char delim) {
   return ncclSuccess;
 }
+#endif  // RCCL_EXPOSE_STATIC
```
corey-derochie-amd: I think it should be the responsibility of the code including this header to exclude it if necessary. These checks are duplicated elsewhere.

Also, these includes are not required within this header code, so should be removed and only included by the code requiring them.

```suggestion
void ncclDebugLog(ncclDebugLogLevel, unsigned long, char const*, int, char const*, ...) {};
ncclResult_t getHostName(char* hostname, int maxlen, const char delim) {
  return ncclSuccess;
}
```

test/_RecorderTests.cpp
```diff
@@ -7,8 +7,12 @@
 #include <gtest/gtest.h>
 #include <rccl/rccl.h>
 
+#ifndef RCCL_EXPOSE_STATIC
 #include "RcclMockFuncs.hpp"
-//#include "TestBed.hpp"
+#else
+#include "info.h"
+#include "comm.h"
+#endif
```
corey-derochie-amd: This code requires info.h and comm.h, so should always include them. It shouldn't need to know about RcclMockFuncs.hpp internals to make that decision.

```suggestion
#endif

#include "info.h"
#include "comm.h"
```

corey-derochie-amd (approved):
LGTM, with minor suggestions.

test/common/RcclMockFuncs.hpp
atulkulk: I have made modifications per your comment. `info.h` is still require in this file to get access to `ncclDebugLogLevel` parameter.

test/_RecorderTests.cpp
atulkulk: Made modifications as per your suggestion.

PR #1665, opened 2025-04-29
Code coverage improvements

atulkulk:
**Work item:** Internal

**What were the changes?**  
Added new unit tests to cover test cases that were not covered by the functional tests for alt_rsmi.cc & argcheck.cc files.

In alt_rsmi.cc, all the static functions were migrated to an internal namespace so that they can be accessed using default visibility in the unit test binary. For argcheck.cc, new test cases were added. All this new unit tests were written using the setup and teardown fixtures. These tests did not work well with the tests run using TestBed framework. A separate `rccl-UnitTestsFixtures` binary was created to avoid clobbering with the existing test execution. 

| **alt_rsmi.cc** | Initial | 90.91% (20/22)    | 77.28% (381/493) | 73.12% (234/320) | 73.37% (135/184) |
|                 | Updated | 95.45% (21/22)    | 97.98% (485/495) | 97.19% (311/320) | 93.33% (168/180) |
| **arg_check.cc**| Initial | 75.00% (3/4)      | 30.16% (19/63)   | 29.55% (39/132)  | 20.83% (15/72)   |
|                 | Updated | 100.00% (4/4)     | 100.00% (63/63)  | 92.42% (122/132) | 66.67% (48/72)   |

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -715,6 +715,57 @@ foreach(SRC_FILE ${SRC_FILES})
   endif()
 endforeach()
 
+# Adding custom target to hipify all the source files
+# This is required to make sure that all the hipified source files are
+# available before compiling the unit tests executable(s)
+add_custom_target(hipify_all DEPENDS ${HIP_SOURCES})
+
+if (BUILD_TESTS)
+  if (ROCM_VERSION VERSION_GREATER_EQUAL "60400" AND CMAKE_BUILD_TYPE MATCHES "Debug")
+    set(HIPIFY_SRC_DIR "${PROJECT_BINARY_DIR}/hipify/src")
+    set(REPLACE_SCRIPT "${CMAKE_SOURCE_DIR}/tools/scripts/replace_static.sh")
+    message ("Replacing static functions in ${HIPIFY_SRC_DIR} with ${REPLACE_SCRIPT} for unit tests")
+    # Create a list of files which needs to be modified to remove static
+    set(TEST_NONSTATIC_SOURCE_FILES
+      ${HIPIFY_SRC_DIR}/misc/alt_rsmi.cc
+    )
+    # Create a custom command to backup the original files and remove static
+    # Always run replace script on hipified files, but preserve original backups
+    foreach(srcfile ${TEST_NONSTATIC_SOURCE_FILES})
+      add_custom_command(
+        OUTPUT "${srcfile}.staticbak"
+        COMMAND bash -c "\
+            ${CMAKE_COMMAND} -E echo 'Processing ${srcfile} for static replacement' && \
+            if [ ! -f '${srcfile}.staticbak' ]; then \
+              ${CMAKE_COMMAND} -E copy '${srcfile}' '${srcfile}.staticbak' && \
+              ${CMAKE_COMMAND} -E echo 'Created backup: ${srcfile}.staticbak'; \
+            fi && \
+            ${CMAKE_COMMAND} -E echo 'Exposing internal functions/variables in ${srcfile}' && \
+            ${REPLACE_SCRIPT} ${srcfile} --replace-vars --exclude-list=${exclude_list} 2>&1 \
+          "
+        DEPENDS "${srcfile}" "${REPLACE_SCRIPT}" hipify_all
+        COMMENT "Removing static from ${srcfile} (backup preserved at ${srcfile}.staticbak)"
+        VERBATIM
+      )
+      list(APPEND STATIC_BAK_FILES "${srcfile}.staticbak")
+    endforeach()
+    add_custom_target(replace_static_in_hipify ALL DEPENDS ${STATIC_BAK_FILES})
+    add_dependencies(replace_static_in_hipify hipify_all)
+  else()
+    # Restore originals in the build directory if backup exists
+    foreach(srcfile ${HIPIFY_SRC_FILES})
+      if(EXISTS "${srcfile}.staticbak")
+        execute_process(
+          COMMAND ${CMAKE_COMMAND} -E copy "${srcfile}.staticbak" "${srcfile}"
+        )
+        execute_process(
+          COMMAND ${CMAKE_COMMAND} -E remove "${srcfile}.staticbak"
+        )
+      endif()
+    endforeach()
+  endif()
+endif()
+
 # Generate device/host tables and all the collective functions that are going to be in librccl.so
 #==================================================================================================
 find_package(Python3 COMPONENTS Interpreter REQUIRED)
@@ -768,6 +819,11 @@ add_library(rccl ${HIP_SOURCES})
 
 ## Set RCCL dependencies
 add_dependencies(rccl git_version_check)                                      # Execute git_version_check during build
+
+if (BUILD_TESTS AND ROCM_VERSION VERSION_GREATER_EQUAL "60400" AND CMAKE_BUILD_TYPE MATCHES "Debug")
+  ## Set static replacement dependency for fixture unit tests
+  add_dependencies(rccl replace_static_in_hipify)
+endif()
 
 ## Set RCCL include directories
 target_include_directories(rccl PRIVATE ${PROJECT_BINARY_DIR}/include)        # for generated rccl.h header
@@ -986,7 +1042,7 @@ if (HAVE_PARALLEL_JOBS)
   target_compile_options(rccl PRIVATE -parallel-jobs=12)
 endif()
 
-if (ROCM_VERSION VERSION_GREATER_EQUAL "60200") 
+if (ROCM_VERSION VERSION_GREATER_EQUAL "60200")
   target_compile_options(rccl PRIVATE --offload-compress)    # Compress GPU code at compile time.
   target_link_libraries(rccl PRIVATE --offload-compress)     # Compress GPU code at link time.
   message(STATUS "--offload-compress enabled - ROCm version >= 6.2.0")
@@ -999,7 +1055,7 @@ target_compile_options(rccl PRIVATE -Werror=sometimes-uninitialized)
 target_compile_options(rccl PRIVATE -Wall)
 target_compile_options(rccl PRIVATE -Werror=deprecated-copy-with-user-provided-copy)
 target_compile_options(rccl PRIVATE -Wno-format-nonliteral)
-target_compile_options(rccl PRIVATE -fgpu-rdc)
+target_compile_options(rccl PRIVATE -fgpu-rdc)      # Generate relocatable device code (required for extern __shared__)
 
 ## Set RCCL compile and linker options for unit tests and code coverage
 if(ENABLE_CODE_COVERAGE)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -678,6 +679,11 @@ foreach(SRC_FILE ${SRC_FILES})
   endif()
 endforeach()
 
+# Adding custom target to hipify all the source files
+# This is required to make sure that all the hipified source files are
+# available before compiling the unit tests executable(s)
+add_custom_target(hipify_all DEPENDS ${HIP_SOURCES})
```
corey-derochie-amd, line 721: We should have done this a long time ago, I think.

install.sh
```diff
@@ -445,15 +453,25 @@ fi
 
 # Optionally, run RCCL-UnitTests, if they're enabled.
 if [[ "${run_tests}" == true ]]; then
-    if [[ -x "./test/rccl-UnitTests" ]]; then
-        if [[ "${run_tests_all}" == true ]]; then
+    if [[ ! -x "./test/rccl-UnitTests" ]]; then
+        echo "RCCL-UnitTests and RCCL-UnitTestsFixtures have not been built yet; Please re-run script with \"-t\" to build the binary."
```
corey-derochie-amd:
```suggestion
        echo "RCCL-UnitTests have not been built yet; Please re-run script with \"-t\" to build the binary."
```

tools/scripts/replace_static.sh
```diff
@@ -0,0 +1,46 @@
...
+fi
+
+TMP_FILE="${SOURCE_FILE}.tmp.$$"
+
+echo "[INFO] Replacing 'static inline' with 'inline' in $SOURCE_FILE"
+sed -E 's/\bstatic[[:space:]]+inline\b/inline/g' "$SOURCE_FILE" > "$TMP_FILE"
+
+echo "[INFO] Replacing 'static' at function start in $SOURCE_FILE"
+# This matches 'static' at the start of a line (possibly with spaces), followed by anything, up to '('
+sed -E -i 's/^([[:space:]]*)static([[:space:]]+)([a-zA-Z_][a-zA-Z0-9_:<>~ \*&]*)\(/\1\3(/g' "$TMP_FILE"
```
corey-derochie-amd: We need to verify that nowhere have we put the `static` keyword on a separate line...

corey-derochie-amd: Function name shouldn't contain spaces, but there could be whitespace after the function name.
```suggestion
sed -E -i 's/^([[:space:]]*)static([[:space:]]+)([a-zA-Z_][a-zA-Z0-9_:<>~\*&]*)([[:space:]]*)\(/\1\3(/g' "$TMP_FILE"
```

I can imagine a lot of ways this pattern could fail, it might be fragile. Perhaps consider if there's another way.

tools/scripts/replace_static.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [[ "$REPLACE_VARS" == "1" ]]; then
+  echo "[INFO] Replacing 'static' at variable definitions in $SOURCE_FILE"
+  # This matches 'static' at the start of a line (possibly with spaces), followed by a type and a variable name
+  sed -E -i 's/^([[:space:]]*)static([[:space:]]+)/\1/g' "$TMP_FILE"
```
corey-derochie-amd, line 83: This will capture static variables declared within function bodies and fields within structs, which is probably not what you want.

tools/scripts/replace_static.sh
corey-derochie-amd: There are functions where other keywords appear ahead of `static`, which this would not capture.

tools/scripts/replace_static.sh
atulkulk: For one of the current file (rccl_wrap.cc) which was used to write new UTs, replacing static vars within function bodies was required. Hence, I have added an option whether one wants to replace vars as well. Once we reach a certain file where replacing vars is not required, we can have a separate process just to replace functions and add it to the list of files for compilation.

tools/scripts/replace_static.sh
atulkulk: I have modified the regex to handle almost all of the cases for function capture.

Here is the output file for `src/` directory where the regex did not replaced the static keyword: 
[remaining_static_src.txt](https://github.com/user-attachments/files/20637150/remaining_static_src.txt)

Here is the output file for `src/` directory where it applied all the regexes to replace static keyword:
[remove_static_src.diff.txt](https://github.com/user-attachments/files/20637155/remove_static_src.diff.txt)

venksubramd:
CudaPtrCheck
TEST_F(ArgCheckTest, CudaPtrCheck_InvalidScenarios) {
This test actually has three tests in it. In general, it is better to avoid multiple independent asserts/tests in one test. The reason is if one assert fails, then the next asserts that are not dependent do not execute. It is OK to have dependent asserts in one test, for example, two asserts that check for two separate values from a single call to a function. It would be better in our case to separate the single test into three tests, one for each of the cases mentioned.

The three tests embedded in this one test are pretty good. Do we need one extra test for the type being different from the cudaMemoryTypeDevice?

PtrCheck
The tests are adequate for this function.

CommCheck
It would be good to split this test into two tests since the condition in the function is a or (||) and not an and (&&). The first test should corrupt startMagic and the second test should corrupt the endMagic.

ArgCheck
TEST_F(ArgCheckTest, ArgsCheck_InvalidInputs) {
We would want to separate this test into multiple tests for the same reason as mentioned in the top, to avoid multiple independent asserts in one test.

Test for invalid root should include two tests, one for value < 0 and the other where the value is info->comm->nRanks?

Likewise for two different range values for info->datatype?

It will be useful to revisit the different ranges for each condition and have two tests for both the bounds.

The remaining tests are good, but need to be split into multiple tests to avoid independent asserts in one test.

Do we need tests for the public API functions in alt_rsmi.cc?

CMakeLists.txt
```diff
@@ -715,6 +715,57 @@ foreach(SRC_FILE ${SRC_FILES})
...
+  set(TEST_NONSTATIC_SOURCE_FILES
+    ${HIPIFY_SRC_DIR}/misc/alt_rsmi.cc
+  )
+  # Create a custom command to backup the original files and remove static
+  # Only create backup and run replace script if backup doesn't already exist
+  foreach(srcfile ${TEST_NONSTATIC_SOURCE_FILES})
+    add_custom_command(
+      OUTPUT "${srcfile}.staticbak"
+      COMMAND bash -c "
+        if [ ! -f '${srcfile}.staticbak' ]; then \\
```
peizhang56: This does not work for the cmake build, only works for ninja.
`*** missing separator (did you mean TAB instead of 8 spaces?).  Stop.`

test/ArgCheckTests.cpp
```diff
@@ -0,0 +1,255 @@
...
+    info.coll = ncclFuncBroadcast;     // Valid collective operation
+    info.comm = &comm;                 // Valid communicator
+    info.sendbuff = sendDevicePtr;     // Use allocated device pointer for send buffer
+    info.recvbuff = recvDevicePtr;     // Use allocated device pointer for receive buffer
+    info.count = 10;                   // Valid count
+    info.opName = "TestOp";            // Valid operation name
+
+    // Test case 1: Invalid reduction operation (out of range)
+    info.op = (ncclRedOp_t)5; // Invalid reduction operation (out of range)
+    ncclResult_t result = ArgsCheck(&info);
```
peizhang56: Having an intermittent coredump here.
image
The reason is userRedOps and userRedOpCapacity are not initialized.

CMakeLists.txt
```diff
@@ -734,11 +734,11 @@ if (BUILD_TESTS)
     add_custom_command(
       OUTPUT "${srcfile}.staticbak"
       COMMAND bash -c "
-        if [ ! -f '${srcfile}.staticbak' ]; then \\
-          ${CMAKE_COMMAND} -E copy '${srcfile}' '${srcfile}.staticbak' && \\
-          ${CMAKE_COMMAND} -E echo 'Exposing internal functions/variables in ${srcfile}' && \\
+        if [ ! -f '${srcfile}.staticbak' ]; then
```
peizhang56: @rahulvaidya20 I think this does not work for the cmake 
image

peizhang56: I have a working version here for the cmake
https://github.com/peizhang56/rccl/commit/d854f01d8cf6f2c3f2facef5dbb6cb5cb752ad14

test/CMakeLists.txt
```diff
@@ -86,8 +86,19 @@ if(BUILD_TESTS)
...
     )
 
-    # Add rccl-UnitTestsFixtures binary
-    add_executable(rccl-UnitTestsFixtures ${TEST_FIXTURE_SOURCE_FILES})
+
+  # Add rccl-UnitTestsFixtures binary
+  add_executable(rccl-UnitTestsFixtures ${TEST_FIXTURE_SOURCE_FILES})
+    
+  ## Set rccl-UnitTests linked libraries
+  target_link_libraries(rccl-UnitTests PRIVATE ${GTEST_BOTH_LIBRARIES})
```
peizhang56: I believe this is done in the foreach loop below `foreach(target ${RCCL_TEST_EXECUTABLES})` ?

atulkulk:
> CudaPtrCheck TEST_F(ArgCheckTest, CudaPtrCheck_InvalidScenarios) { This test actually has three tests in it. In general, it is better to avoid multiple independent asserts/tests in one test. The reason is if one assert fails, then the next asserts that are not dependent do not execute. It is OK to have dependent asserts in one test, for example, two asserts that check for two separate values from a single call to a function. It would be better in our case to separate the single test into three tests, one for each of the cases mentioned.

Addressed!

> The three tests embedded in this one test are pretty good. Do we need one extra test for the type being different from the cudaMemoryTypeDevice?

Addressed!

> PtrCheck The tests are adequate for this function.
> 
> CommCheck It would be good to split this test into two tests since the condition in the function is a or (||) and not an and (&&). The first test should corrupt startMagic and the second test should corrupt the endMagic.

Added separate tests!

> ArgCheck TEST_F(ArgCheckTest, ArgsCheck_InvalidInputs) { We would want to separate this test into multiple tests for the same reason as mentioned in the top, to avoid multiple independent asserts in one test.

Addressed!

> Test for invalid root should include two tests, one for value < 0 and the other where the value is info->comm->nRanks?

Addressed!

> Likewise for two different range values for info->datatype?

Addressed!

> It will be useful to revisit the different ranges for each condition and have two tests for both the bounds.

Addressed!

> The remaining tests are good, but need to be split into multiple tests to avoid independent asserts in one test.
> 
> Do we need tests for the public API functions in alt_rsmi.cc?

Public API functions are covered by rccl-UnitTests and rccl-tests.

rahulvaidya20 (approved):

venksubramd:
The changes look pretty good and ready to close if the tests are passing on the CI.

BertanDogancay (approved):

PR #1676, opened 2025-05-05
Code coverage improvements for alloc.h 

awelling2801:
**Work item:** Internal

**What were the changes?**  
Added Unit tests to increase code coverage for alloc.h
This PR adds one file - test/AllocTests.cpp comprising of 9 tests. 

The test improved code coverage for alloc.h 
|| Function Coverage  | Line Coverage | Region coverage | Branch Coverage |
|-------------| ------------- | ------------- | ------------- | ------------- |
| Default Coverage  |  57.89% (11/19)  |   67.19% (129/192)  |   56.78% (201/354) |   43.48% (60/138)  |
| Updated Coverage |  100.00% (19/19) |   94.79% (182/192) |   72.91% (261/358) |   52.90% (73/138)  |

The outcome was achieved by adding a AllocTests.cpp file comprising of the tests. 
The visibility set to default instead of hidden achieved by   https://github.com/ROCm/rccl/pull/1664

	`/opt/rocm/lib/llvm/bin/llvm-cov show ./rccl-UnitTests --instr-profile=merged.profdata --format=html --output-dir=$WORKDIR/output/rccl_cov_report-lib --project-title=RCCL_Lib_Coverage_Report --object ../librccl.so`

## Approval Checklist

### DIFF
--- a/test/AllocTests.cpp
+++ b/test/AllocTests.cpp
@@ -118,5 +118,32 @@ namespace RcclUnitTesting
         hipFree(d_dst);
     
     }
-    
+
+    TEST(Alloc, ZeroElementMemcpy) {
+        float *d_src = nullptr, *d_dst = nullptr;
+        ASSERT_EQ(hipMalloc(&d_src, sizeof(float)), hipSuccess);
+        ASSERT_EQ(hipMalloc(&d_dst, sizeof(float)), hipSuccess);
+
+        ncclResult_t result = ncclCudaMemcpy<float>(d_dst, d_src, 0);
+        EXPECT_EQ(result, ncclSuccess) << "Zero-element copy should succeed (no-op)";
+
+        hipFree(d_src);
+        hipFree(d_dst);
+    }
+
+    TEST(Alloc, MemcpyNullSrcOrDstPointer) {
+        constexpr size_t N = 16;
+        float* d_valid = nullptr;
+        ASSERT_EQ(hipMalloc(&d_valid, N * sizeof(float)), hipSuccess);
+
+        // Case 1: src is nullptr
+        ncclResult_t result = ncclCudaMemcpy<float>(d_valid, nullptr, N);
+        EXPECT_EQ(result, ncclUnhandledCudaError) << "Expected ncclUnhandledCudaError when src is nullptr";
+
+        // Case 2: dst is nullptr
+        result = ncclCudaMemcpy<float>(nullptr, d_valid, N);
+        EXPECT_EQ(result, ncclUnhandledCudaError) << "Expected ncclUnhandledCudaError when dst is nullptr";
+
+        hipFree(d_valid);
+    }
 } //namespace rccl
--- a/test/AllocTests.cpp
+++ b/test/AllocTests.cpp
@@ -1,0 +1,122 @@
+/*************************************************************************
+ * Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+#include <alloc.h>
+ 
+#include "TestBed.hpp"
+
+template ncclResult_t ncclCudaMemcpy<float>(float*, float*, size_t);
+namespace RcclUnitTesting
+{
+    TEST(Alloc, ncclIbMallocDebugNonZero) {
+        void* ptr = nullptr;
+        size_t size = 4096;
+      
+        ncclResult_t result = ncclIbMalloc(&ptr, size);
+      
+        EXPECT_EQ(result, ncclSuccess);
+        ASSERT_NE(ptr, nullptr);
+      
+        char* char_ptr = static_cast<char*>(ptr);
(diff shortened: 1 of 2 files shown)


Nikhil-Nunna:
There is a failure in the Precheckin pipeline, looks like missing symbols

```2 warnings generated when compiling for host.
[100%] Linking CXX executable rccl-UnitTests
ld.lld: error: undefined symbol: ncclDebugNoWarn
>>> referenced by AllocTests.cpp
>>>               CMakeFiles/rccl-UnitTests.dir/AllocTests.cpp.o:(ncclResult_t ncclCudaMemcpy<float>(float*, float*, unsigned long))
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [test/CMakeFiles/rccl-UnitTests.dir/build.make:488: test/rccl-UnitTests] Error 1
make[1]: *** [CMakeFiles/Makefile2:153: test/CMakeFiles/rccl-UnitTests.dir/all] Error 2
make: *** [Makefile:156: all] Error 2

**Comment by venksubramd** [comment]
The test ncclIbMallocDebugZeroSize is good, it checks for the behavior correctly. The test ncclIbMallocDebugNonZero checks for the happy path. Do we need tests for the situations when the function ncclIbMalloc (that is ncclIbMallocDebug) would return ncclSystemError?

The test for ncclCuMemHostAlloc checks what happens if the input values are NULL (which is good), but we need test that checks what the actual behavior of the code is when all parameters are value. Also, for the negative test for NULL values, we should have multiple tests for different arguments being NULL instead of grouping all NULL arguments into one test.

Rest of the tests in this file, except the last one, check for error condition that results from invalid argument. We need tests that verify the expected behavior of these functions.

The file alloc.h has about two dozen functions, we should revisit and see which of those functions need test since right now we have tests only for a handfull of functions in alloc.h.

**Comment by awelling2801** [comment]
@venksubramd 
For tests concerning `ncclIbMalloc`, I could have a test where `ncclIbMalloc` returns `ncclSystemError`. This would require mocking `posix_memalign` which is tricky (not impossible) to do. Looking into this. 

`ncclCuMemHostAlloc` has two implementations covered by `#if CUDART_VERSION >= 12020`. For RCCL - any argument would return a `ncclInternalError` since it is just not supported. I do not think its worth it to explore this further unless this (and other functions which are still unsupported) have rccl support. 

> Rest of the tests in this file, except the last one, check for error condition that results from invalid argument. We need tests that verify the expected behavior of these functions

As mentioned, these are stubbed functions and the expected output for all of these functions is `ncclInternalError` 
 

> The file alloc.h has about two dozen functions, we should revisit and see which of those functions need test since right now we have tests only for a handfull of functions in alloc.h.

Function Coverage | Line Coverage | Region coverage | Branch Coverage
-- | -- | -- | --
Default Coverage | 57.89% (11/19) | 67.19% (129/192) | 56.78% (201/354) | 43.48% (60/138)
Updated Coverage | 100.00% (19/19) | 94.79% (182/192) | 72.91% (261/358) | 52.90% (73/138)

This coverage effectively validates the current implementation, and I’d consider the test a success.

**Comment by venksubramd** [comment]
@awelling2801 

Since the mocking efforts may take some time, it may be better to open a new ticket for that task. That way we can move forward to close this.

Thanks for your comments on parts that may not be relevant to rccl.

I am OK to close this if we open a new ticket for the mocking efforts.

File: test/AllocTests.cpp [hunk 1]
```diff
@@ -0,0 +1,122 @@
...
+    }
+       
+    TEST(Alloc, ncclCuMemFreeAddr)
+    {
+        void* testPtr = reinterpret_cast<void*>(0xbeefcafe); // Arbitrary non-null pointer
+        ncclResult_t result = ncclCuMemFreeAddr(testPtr);
+        ASSERT_EQ(result, ncclInternalError);   
+    }
+    
+    TEST(Alloc, NcclCudaMemcpy) {
```
rahulvaidya20, line 90: It will be good if we can add more tests here to cover corner cases like `nullptr` handling, `0-byte copy`, etc.

rahulvaidya20 (commented):
We need to revisit `cuMem` related tests once we have the support in ROCm 7.0.

test/AllocTests.cpp
awelling2801: Added

atulkulk:
> @venksubramd For tests concerning `ncclIbMalloc`, I could have a test where `ncclIbMalloc` returns `ncclSystemError`. This would require mocking `posix_memalign` which is tricky (not impossible) to do. Looking into this.
> 
> `ncclCuMemHostAlloc` has two implementations covered by `#if CUDART_VERSION >= 12020`. For RCCL - any argument would return a `ncclInternalError` since it is just not supported. I do not think its worth it to explore this further unless this (and other functions which are still unsupported) have rccl support.

@venksubramd  @awelling2801  Created a new ticket to look into adding new tests using GMock: [ GPUAI-5985 ]

rahulvaidya20 (approved):

atulkulk (approved):

venksubramd:
@awelling2801 @atulkulk  Thanks for creating the new tickets. We can close this one.

PR #1680, opened 2025-05-07
Write JSON file to /tmp directory to avoid incorrect write access in …

mberenjk:
…recorderTest

## Details

SWDEV-531185

**What were the changes?**  
Change output JSON file path to /tmp to avoid incorrect access errors.

The test fails when it runs from /opt/rocm/bin directory

**Additional Documentation:**  

### DIFF
--- a/test/_RecorderTests.cpp
+++ b/test/_RecorderTests.cpp
@@ -25,7 +25,7 @@ namespace RcclUnitTesting
    * ******************************************************************************************/
   TEST(Recorder, ParseJson)
   {
-    setenv("RCCL_REPLAY_FILE", "test.json", 1);
+    setenv("RCCL_REPLAY_FILE", "/tmp/test.json", 1);
 
     int pid = getpid();
     hipStream_t stream;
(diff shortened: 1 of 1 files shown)


AtlantaPepsi (approved):

test/_RecorderTests.cpp
```diff
@@ -25,7 +25,7 @@ namespace RcclUnitTesting
    * ******************************************************************************************/
   TEST(Recorder, ParseJson)
   {
-    setenv("RCCL_REPLAY_FILE", "test.json", 1);
+    setenv("RCCL_REPLAY_FILE", "/tmp/test.json", 1);
```
corey-derochie-amd, line 28: Because the `Recorder` class only reads the `RCCL_REPLAY_FILE` env var in the constructor, this test could fail if other Recorder tests run before it. In other words, this is not an independent test.

The only solution would be to slightly change how `Recorder` works, though.

corey-derochie-amd (dismissed):

test/_RecorderTests.cpp
corey-derochie-amd: Question is, are we okay with this for now?

corey-derochie-amd, line 28: Should we use `mktemp`, or is it fine to just hard-code this temp path?

thananon (approved):

test/_RecorderTests.cpp
corey-derochie-amd: We will deal with this later.

PR #1682, opened 2025-05-08
RCCL Multinode DMA Buffer crash fix

PJAvinash:
## Details

**Work item:** _"Internal"

**What were the changes?**  
wrap_ibv_reg_dmabuf_mr is used instead of wrap_ibv_reg_mr when applicable

To address Multi-node RCCL fail when using DMABuf with RCCL_GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING=1

**Additional Documentation:**  
a new field is added in struct ncclIbGpuFlush to handle file descriptor opening / closing . This will have effect on size of ncclIbRecvCommDev

## Approval Checklist
  - there are any NCCL API version changes NO
  - any changes impact library users, and/or NO
  - any changes impact any other ROCm library. NO

### DIFF
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <new>
 #include <type_traits>
+
 int ncclCudaCompCap();
 
 // PCI Bus ID <-> int64 conversion functions
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -17,7 +17,7 @@
 #include <algorithm>
 #include <new>
 #include <type_traits>
-
+#include <unistd.h>
 int ncclCudaCompCap();
 
 // PCI Bus ID <-> int64 conversion functions
@@ -545,4 +545,25 @@ T* ncclIntruQueueMpscAbandon(ncclIntruQueueMpsc<T,next>* me) {
     return head;
   }
 }
+/**
+ * @brief function to get page size of the system
+ */
+size_t get_sc_page_size(void);
+
+/**
+ * @brief function to get system's page size aligned memory address and buffersize 
+ *
+ * Given a pointer `ptr` to a buffer of size `bufsize`, this function computes:
+ *   1. A new pointer `aligned_ptr` which points to the start of the page-aligned memory region that includes `ptr`.
+ *   2. A new size `aligned_size` that is the minimum number of bytes (aligned to page size) needed to cover the original buffer from `aligned_ptr`.
+ *
+ * This is useful, for example, when performing operations such as memory mapping or advising memory usage (e.g., with `mmap`, `madvise`, `mlock`, etc.), which often require page-aligned memory ranges.
+ * This function doesn't dereferece the input pointer
+ *
+ * @param[in]  ptr           Pointer to the start of the original memory buffer.
+ * @param[in]  bufsize       Size (in bytes) of the original buffer starting at `ptr`.
+ * @param[out] aligned_ptr   Pointer to a variable that will be set to the aligned base address.
+ * @param[out] aligned_size  Pointer to a variable that will be set to the aligned size.
+ */
+void get_aligned_ptr_and_size(const void *ptr, const size_t bufsize, void **aligned_ptr, size_t *aligned_size);
 #endif
--- a/src/include/utils.h
+++ b/src/include/utils.h
@@ -17,7 +17,6 @@
 #include <algorithm>
 #include <new>
 #include <type_traits>
-#include <unistd.h>
 int ncclCudaCompCap();
 
 // PCI Bus ID <-> int64 conversion functions
--- a/src/misc/rocmwrap.cc
+++ b/src/misc/rocmwrap.cc
@@ -100,12 +100,7 @@ static void initOnceFunc() {
     //goto error;
   //}
 
-  /* DMA-BUF support */
-  //ROCm support
-  if (ncclParamDmaBufEnable() == 0 ) {
-    INFO(NCCL_INIT, "Dmabuf feature disabled without NCCL_DMABUF_ENABLE=1");
-    goto error;
-  }
+  
   res = pfn_hsa_system_get_info((hsa_system_info_t) 0x204, &dmaBufSupport);
   if (res != HSA_STATUS_SUCCESS || !dmaBufSupport) {
     INFO(NCCL_INIT, "Current version of ROCm does not support dmabuf feature.");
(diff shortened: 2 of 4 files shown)


src/misc/utils.cc
```diff
@@ -289,3 +289,23 @@ void ncclMemoryStackDestruct(struct ncclMemoryStack* me) {
...
+  static size_t cached_page_size = 0;
+  size_t ps = __atomic_load_n(&cached_page_size,__ATOMIC_RELAXED);
+  if (ps == 0) {
+      ps = (size_t)sysconf(_SC_PAGESIZE);
+      __atomic_store_n(&cached_page_size, ps,__ATOMIC_RELAXED);
+  }
+  return ps;
+}
+
+void get_aligned_ptr_and_size(const void *ptr, const size_t bufsize, void **aligned_ptr, size_t *aligned_size)
```
mustafabar: In include/utils.h `ncclMemoryStackAlloc` seems to achieve a similar purpose. And based on the comment on `ncclIbMallocDebug` in alloc.h callable in ncclIbMalloc, I think it does the same thing and is tested in production

src/transport/net_ib.cc
```diff
@@ -1706,10 +1707,20 @@ ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle
...
+        uint64_t export_offset = 0;
+        void *aligned_ptr = NULL;
+        size_t aligned_size = 0;
+        get_aligned_ptr_and_size(rCommDev->gpuFlush.gpuFlushGpuMem, sizeof(int) /*devicebuffersize*/, &aligned_ptr, &aligned_size);
+        hsa_status_t export_status = pfn_hsa_amd_portable_export_dmabuf(aligned_ptr, aligned_size, &rCommDev->gpuFlush.dmabuf_fd, &export_offset);
+        if(rCommDev->gpuFlush.dmabuf_fd < 0 || export_status != HSA_STATUS_SUCCESS){
+          WARN("Failed to export DMA BUF");
+          goto fail;
+        }
+        NCCLCHECKGOTO(wrap_ibv_reg_dmabuf_mr(&rCommDev->gpuFlush.gpuMr,rCommDev->base.pd,export_offset, sizeof(int),(uint64_t)rCommDev->gpuFlush.gpuFlushGpuMem /*iova*/,rCommDev->gpuFlush.dmabuf_fd, IBV_ACCESS_LOCAL_WRITE|IBV_ACCESS_REMOTE_WRITE|IBV_ACCESS_REMOTE_READ), ret, fail);
```
thananon: This seems to handle dmabuf mode but we are removing the original behavior here (ibv_reg_mr). If the user is not using dmabuf, will this work?

src/misc/rocmwrap.cc
```diff
@@ -100,12 +100,7 @@ static void initOnceFunc() {
...
   //}
 
-  /* DMA-BUF support */
-  //ROCm support
-  if (ncclParamDmaBufEnable() == 0 ) {
-    INFO(NCCL_INIT, "Dmabuf feature disabled without NCCL_DMABUF_ENABLE=1");
-    goto error;
-  }
+  
   res = pfn_hsa_system_get_info((hsa_system_info_t) 0x204, &dmaBufSupport);
```
nileshnegi: why do you want to run these tests if DMABuf is not enabled?

src/transport/net_ib.cc
```diff
@@ -1687,7 +1689,7 @@ ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle
 
   rComm->flushEnabled = ((ncclIbGdrSupport() == ncclSuccess || ncclIbDmaBufSupport(lComm->dev) == ncclSuccess)
                             && (ncclParamIbGdrFlushDisable() == 0)) ? 1 : 0;
-
+  useDMABuff  = (ncclIbDmaBufSupport(lComm->dev) == ncclSuccess);
```
thananon: nitpick: I would move this line above and replace the second condition with `useDMABuff` to prevent double calling.

Also `usdDMABuff` capitalization is introducing another styling of dmabuf in our code base. We should at least reuse the styling that exist. I recommend `useDmaBuf` to match one in `transport/net.cc`

src/misc/rocmwrap.cc
```diff
@@ -155,14 +150,17 @@ static void initOnceFunc() {
...
    * Multiple calls of hsa_init() will return immediately
    * without making any relevant change
    */
   pfn_hsa_init();
-
+  /* DMA-BUF support */
+  //ROCm support
+  if (ncclParamDmaBufEnable() == 0 ) {
+    WARN("Dmabuf feature disabled without NCCL_DMABUF_ENABLE=1");
+  }
```
thananon: Revert this logic to original. The original logic would be if the user does not set this flag, we dont bother checking hsa functions. It was fine as is.

src/misc/utils.cc
thananon:
```suggestion
void get_aligned_ptr_and_size(const void *ptr, const size_t bufsize, void **aligned_ptr, size_t *aligned_size) {
```

src/transport/net_ib.cc
```diff
@@ -1687,7 +1690,7 @@ ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle
 
   rComm->flushEnabled = ((ncclIbGdrSupport() == ncclSuccess || ncclIbDmaBufSupport(lComm->dev) == ncclSuccess)
```
thananon: I would move the useDmaBuf assignment up and reuse it.

```suggestion
  useDmaBuf = (ncclIbDmaBufSupport(lComm->dev) == ncclSuccess);
  rComm->flushEnabled = (ncclIbGdrSupport() == ncclSuccess || useDmaBuf );
```

src/include/utils.h
```diff
@@ -17,7 +17,6 @@
 #include <algorithm>
 #include <new>
 #include <type_traits>
-
```
thananon: line removal.

thananon (approved):
Functionally ok. Commented aesthetic issues.
I tested this on non DMAbuf machine.

mustafabar (changes requested):
Please review if the newly added logic for aligned allocation is necessary based on the comments above

thananon:
@mustafabar I reviewed the logic and @PJAvinash pointed out the needs for this function as the existing ones are allocating aligned buffer while this flow does not need allocation. Alternatively we can move that function logic into the place we use it (net_ib.cc) but the calculation for page alignment will be the same.

mustafabar (approved):
Approving for staging

thananon:
Merged without gfx942 CI result as it is known to fail from different issue. This PR unblocks staging and expected to get gfx942 tests in staging.

PR #1689, opened 2025-05-12
Code coverage improvements - src/transport/shm.cc

atulkulk:
Work item: Internal

What were the changes?
Added new unit tests to cover test cases that were not covered by the functional tests for src/transport/shm.cc file.

Why were the changes made?

How was the outcome achieved?
New unit tests were added.

| **shm.cc** | Initial | 78.26% (18/23)    | 48.29% (198/410) | 35.84% (286/798) | 20.77% (81/390) |
|                 | Updated | 100.00% (23/23)    | 93.90% (385/410) | 69.55% (555/798) | 49.99% (193/390) |

In addition to default execution of the unit test binary, following execution commands needs to be executed to cover certain portion of the code:

```
NCCL_SHM_USE_CUDA_MEMCPY=1 NCCL_SHM_MEMCPY_MODE=1 ./rccl-UnitTestsFixtures --gtest_filter="ShmProxyTest.ShmSendRecvProxyFree"
NCCL_SHM_USE_CUDA_MEMCPY=1 NCCL_SHM_MEMCPY_MODE=2 ./rccl-UnitTestsFixtures --gtest_filter="ShmProxyTest.ShmSendRecvProxyFree"
NCCL_SHM_USE_CUDA_MEMCPY=1 NCCL_SHM_LOCALITY=3 ./rccl-UnitTestsFixtures --gtest_filter="ShmProxyTest.ShmSendRecvProxyFree"
```

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -31,6 +31,40 @@ if(BUILD_TESTS)
 
   include_directories(${GTEST_INCLUDE_DIRS} ./common)
 
+    # Common include directories
+  set(RCCL_COMMON_INCLUDE_DIRS
+    ${GTEST_INCLUDE_DIRS}
+    ${PROJECT_BINARY_DIR}/include # for generated rccl.h header
+    ${PROJECT_BINARY_DIR}/hipify/src/include  # for rccl_bfloat16.h
+    ${PROJECT_BINARY_DIR}/hipify/gensrc # for rccl_bfloat16.h
+    ${PROJECT_BINARY_DIR}/hipify/src # for graph/topo.h
+    ${PROJECT_BINARY_DIR}/hipify/src/include/plugin # for recorder tests, nccl_tuner.h
+    ${ROCM_PATH}/include
+    ${ROCM_PATH}
+  )
+
+  # Common compile definitions
+  set(RCCL_COMMON_COMPILE_DEFS ROCM_PATH="${ROCM_PATH}")
+  if(LL128_ENABLED)
+    list(APPEND RCCL_COMMON_COMPILE_DEFS ENABLE_LL128)
+  endif()
+  if(OPENMP_TESTS_ENABLED)
+    list(APPEND RCCL_COMMON_COMPILE_DEFS ENABLE_OPENMP)
+  endif()
+  list(APPEND RCCL_COMMON_COMPILE_DEFS __HIP_PLATFORM_AMD__)
+
+  # Common link libraries
+  set(RCCL_COMMON_LINK_LIBS
+    ${GTEST_BOTH_LIBRARIES}
+    hip::host hip::device hsa-runtime64::hsa-runtime64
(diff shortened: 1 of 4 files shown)


venksubramd:
Looking at SchTests.cpp I have a feeling we are reaching to test functions that are not visible from the outside. This is evident form the effort we put in to redefine so many structs within the test, thus violating the DRY principle. This makes the tests brittle if and when the structures change. It also increases the cost of writing these tests and maintaining them.

sch.cc has two kinds of functions. Those that are internal to this file and those that are implemented from the declaration in sch.h. We should focus on testing only the functions that are visibile via sch.h. Any function that is internal to sch.cc should be tested only via the public visible interface for those functions, and thus we should not be directly working with, in the tests, any struct that is internal to sch.cc.

thananon (approved):

test/CMakeLists.txt
```diff
@@ -31,6 +31,40 @@ if(BUILD_TESTS)
...
+  endif()
+  list(APPEND RCCL_COMMON_COMPILE_DEFS __HIP_PLATFORM_AMD__)
+
+  # Common link libraries
+  set(RCCL_COMMON_LINK_LIBS
+    ${GTEST_BOTH_LIBRARIES}
+    hip::host hip::device hsa-runtime64::hsa-runtime64
+    Threads::Threads
+    dl
+    fmt::fmt
```
rahulvaidya20: Can you link only the headers `fmt::fmt-header-only`?

atulkulk: done.

rahulvaidya20 (approved):

venksubramd:
if (result != ncclSuccess) {
      fprintf(stderr, "Failed to allocate shareable buffer: %d\n", result);
      return;
    }

In addition to fprintf, should we force fail this test to indicate the failure?

TEST_F(ShmTests, ShmImportShareableBufferNullPointerChecks) {
It is convenient to write multiple tests in one test, but it is better to avoid that. Failure of one test should not prevent running the other tests that are independent.

As a separate task we need to revisit the duplication of structures from the code into the test file and see if there are alternate ways to avoid this.

In general the tests in this file (except for the comments above) look good. We can move towards closing this.

atulkulk:
> ```
> if (result != ncclSuccess) {
>   fprintf(stderr, "Failed to allocate shareable buffer: %d\n", result);
>   return;
> }
> ```
> 
> In addition to fprintf, should we force fail this test to indicate the failure?
> 
> TEST_F(ShmTests, ShmImportShareableBufferNullPointerChecks) { It is convenient to write multiple tests in one test, but it is better to avoid that. Failure of one test should not prevent running the other tests that are independent.
> 
> As a separate task we need to revisit the duplication of structures from the code into the test file and see if there are alternate ways to avoid this.
> 
> In general the tests in this file (except for the comments above) look good. We can move towards closing this.

Comments have been addressed!

PR #1690, opened 2025-05-12
Added tests for Ipcsocket

awelling2801:
**Work item:** _"Internal"

**What were the changes?**  
Added Unit tests to increase code coverage for ipcsocket.cc
This PR adds one file - test/IpcsocketTests.cpp comprising of 5 tests.

The test improved code coverage for ipcsocket.cc
|--|-- | -- | -- | --
Default Coverage | 42.86% (3/7) |   24.68% (39/158) |   22.36% (53/237) |   18.29% (15/82)
Updated Coverage |  100.00% (7/7) |   77.71% (122/157) |   58.09% (79/136) |   35.00% (28/80)

The outcome was achieved by adding a IpcsocketTests.cpp file comprising of the tests.
The environment variables used were as follows - 

`NCCL_DEBUG=TRACE NCCL_DEBUG_SUBSYS=ALL RCCL_ENABLE_INTRANET=1 NCCL_DMABUF_ENABLE=1 RCCL_GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING=0 ./rccl-UnitTests --gtest_filter=Ipcsocket.SendAndReceiveFd:Ipcsocket.SocketInitNullHandle:Ipcsocket.SocketGetFdNullHandle:Ipcsocket.SocketCloseNullHandle:Ipcsocket.SocketCloseNegativeHandle `

## Approval Checklist

### DIFF
--- a/test/IpcsocketTests.cpp
+++ b/test/IpcsocketTests.cpp
@@ -1,0 +1,157 @@
+/*************************************************************************
+ * Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+#include <sys/wait.h>
+#include<proxy.h> 
+#include <comm.h>
+#include <ipcsocket.h>
+
+#include "TestBed.hpp"
+
+namespace RcclUnitTesting
+{
+  TEST(Ipcsocket, SocketInitNullHandle){
+    int rank = 0;
+    uint64_t hash = 0x1234;
+    volatile uint32_t abortFlag = 0;
+    // Call function with NULL handle
+    ncclResult_t result = ncclIpcSocketInit(NULL, rank, hash, &abortFlag);
+    // Check that it returns the correct error code
+    ASSERT_EQ(result, ncclInternalError);
+  }
+
+  TEST(Ipcsocket, SocketGetFdNullHandle){
+    int fd = -1;
+    ncclResult_t result = ncclIpcSocketGetFd(nullptr, &fd);
+    EXPECT_EQ(result, ncclInvalidArgument);
+  }
+
+  TEST(Ipcsocket, SocketCloseNullHandle){
+    ncclResult_t result = ncclIpcSocketClose(nullptr);
+    EXPECT_EQ(result, ncclInternalError);
+  }
+
+  TEST(Ipcsocket, SocketCloseNegativeHandle){
+    ncclIpcSocket handle = {};
+    handle.fd = -1;
+    ncclResult_t result = ncclIpcSocketClose(&handle);
+    EXPECT_EQ(result, ncclSuccess);
+  }
+
+  TEST(Ipcsocket, SendAndReceiveFd) {
+    int pipeFd[2]; // for sync from child -> parent
+    ASSERT_EQ(pipe(pipeFd), 0);
+  
+    pid_t pid = fork();
+    ASSERT_NE(pid, -1);
+  
+    const int rank = 1;
+    const uint64_t hash = 0x12345678;
+    volatile uint32_t abortFlag = 0;
+  
+    if (pid == 0) {
+      // === Child: Receiver ===
+      close(pipeFd[0]);
+  
+      char sockPath[108];
+      snprintf(sockPath, sizeof(sockPath), "/tmp/ipc_sock_%lx", hash);
+      unlink(sockPath);
+      
+      int listenFd = socket(AF_UNIX, SOCK_SEQPACKET, 0);
+      ASSERT_GT(listenFd, 0);
+      
+      struct sockaddr_un addr = {};
+      addr.sun_family = AF_UNIX;
+      strncpy(addr.sun_path, sockPath, sizeof(addr.sun_path) - 1);
+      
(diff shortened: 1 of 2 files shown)


test/IpcsocketTests.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "TestBed.hpp"
+
+namespace RcclUnitTesting
+{
+  TEST(Ipcsocket, SocketInitNullHandle){
+    int rank = 0;
+    uint64_t hash = 0x1234;
+    volatile uint32_t abortFlag = 0;
+    // Call function with NULL handle
+    ncclResult_t result = ncclIpcSocketInit(NULL, rank, hash, &abortFlag);
```
corey-derochie-amd, line 22: These symbols are not exported from librccl, so compilation fails.

corey-derochie-amd (dismissed):

test/IpcsocketTests.cpp
corey-derochie-amd: It may be necessary to link ipcsocket.cc directly into rccl-UnitTests to test this unit.

awelling2801: This commit follows - https://github.com/ROCm/rccl/pull/1664. Which sets the visibility to default instead of hidden to expose internal library symbols for unit testing the RCCL library. 
If you are testing this for now - I would suggest changing the visibility to default in CMakeLists.txt

venksubramd:
The test for ncclIpcSocketInit is checking if the returned error code for handle being NULL is correct. This tests the initial check done by the ncclIpcSocketInit function. However, we do not have any tests for what the ncclIpcSocketInit actually is supposed to do. What is the intent of ncclIpcSocketInit, what is it supposed to achieve, and what are the possible failures it may encounter beyond the handle being NULL? We would need a lot more tests for that, beyond the one test we currently have.


The test for ncclIpcSocketGetFd is checking if the returned error code for handle being NULL is correct. We would use at least two more tests to check what the function should do if the handle is not NULL but for different values for the fd argument.

We have two tests for ncclIpcSocketClose, the first checks the behavior if the handle is NULL. The second test checks the behavior if the handle's fd is less than zero. We could use a few more tests for the two steps after the initial few checks in this function. 

We have one test for the ncclIpcSocketRecvFd function. This function, however, has only one line where it calls the ncclIpcSocketRecvMsg. It may be better to write tests for the ncclIpcSocketRecvMsg function instead of ncclIpcSocketRecvFd or may be a very small test for ncclIpcSocketRecvFd that merely checks that this function calls ncclIpcSocketRecvMsg with proper arguments, by mocking ncclIpcSocketRecvMsg. This one test is quite long and has a if-else part. Generally we should avoid branches in unit tests as we have no control, in that case, over running all paths in a single run of the test.

We don't have any tests for ncclIpcSocketSendFd and ncclIpcSocketSendMsg?

awelling2801:
>  However, we do not have any tests for what the ncclIpcSocketInit actually is supposed to do. What is the intent of ncclIpcSocketInit, what is it supposed to achieve, and what are the possible failures it may encounter beyond the handle being NULL?

These scenarios are already exercised by other test cases. It would be helpful to review the existing code coverage first to avoid redundant test efforts.

There cases that need mocking for system level functions need their own PR.

>  We would use at least two more tests to check what the function should do if the handle is not NULL but for different values for the fd argument. 

Already covered by other test cases and do not need their own tests 

> We have two tests for ncclIpcSocketClose, the first checks the behavior if the handle is NULL. The second test checks the behavior if the handle's fd is less than zero. We could use a few more tests for the two steps after the initial few checks in this function.

The case for - 
`#ifndef USE_ABSTRACT_SOCKET
  if (handle->socketName[0] != '\0') {
    unlink(handle->socketName);
  }
#endif` 
The unlink() block is conditionally compiled only when USE_ABSTRACT_SOCKET is not defined. In our current build setup, USE_ABSTRACT_SOCKET is defined, so this section is omitted entirely during compilation. Hence we do not need tests for it.
The rest of the path is already covered by other tests and we do not need redundancy. 

> It may be better to write tests for the ncclIpcSocketRecvMsg function instead of ncclIpcSocketRecvFd or may be a very small test for ncclIpcSocketRecvFd that merely checks that this function calls ncclIpcSocketRecvMsg with proper arguments, by mocking ncclIpcSocketRecvMsg. 

Not sure why this is necessary. The ncclIpcSocketRecvFd does nothing except call ncclIpcSocketRecvMsg. 

> We don't have any tests for ncclIpcSocketSendFd and ncclIpcSocketSendMsg?

Tests for both send and receive. Not really possible to do one without the other and this was after discussion with the team

venksubramd:
@awelling2801 

Could you please create a new PR for the tasks that will require mocking so we can focus on that effort separately. For the parts that are tested using other tests, could you please check if those tests actually examine the behavior (which is necessary) or there merely focus on code coverage (which is not sufficient by itself).

atulkulk:
> @awelling2801
> 
> Could you please create a new PR for the tasks that will require mocking so we can focus on that effort separately. For the parts that are tested using other tests, could you please check if those tests actually examine the behavior (which is necessary) or there merely focus on code coverage (which is not sufficient by itself).

Created ticket for additional work: [ GPUAI-5988 ]

venksubramd:
@awelling2801 @atulkulk  Thank you for creating ticket for the additional work. We can close this one.

atulkulk (approved):

corey-derochie-amd (approved):

PR #1694, opened 2025-05-14
Created coverage tests for rccl_wrap

awelling2801:
**Work item:** _"Internal"

**What were the changes?**  
Added Unit tests to increase code coverage for rccl_wrap.cc
This PR adds one file - test/RcclWrapTests.cpp comprising of 6 tests.

The test improved code coverage for rccl_wrap.cc
|--|-- | -- | -- | --
Default Coverage | 33.33% (1/3)  |     14.81% (8/54)  |     12.05% (10/83)  |   17.50% (7/40)
Updated Coverage |  100.00% (3/3)  |   100.00% (54/54) |    93.94% (62/66) |     62.50% (25/40)

The outcome was achieved by adding a RcclWrapTests.cpp file comprising of the tests.
A compiler macro was needed to run these tests - 
`-DRCCL_EXPOSE_STATIC=ON`

**Additional Documentation:**  
This PR also fixes a bug to use RCCL_EXPOSE_STATIC and use rccl_static and rccl_static_inline correctly.

## Approval Checklist

### DIFF
--- a/test/RcclWrapTests.cpp
+++ b/test/RcclWrapTests.cpp
@@ -5,7 +5,7 @@
  ************************************************************************/
 
 #include <gtest/gtest.h>
-#include <rccl/rccl.h> 
+#include <rccl/rccl.h>
 #include <comm.h>
 #include <graph/topo.h>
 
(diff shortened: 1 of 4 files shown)


test/RcclWrapTests.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <graph/topo.h>
+
+#include "TestBed.hpp"
+
+
+namespace RcclUnitTesting
+{
+  TEST(Rcclwrap, RcclGetAlgoInfoTest) {
+    #ifndef RCCL_EXPOSE_STATIC
+    #error "RCCL_EXPOSE_STATIC is not defined at compile time"
```
BertanDogancay: Looks like the CI is failing due to `RCCL_EXPOSE_STATIC` not being defined which is the default case. Is this intentional? Why don't we just skip the test with a warning message instead if this variable is not defined at compile time?

venksubramd:
Due to the simplicity of the rcclGetAlgoInfo function we will need only one test for this function. The current test, however, only checks for the return value. The test needs to be extended with a few additional asserts [in this case since they are all related to one call, it is a good place to have multiple (non-independent/related) asserts]. The additional asserts should verify the value for algo, protocol, and maxChannels.

The test for rcclFuncMaxSendRecvCount needs one more assert to verify that the value of maxCount has been set properly. This function is also simple and one test is adequate.

There are multiple tests for rcclUpdateCollectiveProtocol. Most of these tests are right on target. The first test, however, is checking if the value of info.protocol is one of two values. Would it be better to have two tests where the first checks for the value being NCCL_PROTO_LL128 and the second checks for the value being NCCL_PROTO_LL? That way we can be sure which situation leads to which of these values being set.

Do we need test for the rcclUpdateThreadThreshold function?

awelling2801:
@venksubramd 
You are correct about rcclGetAlgoInfo and rcclFuncMaxSendRecvCount. 
Added an assert for rcclFuncMaxSendRecvCount but I will have another PR for rcclGetAlgoInfo. 

NCCL_PROTO_LL128 and NCCL_PROTO_LL are both compiler options. one of them can be used at a time while compiling. Hence the decision. 

 rcclUpdateThreadThreshold is a newer function and can have another PR as well.

rahulvaidya20 (approved):

atulkulk (approved):

PR #1700, opened 2025-05-20
Created coverage tests for coll_reg.cc

awelling2801:
**Work item:** _"Internal"

**What were the changes?**  
Added Unit tests to increase code coverage for coll_reg.cc
This PR adds one file - test/CollRegTests.cpp comprising of 6 tests.

The test improved code coverage for coll_reg.cc

The outcome was achieved by adding a RcclWrapTests.cpp file comprising of the tests.
A compiler macro was needed to run these tests -
-DRCCL_EXPOSE_STATIC=ON

## Approval Checklist

### DIFF
--- a/test/CollRegTests.cpp
+++ b/test/CollRegTests.cpp
@@ -1,0 +1,102 @@
+/*************************************************************************
+ * Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+
+#include "TestBed.hpp"
+
+#include "CollRegUtils.hpp"
+
+namespace RcclUnitTesting
+{
+  
+TEST(CollReg, RegisterCheckP2P_ConnectedWithP2PFlag) {
+  ncclComm comm = {};
+  ncclConnector conn = {};
+  conn.connected = 1;
+  conn.conn.flags = NCCL_P2P_WRITE;
+
+  bool needReg = false;
+  EXPECT_EQ(registerCheckP2PConnection(&comm, &conn, nullptr, 1, &needReg), ncclSuccess);
+  EXPECT_TRUE(needReg);
+}
+
+TEST(CollReg, RegisterCheckP2P_ConnectedWithNoP2PFlag) {
+  ncclComm comm = {};
+  ncclConnector conn = {};
+  conn.connected = 1;
+  conn.conn.flags = 0;
+
+  bool needReg = true;
+  EXPECT_EQ(registerCheckP2PConnection(&comm, &conn, nullptr, 1, &needReg), ncclSuccess);
+  EXPECT_FALSE(needReg);
+}
+
+TEST(CollReg, RegisterCheckP2P_NotConnected_CanConnectTrue) {
+  // Save original canConnect
+  ScopedCanConnectOverride _(&(*ncclTransports[0]), mockCanConnect);
+  auto* originalCanConnect = ncclTransports[0]->canConnect;
+  ncclTransports[0]->canConnect = mockCanConnect;
+
+  ncclComm comm = {};
+  comm.rank = 0;
+
+  ncclPeerInfo peers[2] = {};
+  comm.peerInfo = peers;
+
+  // Set myInfo and peerInfo
+  peers[0].hostHash = 123;
+  peers[0].pidHash = 999;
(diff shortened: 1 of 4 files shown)


thananon (approved):

test/CollRegTests.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "CollRegUtils.hpp"
+
+namespace RcclUnitTesting
+{
+  
+TEST(CollReg, RegisterCheckP2P_ConnectedWithP2PFlag) {
+  ncclComm comm = {};
+  ncclConnector conn = {};
+  conn.connected = 1;
+  conn.conn.flags = NCCL_P2P_WRITE;
```
awelling2801, line 21: Check for NCCL_P2P_READ as well. Add another test

test/CollRegTests.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  bool needReg = false;
+  EXPECT_EQ(registerCheckP2PConnection(&comm, &conn, nullptr, 1, &needReg), ncclSuccess);
+  EXPECT_TRUE(needReg);
+}
+
+TEST(CollReg, RegisterCheckP2P_ConnectedWithNoP2PFlag) {
```
awelling2801, line 28: This test should go first.

test/CollRegTests.cpp
```diff
@@ -0,0 +1,102 @@
...
+TEST(CollReg, RegisterCheckP2P_NotConnected_CanConnectFalse) {
+  ScopedCanConnectOverride _(&(*ncclTransports[0]), mockCanConnectFalse); 
+
+  ncclComm comm = {};
+  comm.rank = 0;
+  ncclPeerInfo peers[2] = {};
+  comm.peerInfo = peers;
+
+  // Set up myInfo and peerInfo to block P2P
+  peers[0].hostHash = 123; peers[0].pidHash = 999;
```
awelling2801, line 87: Reduce to check what is the data required.

venksubramd:
Do we need tests for ncclRegisterCollNvlsBuffers and ncclRegisterCollBuffers. The second function especially is a very large one, sadly.

registerCheckP2PConnection
The tests for registerCheckP2PConnection are fairly good.

venksubramd:
From earlier review: Do we need tests for ncclRegisterCollNvlsBuffers and ncclRegisterCollBuffers. The second function especially is a very large one, sadly. If we do have tests for these, could you please check if those tests actually assert behavior and do not merely aim for code coverage?

Reviewing ShmTests:
The tests for shmSendProxyConnect look good. They focus not only on the return result but also on the values being set. Nice test for the negative condition as well.

Likewise the tests for shmRecvProxyConnect look good.

The test TEST_F(ShmTests, ShmSendRecvProxyFree) is checking for return condition and also uses a if to invoke shmSendProxyFree or shmRecvProxyFree. We need tests for the actual behavior of this code. If such a test exists elsewhere we need to ascertain that those tests are actually doing the right thing (checking for behavior) and not focusing only on code coverage.

Could we please separate the tests for shmSendProxyProgress into multiple tests so each test can focus on one condition and we can avoid one big test with multiple independent asserts?

Likewise for shmRecvProxyProgress, it will be great if we can separate the tests instead of placing multiple independent asserts into one test.

The tests for shmSendProxySetup look pretty good. Likewise tests for shmRecvProxySetup are pretty good.

Nice work on ncclShmAllocateShareableBuffer with multiple tests that focus on different aspects. Likewise nice work in tests for ncclShmImportShareableBuffer.

Different tests are calling ncclShmIpcClose, but do we need tests for the ncclShmIpcClose function?

Do we need a test for shmCanConnect, shmSendSetup, shmRecvSetup, shmSendConnect, shmRecvConnect, 
These functions are being exposed through the data structure at the bottom of the file.

test/CollRegTests.cpp
awelling2801: Unnecessary

test/CollRegTests.cpp
awelling2801: Unnecessary

awelling2801:
@venksubramd ShmTests has already been addressed in other PR. This branch was created from ShmTests branch.

Most of ncclRegisterCollNvlsBuffers and ncclRegisterCollBuffers is under an `#if CUDART_VERSION >= 11030` and will not be used by AMD, rest has been covered by other functions.

venksubramd:
@awelling2801  Sounds good, thank you. This is ready to be closed.

rahulvaidya20 (approved):

PR #1707, opened 2025-05-22
 Fix Concurrency problems in mscclInit() function

PJAvinash:
## Details

**Work item:** "Internal"
**What were the changes?**  
Segmentation fault due to improper use of lock guard

In SplitComm, different communicators may have same ranks in a Process, and shared mutable data suffers data race problems
Altering the scope of mutex lock guard
Read-write on shared mutable data should be made atomic 

**Additional Documentation:**  
mscclinit state has following valid transitions FALSE-> TRUE, FALSE -> FALSE, TRUE -> FALSE. Hence we need an additional mutex to avoid TOC-TOU problem in mscclInit()

Explicit memory barriers are not required as this program executes on hardware with Total Store Ordering.

## Approval Checklist

### DIFF
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -249,6 +249,8 @@ ncclResult_t mscclInit(ncclComm_t comm) {
 
     // freeAlgoHandles and needsProxy are initialized globally once and before algorithm pre-processing and connection
     if (!mscclInitialized(comm->rank)) {
+      static std::mutex initMutex;
+      std::lock_guard<std::mutex> lock(initMutex);
       status.freeAlgoHandles.resize(MSCCL_MAX_NUM_ALGOS);
       for (int i = 0; i < MSCCL_MAX_NUM_ALGOS; i++) {
         status.freeAlgoHandles[i] = MSCCL_MAX_NUM_ALGOS - i - 1;
--- a/src/misc/msccl/msccl_lifecycle.cc
+++ b/src/misc/msccl/msccl_lifecycle.cc
@@ -264,13 +264,16 @@ ncclResult_t mscclInit(ncclComm_t comm) {
         auto &m = status.algoMetas[i];
         if (m.nRanks == comm->nRanks) {
           // Load algorithms
-          if (status.rankToAlgoHandles[i].find(comm->rank) == status.rankToAlgoHandles[i].end()) {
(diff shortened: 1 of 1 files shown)


src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -264,31 +264,37 @@ ncclResult_t mscclInit(ncclComm_t comm) {
         auto &m = status.algoMetas[i];
         if (m.nRanks == comm->nRanks) {
           // Load algorithms
-          if (status.rankToAlgoHandles[i].find(comm->rank) == status.rankToAlgoHandles[i].end()) {
+          mscclAlgoHandle_t mscclAlgoHandle;
+          {
```
thananon, line 270: I understand the general idea but I am curious on the use of curly braces here. Is this for mutex scope? Do we need it?

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -264,31 +264,37 @@ ncclResult_t mscclInit(ncclComm_t comm) {
...
-          // Connect algorithms
-          mscclAlgoHandle_t mscclAlgoHandle = status.rankToAlgoHandles[i][comm->rank];
           if (status.connectedAlgos[comm].find(mscclAlgoHandle) == status.connectedAlgos[comm].end()) {
             NCCLCHECK(mscclSetupConnections(status.hostAlgos[mscclAlgoHandle], comm));
             status.connectedAlgos[comm].insert(mscclAlgoHandle);
           }
         }
       }
     }
+    {
```
thananon, line 286: same comment. Why these curly braces?

src/misc/msccl/msccl_lifecycle.cc
PJAvinash: That is for mutex scope. Once the mutex goes out of scope, the lock will be released.

corey-derochie-amd (changes requested):
The only reason the issue is resolved is because access to members of `status` are being mutually excluded. This fix only hides the true problem, which could pop up elsewhere.

The bug is that `status` is being accessed across threads, not that its access is not being protected. *By design*, `status` should not be shared between threads.

The *real* cause is that the map containing each `status` by rank uses only the rank index. It is not considering ranks of the same index on different communicators. (See msccl_status.cc)

thananon:
I think the PR is sufficient for this issue. If we want to rewrite MSCCL to be fully thread-safe, it should be a new scope.

PR #1715, opened 2025-05-29
SPLITCOMM design fix in src/misc/msccl 

PJAvinash:
## Details

We are using rank as a key to fetch mscclStatus , when comm split happens , different communicators in the same process space may have same rank. Ideally each communicator should have its own status. Hence we change the code to use comm instead of comm->rank

**Work item:** "Internal"

**What were the changes?**  
use comm instead of comm->rank, and necessary changes for the same

Better design 

making the code changes for better design and without breaking the existing intended behavior in default cases

**Additional Documentation:**  
static unordered_map<ncclComm_t, shared_ptr<mscclRankState>> rankStates is still not thread safe, but accesses now are based on ncclComm_t comm, which is unique for each thread or groups of threads associated to each communicator. Hence , it avoids some race problems.

Standalone.SplitComms_RankCheck works as intended after this fix.

## Approval Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,41 +2,41 @@
 
 Full documentation for RCCL is available at [https://rccl.readthedocs.io](https://rccl.readthedocs.io)
 
-## Unreleased - RCCL 2.24.3 for ROCm 6.5.0
+## Unreleased - RCCL 2.25.1 for ROCm 7.0.0
 
 ### Resolved issues
+
 * Resolved an issue when using more than 64 channels when multiple collectives are used in the same `ncclGroup()` call.
+* Fixed unit test failures in tests ending with `ManagedMem` and `ManagedMemGraph` suffixes.
+* Fixed the known issue "When splitting a communicator using `ncclCommSplit` in some GPU configurations, MSCCL initialization can cause a segmentation fault." with a design change to use comm instead of rank for mscclStatus. The Global map for comm -> mscclStatus is still not thread safe but should be explicitly handled by mutexes for read writes. This is tested for correctness, but we plan to use a thread safe map data structure in upcoming changes.
 
 ### Added
 
 * Added new GPU target `gfx950`.
-
-### Changed
-
-* Compatibility with NCCL 2.24.3
-
-## Unreleased - RCCL 2.23.4 for ROCm 6.4.1
-
-### Added
-
+* Added support for `unroll=1` in device-code generation to improve performance
+* Set a default of 112 channels for a single node with `8 * gfx950`
 * Added MSCCL support for multinode gfx942/gfx950 (i.e., 16 and 32 GPUs). To enable, set the
   environment variable `RCCL_MSCCL_FORCE_ENABLE=1`. Max message size for MSCCL AllGather usage is `12292 * sizeof(datatype) * nGPUs`.
-* Added synchronization before destroying proxy thread to fix a rare hang caused by early termination.
 
 ### Changed
 
 * Compatibility with NCCL 2.23.4
+* Compatibility with NCCL 2.24.3
+* Compatibility with NCCL 2.25.1
+
+## RCCL 2.22.3 for ROCm 6.4.1
 
 ### Resolved issues
 
 * Fixed the accuracy issue for MSCCLPP `allreduce7` kernel in graph mode.
 * Fixed IntraNet performance.
+* Fixed an issue where, in rare circumstances, the application could stop responding due to a proxy thread synchronization issue.
 
 ### Known issues
 
 * When splitting a communicator using `ncclCommSplit` in some GPU configurations, MSCCL initialization can cause a segmentation fault.
   The recommended workaround is to disable MSCCL with `export RCCL_MSCCL_ENABLE=0`.
-* Within the RCCL-UnitTests test suite, failures occur in tests ending with the `.ManagedMem` and `.ManagedMemGraph` suffixes. These failures only affect the test results and do not affect the RCCL component itself. This issue will be resolved in the next major release.
+* Within the RCCL-UnitTests test suite, failures occur in tests ending with the `ManagedMem` and `ManagedMemGraph` suffixes. These failures only affect the test results and do not affect the RCCL component itself. This issue will be resolved in the next major release.
 
 ## RCCL 2.22.3 for ROCm 6.4.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 * Resolved an issue when using more than 64 channels when multiple collectives are used in the same `ncclGroup()` call.
 * Fixed unit test failures in tests ending with `ManagedMem` and `ManagedMemGraph` suffixes.
-* Fixed the known issue "When splitting a communicator using `ncclCommSplit` in some GPU configurations, MSCCL initialization can cause a segmentation fault." with a design change to use comm instead of rank for mscclStatus. The Global map for comm -> mscclStatus is still not thread safe but should be explicitly handled by mutexes for read writes. This is tested for correctness, but we plan to use a thread safe map data structure in upcoming changes.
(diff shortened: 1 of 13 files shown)


wenkaidu (approved):

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -243,12 +244,14 @@ ncclResult_t mscclSchedulerInit(ncclComm_t comm, int* numChannelsRequired) {
...
+ncclResult_t mscclInit(const ncclComm_t comm) {
   {
-    mscclStatus& status = mscclGetStatus(comm->rank);
+    mscclStatus& status = mscclGetStatus(comm);
 
     // freeAlgoHandles and needsProxy are initialized globally once and before algorithm pre-processing and connection
-    if (!mscclInitialized(comm->rank)) {
+    if (!mscclInitialized(comm)) {
+      static std::mutex initMutex;
+      std::lock_guard<std::mutex> lock(initMutex);
```
thananon, line 254: Does this `initMutex`'s scope covers what `loadAlgoMutex` tries to cover? I'm asking because I am not sure how C++ mutex works.

thananon (commented):
Generally I think the interface change is done well. I just have a question about the mutex scope.

thananon (changes requested):
We should remove this message from CHANGELOG.md and add an entry to Resolved issues section.

```
When splitting a communicator using ncclCommSplit in some GPU configurations, MSCCL initialization can cause a segmentation fault. The recommended workaround is to disable MSCCL with export RCCL_MSCCL_ENABLE=0.
```

src/misc/msccl/msccl_lifecycle.cc
corey-derochie-amd: The `status` object is no longer thread-shared, so this mutex is no longer necessary.

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -264,31 +267,37 @@ ncclResult_t mscclInit(ncclComm_t comm) {
...
+          mscclAlgoHandle_t mscclAlgoHandle;
+          {
             static std::mutex loadAlgoMutex;
             std::lock_guard<std::mutex> lock(loadAlgoMutex);
-            NCCLCHECK(mscclLoadAlgo(m.filePath.c_str(), &(status.rankToAlgoHandles[i][comm->rank]), comm->rank));
+            if (status.rankToAlgoHandles[i].find(comm->rank) == status.rankToAlgoHandles[i].end()){
+              NCCLCHECK(mscclLoadAlgo(m.filePath.c_str(), &(status.rankToAlgoHandles[i][comm->rank]), comm));
+            }
+            // Connect algorithms
+            mscclAlgoHandle = status.rankToAlgoHandles[i][comm->rank];
```
corey-derochie-amd, line 278: This mutex is only meant to protect the `mscclLoadAlgo()` call, not use of `status`, so this change is no longer needed.

src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -264,31 +267,37 @@ ncclResult_t mscclInit(ncclComm_t comm) {
...
-    status.lastStream = nullptr;
-    NCCLCHECK(mscclInitWorkFifoStatus(&(status.defaultWorkFifoStatus)));
+      status.workIndex = 1;
+      NCCLCHECK(ncclCudaCalloc(&status.syncFlags, MSCCL_MAX_NUM_THREAD_BLOCKS));
+      status.lastStream = nullptr;
+      NCCLCHECK(mscclInitWorkFifoStatus(&(status.defaultWorkFifoStatus)));
 
-    mscclSetInitialized(comm->rank);
+      mscclSetInitialized(comm);
+    }
```
corey-derochie-amd, line 300: Are `mscclInitialized()`/`mscclSetInitialized()` already protected internally? Not sure that this mutex is necessary either.

src/misc/msccl/msccl_status.cc
```diff
@@ -7,68 +7,73 @@
...
 };
 
 static mutex rankStatesMutex;
-static unordered_map<int, shared_ptr<mscclRankState>> rankStates;
+/*
+ * @brief rankStates is intended to hold mscclRankState for each communicator in a rccl process.
+ * "rankStates" is not threadsafe, hence read/writes on this data strcutures need to be handled explicitly by
+ * block of code that is accessing the elements in this map using a lock guard or any mutual exclusion device.
+ */
+static unordered_map<ncclComm_t, shared_ptr<mscclRankState>> rankStates;
```
corey-derochie-amd, line 32: Question: can a logical rank get a new communicator partway through the process?

The whole reason we needed this solution in the first place is because we would set up ranks in one thread, and then run the collective for that rank in a new worker thread, so it couldn't be thread_local. I'm just worried that a similar discontinuity could exist with communicators.

Thoughts? @wenkaidu @thananon ?

corey-derochie-amd:
> We should remove this message from CHANGELOG.md and add an entry to Resolved issues section.
> 
> ```
> When splitting a communicator using ncclCommSplit in some GPU configurations, MSCCL initialization can cause a segmentation fault. The recommended workaround is to disable MSCCL with export RCCL_MSCCL_ENABLE=0.
> ```

We should not remove the note from a previous versions changes, only add that it was Resolved in the new version.

corey-derochie-amd:
These changes look correct and are exactly what I expect. I only had a couple questions.

Once the changelog is updated, this is probably good to go in.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 * Resolved an issue when using more than 64 channels when multiple collectives are used in the same `ncclGroup()` call.
 * Fixed unit test failures in tests ending with `ManagedMem` and `ManagedMemGraph` suffixes.
+* Fixed the known issue "When splitting a communicator using `ncclCommSplit` in some GPU configurations, MSCCL initialization can cause a segmentation fault." with a design change to use comm instead of rank for mscclStatus. The Global map for comm -> mscclStatus is still not thread safe but should be explicitly handled by mutexes for read writes. This is tested for correctness, but we plan to use a thread safe map data structure in upcoming changes.
```
```suggestion
* Fixed the known issue "When splitting a communicator using `ncclCommSplit` in some GPU configurations, MSCCL initialization can cause a segmentation fault." with a design change to use `comm` instead of `rank` for `mscclStatus`. The Global map for `comm` to `mscclStatus` is still not thread safe but should be explicitly handled by mutexes for read writes. This is tested for correctness, but there is a plan to use a thread-safe map data structure in upcoming changes.
```

amd-jnovotny (approved):
Looks good. Just some minor suggestions.

src/misc/msccl/msccl_lifecycle.cc
PJAvinash: This mutex is necessary to avoid Time of Check - Time of Update concurrency bug. The Init status have following valid transitions FALSE -> FALSE, FALSE -> TRUE , TRUE -> FALSE.  A race condition on mscclStatus from different threads but same communicator may force TRUE -> TRUE transition , and that fails the assertion in code for valid transitions. A mutex will make the pair of read and possible write atomic.

amd-jnovotny (approved):

src/misc/msccl/msccl_lifecycle.cc
PJAvinash: This will be addressed in a subsequent PR for making global map thread safe.

src/misc/msccl/msccl_lifecycle.cc
PJAvinash: As discussed, even if both reads and writes on a shared data structure are protected with locks , the TOC-TOU problem can still arise as read along with a  write as a unit is not atomic.

PJAvinash:
> We should remove this message from CHANGELOG.md and add an entry to Resolved issues section.
> 
> ```
> When splitting a communicator using ncclCommSplit in some GPU configurations, MSCCL initialization can cause a segmentation fault. The recommended workaround is to disable MSCCL with export RCCL_MSCCL_ENABLE=0.
> ```
The CHANGELOG.md is updated as requested.

thananon (approved):
Per discussion, this PR is functional. We are getting this merged. There will be another PR to improve upon this issue.

For the record, the CI has been completed successfully but there is a new change in CHANGELOG.md that triggers a rerun. Since CHANGELOG.md has no functional impact, we agree that this PR is passing the CI.

src/misc/msccl/msccl_status.cc
thananon: >  can a logical rank get a new communicator partway through the process?

That would be... unnatural.

PR #1720, opened 2025-06-03
Remove 'warpSize' compiler constant as it is deprecated in ROCm 7.0.

thananon:
## Details

**Work item:** SWDEV-535599

**What were the changes?**  
Remove `warpSize` compiler constant from our code as it is deprecated in ROCm 7.0.

Constant deprecated in ROCm 7.0.

On the device side, relying on header detection of __GFX9__ to set the correct warp size. For the host, we use runtime querying to set warp size. 

**Additional Documentation:**  
This PR untangles the device/host WARP_SIZE macro. I went through the code to make sure that every reference of WARP_SIZE on host side is addressed and leave the device side untouched.

## Approval Checklist

### DIFF
--- a/src/enqueue.cc
+++ b/src/enqueue.cc
@@ -52,6 +52,13 @@ static ncclKernelMatch const ncclKerns[3] = {
 };
 #endif
 
+static int rcclProtoGrainSize(int proto, ncclComm *comm){
+  return proto == NCCL_PROTO_LL ? 16 :
+        proto == NCCL_PROTO_LL128 ? comm->WarpSize*NCCL_LL128_SHMEM_ELEMS_PER_THREAD/NCCL_LL128_LINEELEMS*NCCL_LL128_DATAELEMS*sizeof(uint64_t) :
+        proto == NCCL_PROTO_SIMPLE ? 512 :
+        -1;
+}
+
 NCCL_PARAM(L1SharedMemoryCarveout, "L1_SHARED_MEMORY_CARVEOUT", 0);
 
 // Returns maximum kernel stack size of all CUDA kernels
@@ -197,7 +204,7 @@ static void finishPlan(struct ncclComm* comm, struct ncclKernelPlan* plan) {
   size_t workBytes = plan->workBytes;
   size_t batchBytes = plan->nWorkBatches*sizeof(struct ncclDevWorkBatch);
 
-  plan->threadPerBlock = std::max(plan->threadPerBlock, NCCL_MIN_NTHREADS);
+  plan->threadPerBlock = std::max(plan->threadPerBlock, 4*comm->WarpSize /*NCCL_MIN_NTHREADS*/);
 
   // If we can fit everything into the kernel args we do so.
   if (sizeof(ncclDevKernelArgs) + batchBytes + workBytes <= comm->workArgsBytes) {
@@ -668,7 +675,8 @@ static ncclResult_t scheduleCollTasksToPlan(
       }
 
       uint32_t chunkSize, directFlags=0;
-      size_t grainSize = ncclProtoGrainSize(task->protocol);
+      size_t grainSize = rcclProtoGrainSize(task->protocol, comm);
+
       if (countLo != 0) {
         NCCLCHECK(calcCollChunking(comm, task, /*nChannels=*/1, globalBytesPerElement*countLo, &chunkSize, &directFlags, &proxyOpLo));
         devWork->cbd.chunkGrainsLo = chunkSize/grainSize;
@@ -775,9 +783,9 @@ static ncclResult_t scheduleCollTasksToPlan(
           ncclProtoToString(task->protocol),
           (long)task->count, task->devFuncId, devWork->channelLo, devWork->channelHi,
           (long)devWork->cbd.countLo, (long)devWork->cbd.countMid, (long)devWork->cbd.countHi,
-          int(devWork->cbd.chunkGrainsLo*ncclProtoGrainSize(task->protocol)),
-          int(devWork->cbd.chunkGrainsMid*ncclProtoGrainSize(task->protocol)),
-          int(devWork->cbd.chunkGrainsHi*ncclProtoGrainSize(task->protocol)));
+          int(devWork->cbd.chunkGrainsLo*rcclProtoGrainSize(task->protocol, comm)),
+          int(devWork->cbd.chunkGrainsMid*rcclProtoGrainSize(task->protocol, comm)),
+          int(devWork->cbd.chunkGrainsHi*rcclProtoGrainSize(task->protocol), comm));
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/enqueue.cc
```diff
@@ -52,6 +52,13 @@ static ncclKernelMatch const ncclKerns[3] = {
 };
 #endif
 
+static int rcclProtoGrainSize(int proto, ncclComm *comm){
+  return proto == NCCL_PROTO_LL ? 16 :
+        proto == NCCL_PROTO_LL128 ? comm->WarpSize*NCCL_LL128_SHMEM_ELEMS_PER_THREAD/NCCL_LL128_LINEELEMS*NCCL_LL128_DATAELEMS*sizeof(uint64_t) :
```
corey-derochie-amd: To aid readability
```suggestion
        proto == NCCL_PROTO_LL128 ? comm->WarpSize*(NCCL_LL128_SHMEM_ELEMS_PER_THREAD/NCCL_LL128_LINEELEMS)*NCCL_LL128_DATAELEMS*sizeof(uint64_t) :
```

src/enqueue.cc
```diff
@@ -1789,11 +1797,11 @@ static ncclResult_t topoGetAlgoInfo(
...
   }
   if (info->protocol == NCCL_PROTO_SIMPLE) {
-    if (info->algorithm == NCCL_ALGO_RING) nt += WARP_SIZE; // Extra warp for sync
+    if (info->algorithm == NCCL_ALGO_RING) nt += comm->WarpSize; // Extra warp for sync
     // More threads or sync warps needed due to split thread model
-    if (info->algorithm == NCCL_ALGO_TREE) nt += 4*WARP_SIZE;
+    if (info->algorithm == NCCL_ALGO_TREE) nt += 4*comm->WarpSize;
   }
-  nt = nt/WARP_SIZE < 3 ? 3*WARP_SIZE : nt;
+  nt = nt/comm->WarpSize < 3 ? 3*comm->WarpSize : nt;
```
nusislam, line 1827: Where are you storing the `WARPSIZE` in `comm->WarpSize`?

src/include/device.h
```diff
@@ -316,7 +321,7 @@ struct alignas(16) ncclDevWorkColl {
 };
 
 
-__host__ __device__ constexpr int ncclProtoGrainSize(int proto) {
+__device__ constexpr int ncclProtoGrainSize(int proto) {
```
nusislam, line 333: Why is the specifier changed?

src/enqueue.cc
```diff
@@ -52,6 +52,13 @@ static ncclKernelMatch const ncclKerns[3] = {
 };
 #endif
 
+static int rcclProtoGrainSize(int proto, ncclComm *comm){
+  return proto == NCCL_PROTO_LL ? 16 :
+        proto == NCCL_PROTO_LL128 ? comm->WarpSize*NCCL_LL128_SHMEM_ELEMS_PER_THREAD/NCCL_LL128_LINEELEMS*NCCL_LL128_DATAELEMS*sizeof(uint64_t) :
+        proto == NCCL_PROTO_SIMPLE ? 512 :
+        -1;
```
corey-derochie-amd:
```suggestion
  switch (proto) {
    case NCCL_PROTO_LL: return 16;
    case NCCL_PROTO_LL128: return comm->WarpSize*(NCCL_LL128_SHMEM_ELEMS_PER_THREAD/NCCL_LL128_LINEELEMS)*NCCL_LL128_DATAELEMS*sizeof(uint64_t);
    case NCCL_PROTO_SIMPLE: return 512;
    default: return -1;
  }
```

src/include/device.h
```diff
@@ -67,7 +67,12 @@ union ncclLLFifoLine {
   int4 i4;
 };
 
-#define WARP_SIZE warpSize
+#if defined(__GFX9__)
+#define WARP_SIZE 64
+#else
+#define WARP_SIZE 32
+#endif
```
corey-derochie-amd: `WARP_SIZE` should *only* be defined for device compilation to avoid misuse.
```suggestion
#if __HIP_DEVICE_COMPILE__
#if defined(__GFX9__)
#define WARP_SIZE 64
#else
#define WARP_SIZE 32
#endif
#endif
```

src/include/device.h
corey-derochie-amd: Use of `WARP_SIZE` is no longer valid in host code.

src/enqueue.cc
thananon: We have this for a while now: https://github.com/ROCm/rccl/blob/e94b36024615b65ddcd372ac39d9e9f5f43f685f/src/init.cc#L681

src/enqueue.cc
thananon: I can change this but this is a direct copy of the old function. Do you want to change?
https://github.com/ROCm/rccl/blob/e94b36024615b65ddcd372ac39d9e9f5f43f685f/src/include/device.h#L319-L324

corey-derochie-amd: It's "new code" so I don't see a problem with it being modified. Alternatively, if it must always be kept in sync with `ncclProtoGrainSize()`, maybe it could be written as a template or something. Not sure if that's necessary though.

src/include/device.h
thananon: This somehow does not work as it seems like `__device__` functions still get compiled on host (just cant call from host). If we want this, we have to guard all our device functions with `if __HIP_DEVICE__COMPILE__` which sounds extreme.

In the current PR, `WARP_SIZE` will still be defined as 32 but it will not be used as I removed `__host__` decoration from functions that reference it.

thananon:
@BertanDogancay should aware of this as this will make the next sync more painful.

src/enqueue.cc
PJAvinash: Not initializing warpSize in comm, or use of warpSize before initialization of communicator, may lead to division by zero errors or Undefined behavior. It is better if we can find a way to define warpSize as const , and must be initialized before accessing an fields in the object.

thananon: @corey-derochie-amd advise against global const for now.

corey-derochie-amd (approved):

nusislam (approved):

src/include/device.h
```diff
@@ -67,7 +67,21 @@ union ncclLLFifoLine {
...
+  #else
+  #define WARP_SIZE 32
+  #endif
+#else
+ /* IMPORTANT:
+  * WARP_SIZE should NEVER be referenced by host code in RCCL. It is defined here
+  * solely as a workaround to allow RCCL to compile, since the host still compiles __device__ functions,
+  * and WARP_SIZE needs to be defined. These __device__ functions will not be called from the host.
+  * The host warp size is handled in src/enqueue.cc by calling hipDeviceGetAttributes(). */
+  #define WARP_SIZE 32
```
corey-derochie-amd, line 82: After solving the other problem, does the host-side value matter? If possible, this value should be "weird" to communicate something is wrong if you use it.

src/include/device.h
```diff
@@ -67,7 +67,21 @@ union ncclLLFifoLine {
   int4 i4;
 };
 
-#define WARP_SIZE warpSize
+#if __HIP_DEVICE_COMPILE__
+  #if defined(__GFX9__)
+  #define WARP_SIZE 64
+  #else
+  #define WARP_SIZE 32
+  #endif
```
corey-derochie-amd, line 75: I'm a bit nervous about a blanket `#else` here, as it will silently compile on future architectures. Instead, if we can specify this as explicitly as possible, it will fail to compile on an architecture we forget to specify, which is good.

src/include/device.h
thananon: 0 and -1 does not work as it is used to define array size.
1 does not work as it will cause some assert (some number/size being too low)
Other number is probably fine but I think 32 is as good as any number. What do you think?

src/include/device.h
thananon: This is following[ ROCm guideline](https://rocm.docs.amd.com/en/docs-6.4.0/about/release-notes.html#:~:text=For%20cases%20where,if%20defined(__GFX9__)). I am not sure about alternatives.

corey-derochie-amd (approved):

thananon:
Merging now with outstanding CI to make staging run.
- Azure CI (Known issue, MSCCLPP)
- Extra Pipeline (Failed in docker stage)

PR #1725, opened 2025-06-04
Added tests for transport.cc

awelling2801:
**Work item:** _"Internal"

**What were the changes?**  
Added Unit tests to increase code coverage for transport.cc
This PR adds two files - test/TransportTests.cpp comprising of 6 tests and a supporting file - test/common/TransportUtils.hpp

The test improved code coverage for transport.cc
|--|-- | -- | -- | --
Default Coverage |  60.00% (6/10) |    64.31% (227/353)  |    53.27% (472/886)  |   38.89% (189/486)
Updated Coverage |   100.00% (10/10) |    95.16% (354/372)  |  74.07% (657/887)  |  54.53% (265/486)

The outcome was achieved by adding a TransportTests.cpp file comprising of the tests.

## Approval Checklist

### DIFF
--- a/test/TransportTests.cpp
+++ b/test/TransportTests.cpp
@@ -1,0 +1,243 @@
+/*************************************************************************
+ * Copyright (c) 2023 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include <gtest/gtest.h>
+#include <rccl/rccl.h>
+#include "TestBed.hpp"
+#include "TransportUtils.hpp"
+
+namespace RcclUnitTesting
+{
+
+TEST(TransportTest, CollNetRecvSetup) {
+  constexpr int nranks = 2;
+  constexpr int nNodes = 2;
+
+  // --- Setup comm ---
+  ncclComm comm = {};
+  comm.rank = 0;
+  comm.nRanks = nranks;
+  comm.nNodes = nNodes;
+  comm.node = 0;
+  ncclPeerInfo peerInfo[3] = {};
+  comm.peerInfo = peerInfo;
+
+  // --- Setup channel ---
+  ncclChannel channel = {};
+  static ncclChannelPeer peerArray[3];
+  static ncclChannelPeer* peerPtrs[3] = {
+    &peerArray[0], &peerArray[1], &peerArray[2]
+  };
+  channel.peers = peerPtrs;
+
+  // Step 1: Allocate device-side array of ncclDevChannelPeer
+  ncclDevChannelPeer* devPeerArrayDevice;
+  hipMalloc(&devPeerArrayDevice, sizeof(ncclDevChannelPeer) * 3);
+
+  // Step 2: Create host-side array of device pointers
+  ncclDevChannelPeer* devPeerPtrsHost[3] = {
+    devPeerArrayDevice + 0,
+    devPeerArrayDevice + 1,
+    devPeerArrayDevice + 2
+  };
+
+  // Step 3: Allocate device-side array of device pointers
+  ncclDevChannelPeer** devPeerPtrsDevice;
+  hipMalloc(&devPeerPtrsDevice, sizeof(ncclDevChannelPeer*) * 3);
+
+  // Step 4: Copy host-side array of device pointers to device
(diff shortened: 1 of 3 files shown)


thananon:
waiting for conflict resolution.

thananon:
Looks like missing symbols/ something in linking stage on our Azure pipeline.

```
ld.lld: error: undefined symbol: ncclTransportCollNetSetup(ncclComm*, ncclTopoGraph*, ncclChannel*, int, int, int, int, ncclConnect*)

ld.lld: error: undefined symbol: ncclTransportCollNetFree(ncclComm*)
>>> referenced by TransportTests.cpp

ld.lld: error: undefined symbol: dumpData(ncclConnect*, int)
>>> referenced by TransportTests.cpp
```

venksubramd:
Do we need tests for ncclTransportP2pConnect, ncclTransportCheckP2pType?

There are two tests for ncclTransportCollNetSetup and both check the function ran successfully. Are there situations where it could fail and if so, we need tests for that.

The test for ncclTransportCollNetFree needs to check for values besides return value of success.

awelling2801:
@venksubramd 
ncclTransportP2pConnect and ncclTransportCheckP2pType have code coverage for all cases and do not need their own tests. 
For ncclTransportCollNetSetup and ncclTransportCollNetFree could use failure cases with mocked functions but it won't help code coverage as such

thananon (approved):

awelling2801:
Mocking system-level functions would require significant effort and likely a separate PR. I’m deferring this for now and will revisit it later.

venksubramd:
@awelling2801 

I am hoping we are not focusing on code coverage as a measure to achieve. We can write useless tests to get high code coverage. It would be good to take another look at these for the true reasons to write the tests instead of aiming for coverage numbers (alone).

atulkulk:
> @awelling2801
> 
> I am hoping we are not focusing on code coverage as a measure to achieve. We can write useless tests to get high code coverage. It would be good to take another look at these for the true reasons to write the tests instead of aiming for coverage numbers (alone).

Ticket for mocking efforts: [ GPUAI-5989 ]

test/TransportTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+  // Allocate peer array for the channel (comm.nRanks + 1 for dummy peer)
+  channel->peers = new ncclChannelPeer*[comm.nRanks + 1];
+  for (int r = 0; r <= comm.nRanks; ++r) {
+    channel->peers[r] = nullptr;
+  }
+
+  // Create dummy peer at index nRanks
+  ncclChannelPeer* peer = new ncclChannelPeer();
+  channel->peers[comm.nRanks] = peer;
```
corey-derochie-amd, line 198: You could construct on the stack instead of managing heap allocation.
```suggestion
  ncclChannelPeer peer;
  channel->peers[comm.nRanks] = &peer;
```

test/TransportTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+  comm.nRanks = 0; // So comm.channels[0].peers[0] is accessed
+
+  // Access embedded array directly (don't assign to comm.channels)
+  ncclChannel* channel = &comm.channels[0];
+
+  // Allocate peer array for the channel (comm.nRanks + 1 for dummy peer)
+  channel->peers = new ncclChannelPeer*[comm.nRanks + 1];
+  for (int r = 0; r <= comm.nRanks; ++r) {
+    channel->peers[r] = nullptr;
+  }
```
corey-derochie-amd, line 194: You could use a vector instead of managing head allocation.
```suggestion
  std::vector<ncclChannelPeer*>(comm.nRanks + 1) peers;
  channel->peers = peers.data();
```

corey-derochie-amd (approved):
lgtm, suggestions aren't necessary, just for consideration

test/TransportTests.cpp
awelling2801: Thanks for the suggestions — they’re all fair points.  I’ve reviewed it again and I believe the PR is good to go. It meets the criteria and the tests are meaningful. Any remaining gaps can be handled in follow-ups.

test/TransportTests.cpp
awelling2801: Thanks for the suggestions — they’re all fair points. I’ve reviewed it again and I believe the PR is good to go. It meets the criteria and the tests are meaningful. Any remaining gaps can be handled in follow-ups.

PR #1727, opened 2025-06-05
Compiler Warning Fixes 

PJAvinash:
## Details

Addressing (most) of the compiler warnings with as minimal changes as possible.

**Work item:** "Internal"
**What were the changes?**  
Removing unused variables
Relocation of #If defined(...) or variable definitions to define variables if necessary
Relocation of some function definitions in msccl ( from .h to .cc)

Too many compiler warning while building rccl

**Additional Documentation:**  
- There are few warning from ext-src directory, They will be addressed as per our future plans.
- There is a minor bug fix in alt_rsmi.cc (Based on the implementation context and compiler warning)

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -832,6 +832,7 @@ target_compile_options(rccl PRIVATE -Werror=uninitialized)
 target_compile_options(rccl PRIVATE -Werror=sometimes-uninitialized)
 target_compile_options(rccl PRIVATE -Wall)
 target_compile_options(rccl PRIVATE -Wno-format-nonliteral)
+target_compile_options(rccl PRIVATE -Wno-unused-function)
 target_compile_options(rccl PRIVATE -fgpu-rdc)               # Generate relocatable device code (required for extern __shared__)
 target_compile_options(rccl PRIVATE -fvisibility=hidden)     # Set symbol visibility to hidden by default
 if (HAVE_KERNARG_PRELOAD)
--- a/src/device/primitives.h
+++ b/src/device/primitives.h
@@ -25,7 +25,7 @@
   if (nthreads == NCCL_MAX_NTHREADS) { \
     __THREAD_FENCE; __builtin_amdgcn_s_barrier(); \
   } else { \
-    const int w = threadIdx.x/WARP_SIZE; \
+    /**const int w = threadIdx.x/WARP_SIZE //unused var**/; \
     const int wid = threadIdx.x%WARP_SIZE; \
     if (wid == 0) { \
       barrier_next += nthreads/WARP_SIZE; \
--- a/src/graph/tuning.cc
+++ b/src/graph/tuning.cc
@@ -167,7 +167,7 @@ static struct tuningModel tuning_model_0 {
     { 1.0, 0.8, 0.2, 1.0, 1.0, 0.3, 1.0, 0.1, 0.1, 0.2, 0.2, 0.1, 0.5, 1.0, 0.8, 0.8, 1.0, 0.9, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, },
   },
 
-  .llProtoRanges = {RCCL_LL_LIMITS_UNDEFINED},
+  .llProtoRanges = {{{RCCL_LL_LIMITS_UNDEFINED}}},
 };
 
 static struct tuningModel tuning_model_1 {
@@ -199,7 +199,7 @@ static struct tuningModel tuning_model_1 {
     { 0.3, 1.0, 0.3, 0.1, 0.1, 0.1, 0.3, 0.7, 1.0, 0.2, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.2, 0.3, 0.5, 0.9, 1.0, 1.0, 1.0, 1.0, },
   },
 
-  .llProtoRanges = {RCCL_LL_LIMITS_UNDEFINED},
+  .llProtoRanges = {{{RCCL_LL_LIMITS_UNDEFINED}}},
 };
 
 static struct tuningModel tuning_model_2 {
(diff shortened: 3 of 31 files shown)


src/misc/msccl/msccl_lifecycle.cc
```diff
@@ -30,7 +30,6 @@
 RCCL_PARAM(MscclEnabled, "MSCCL_ENABLE", 1);
 RCCL_PARAM(MscclForceEnabled, "MSCCL_FORCE_ENABLE", 0);
 RCCL_PARAM(MscclEnableSingleProcess, "MSCCL_ENABLE_SINGLE_PROCESS", 1);
-static const char* mscclAlgoFilePathEnv = "MSCCL_ALGO_FILE_PATH";
```
nileshnegi, line 33: ~~this is probably a typo... i think this var should be renamed to `mscclAlgoFilePath`~~
there is a missing getenv() for `mscclAlgoFilePathEnv` like [this line](https://github.com/ROCm/rccl/blob/develop/src/misc/msccl/msccl_lifecycle.cc#L141)

nileshnegi: or, we can eliminate this option and focus on the existing `MSCCL_ALGO_DIR`
best ensure that `MSCCL_ALGO_FILE_PATH` is no longer needed/supported

nileshnegi (approved):

src/device/primitives.h
```diff
@@ -25,7 +25,7 @@
   if (nthreads == NCCL_MAX_NTHREADS) { \
     __THREAD_FENCE; __builtin_amdgcn_s_barrier(); \
   } else { \
-    const int w = threadIdx.x/WARP_SIZE; \
+    /**const int w = threadIdx.x/WARP_SIZE //unused var**/; \
```
corey-derochie-amd, line 22: This can cause a sync merge conflict either way, so we should probably just delete this line. Although I do like that when we get the merge conflict, this comment will tell us why it was removed...

src/misc/alt_rsmi.cc
```diff
@@ -48,7 +48,7 @@ struct ARSMI_systemNode {
     std::string s_card;
 };
 
-static const char *kPathDRMRoot = "/sys/class/drm";
+// static const char *kPathDRMRoot = "/sys/class/drm"; // unused variable
```
corey-derochie-amd, line 51: Is this variable used in NCCL? We need a strategy around handling disabled NCCL code. In many places, we preserve NCCL code with `#if`/`#else` and "in theory" you could build RCCL on NVIDIA and it should work. (I don't think it's actually true though, and we don't test it, so should we bother?)

src/transport/p2p.cc
```diff
@@ -1089,6 +1086,9 @@ static ncclResult_t p2pProxyRegister(struct ncclProxyConnection* connection, str
     regAddr = (void*)((uintptr_t)regAddr + ipcExpInfo->offset);
   } else {
 #if CUDART_VERSION >= 11030
+    bool mapped = false;
+    bool imported = false;
+    CUmemGenericAllocationHandle handle;
```
corey-derochie-amd, line 1091: Alternative would be to add this, which would make it purely additive instead of removing NCCL code. Not sure if it's a good way to handle this kind of thing, though.

```
    #else
        (void)mapped;
        (void)imported;
        (void)handle;
```

src/misc/alt_rsmi.cc
thananon: I guarantee that it will not work as we have hip call all over the place.

thananon (approved):

src/misc/alt_rsmi.cc
corey-derochie-amd: Are the HIP calls only within guarded sections, though?

src/transport/coll_net.cc
corey-derochie-amd, line 1: What warning is addressed by this? I think these were moved up in a NCCL sync to conform to what NCCL is doing.

PJAvinash: Unused label "peermem:"

corey-derochie-amd (approved):

src/misc/alt_rsmi.cc
nileshnegi: this is RCCL-only file, right? should not affect NCCL syncs...

CMakeLists.txt
```diff
@@ -837,7 +837,10 @@ target_compile_options(rccl PRIVATE -Werror=uninitialized)
 target_compile_options(rccl PRIVATE -Werror=sometimes-uninitialized)
 target_compile_options(rccl PRIVATE -Wall)
```
mustafabar, line 838: Suggest to delete for non-verbose builds. However, turn on all warnings with `-Wall` when `VERBOSE=1` or `build_verbose` mode is enabled. This will allow us to see the warnings we trigger on demand, and at the same time, will not unnecessarily flood non-devs logs with clang warnings
```suggestion
```

mustafabar (approved):

CMakeLists.txt
mustafabar: @corey-derochie-amd and @nileshnegi. 
The suggestion above is probably better than fixing warnings that we did not create, only to create conflicts. The originators of the warning do use `-Wall` and may eventually address them.

src/graph/tuning.cc
```diff
@@ -167,7 +167,7 @@ static struct tuningModel tuning_model_0 {
     { 1.0, 0.8, 0.2, 1.0, 1.0, 0.3, 1.0, 0.1, 0.1, 0.2, 0.2, 0.1, 0.5, 1.0, 0.8, 0.8, 1.0, 0.9, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, },
   },
 
-  .llProtoRanges = {RCCL_LL_LIMITS_UNDEFINED},
+  .llProtoRanges = {{{RCCL_LL_LIMITS_UNDEFINED}}},
```
mustafabar, line 170: I'm not sure if this resolves the warning. Please double-check. I suggest to comment out all these lines that assign to RCCL_LL_LIMITS_UNDEFINED and rely on the fact that the array will be value initialized to 0. In any case, this line of code will not work as it thinks it does. https://github.com/ROCm/rccl/commit/70cb1c2cd42233abf5fddf8d8b2dbfd7aa73853e

thananon:
Lets bring this up to current and merge.

PJAvinash:
Closing this PR as a new PR for the same is opened on latest develop changes.

PR #1729, opened 2025-06-05
Support fused all reduce and elementwise operations

wenkaidu:
## Details

SWDEV-536678

**What were the changes?**  
Support fused all reduce and elementwise operations

Support fused all reduce and elementwise operations

Added new collective API ncclAllReduceWithBias(const void* sendbuff, void* recvbuff, size_t count, ncclDataType_t datatype, ncclRedOp_t op, ncclComm* comm, cudaStream_t stream, const void* acc)

**Additional Documentation:**  
For backward compatibility, API ncclAllReduceWithBias is available when RCCL_ALLREDUCE_WITH_BIAS is defined in rccl.h

## Approval Checklist


src/device/common_kernel.h
```diff
@@ -207,15 +207,218 @@ __device__ __forceinline__ void reduceCopyPacks(
...
+      #pragma unroll Unroll
+      for (int u=0; u < Unroll; u++) {
+        if (s < MultimemSrcs) {
+          // applyLoadMultimem uses relaxed semantics for same reason we use volatile below.
+          // coverity[dead_error_line]
+          tmp[u] = applyLoadMultimem<RedFn, BytePerPack>(redFn, minSrcs[s]);
+        } else {
+          // Use volatile loads in case credits are polled for with volatile (instead of acquire).
+          tmp[u] = ld_volatile_global<BytePerPack>(minSrcs[s]);
+          if (!biasLoaded && s == 1) {
```
ycui1984: I am wondering whether this will be dead code, line 287 will always be executed and set that as true, so here we are always going to be false, and not executed?

src/device/common_kernel.h
```diff
@@ -207,15 +207,218 @@ __device__ __forceinline__ void reduceCopyPacks(
...
+      #pragma unroll Unroll
+      // Yes, for some template arguments this code will be unreachable.  That's fine.
+      // coverity[dead_error_begin]
+      for (int u=0; u < Unroll; u++) {
+        // coverity[dead_error_condition]
+        if (d < MultimemDsts) {
+          multimem_st_global(minDsts[d], acc[u]);
+        } else {
+          if (d == 0) {
+            if (!biasLoaded) {
```
ycui1984: same here as above

src/device/common_kernel.h
```diff
@@ -207,15 +207,218 @@ __device__ __forceinline__ void reduceCopyPacks(
...
+              bias[u] = ld_volatile_global<BytePerPack>(accPtr);
+              accPtr += WARP_SIZE*BytePerPack;
+            }
+            st_global<BytePerPack>(minDsts[d], applyReduce(redFn, acc[u], bias[u]));
```
ycui1984, line 361: I doubt the optimization is pipelining, since it just move the load operation with other load operation (so we are batching the load at the same place). So that when we do the store, load is not in critical path anymore, right?

ycui1984 (commented):
I could see 10% improvements with all_reduce_with_bias_perf benchmarks, but I think the improvements are from batching the load together instead of pipelining

src/device/common_kernel.h
wenkaidu: L287 is executed when conditions in L267 is met

ycui1984: L268-> 398 are in a big while if I read correctly

src/device/common_kernel.h
wenkaidu: I think L291, i.e. "s < MinSrcs", will gate 2nd attempt of preloading acc. So it is not always set to true.

src/device/common_kernel.h
wenkaidu: Normally we don't wait for store in these functions. That will be handled in memory fence later

src/device/common_kernel.h
wenkaidu: You are right! I will push a fix soon

ycui1984 (approved):
lgtm

src/collectives.cc
```diff
@@ -133,6 +133,35 @@ ncclResult_t ncclAllReduce_impl(const void* sendbuff, void* recvbuff, size_t cou
...
+  struct ncclInfo info = { ncclFuncAllReduce, "AllReduce",
+    sendbuff, recvbuff, count, datatype, op, 0, comm, stream, /* Args */
+    ALLREDUCE_CHUNKSTEPS, comm -> rcclUseOneSlice ? ALLREDUCE_SLICESTEPS_SINGLE_NODE : ALLREDUCE_SLICESTEPS, acc };
+
+  if (!mscclIsCaller()) // when msccl falls back to
+  {
+    NCCLCHECK(Recorder::instance().record(rrAllReduce, info));
+  }
+
+  if (mscclAvailable(comm) && !mscclIsCaller() && acc == nullptr) {
```
AtlantaPepsi: whats the point of this `acc == nullptr` here given the previous check? In that case would this fused AR kernel ever call msccl, especially given msccl kernel dont take into the extra vector?

src/include/recorder.h
```diff
@@ -92,6 +92,7 @@ struct rcclApiCall {
   uint64_t              opCount = 0;
   const void*           sendbuff = NULL;
   void*                 recvbuff = NULL;
+  const void*           acc = NULL;
```
AtlantaPepsi, line 99: In this case, structured logging would just tell the pointer of the vector with no extra meaning. We should remove this and instead dump the whole vector as trailing info after this call entry, similar to how alltoallv dumps `sdispls` etc.. Perhaps we could defer change to recorder or adding record statement in fused AR kernel at all, and I can quickly add a patch after this PR merge, with both recorder/replayer support? I have to file another PR anyway on some similar API like ncclGroupSimulateEnd/InitRankConfig with trailing info to report

AtlantaPepsi (dismissed):

src/collectives.cc
wenkaidu: Yes, I will remove acc check here.

src/include/recorder.h
wenkaidu: Yes, this PR will only make existing Recorder UT to pass. Please add full fused AR replay functionality to your PR

AtlantaPepsi (approved):

BertanDogancay (approved):

BertanDogancay:
Do we want to keep the unroll changes reverted?

wenkaidu:
> Do we want to keep the unroll changes reverted?

Yes, this split single kernel to 6 smaller ones, cutting down number of device functions in kernel to 1/6

wenkaidu:
Failed CI: precheckin: Test ubuntu22 && 8gfx950
AllReduce:ManagedMem
I have seen random failures with "precheckin: Test ubuntu22 && 8gfx950" on develop branch as well. Thus ignoring this failure.

PR #1733, opened 2025-06-06
replayer update v0

AtlantaPepsi:
## Details
The new RCCL replayer, together with modifications for structured logging inside RCCL
**Work item:** Internal

**What were the changes?**  
Rewritten RCCL replayer, which will now be based on the **binary** output of recorder log instead of debug info. Should support replay of all public RCCL calls including collectives, communicator ops, UBR, with the exception custom redop, simulated groupEnd and communicator init with config.

**Additional Documentation:**  
Will need to update the documentation and user instructions later. At current stage testing options are limited, so we have to rely on feedback from dogfooding before adding future features to the replayer  

## Approval Checklist

### DIFF
--- a/src/collectives.cc
+++ b/src/collectives.cc
@@ -168,12 +168,14 @@ ncclResult_t ncclAllToAll_impl(const void* sendbuff, void* recvbuff, size_t coun
     int nRanks;
     NCCLCHECK(ncclCommCount(comm, &nRanks));
     if (count == 0) return ncclSuccess;
+    if (!mscclIsCaller()) Recorder::instance().skip(true);
     NCCLCHECK(ncclGroupStart());
     for (int r=0; r<nRanks; r++) {
       NCCLCHECK(ncclSend(((char*)sendbuff)+r*rankOffset, count, datatype, r, comm, stream));
       NCCLCHECK(ncclRecv(((char*)recvbuff)+r*rankOffset, count, datatype, r, comm, stream));
     }
     NCCLCHECK(ncclGroupEnd());
+    if (!mscclIsCaller()) Recorder::instance().skip(false);
     return ncclSuccess;
   }
 }
@@ -203,6 +205,7 @@ ncclResult_t ncclAllToAllv_impl(const void *sendbuff, const size_t sendcounts[],
 
   int nRanks;
   NCCLCHECK(ncclCommCount(comm, &nRanks));
+  if (!mscclIsCaller()) Recorder::instance().skip(true);
   NCCLCHECK(ncclGroupStart());
   for (int r=0; r<nRanks; r++) {
     NCCLCHECK(ncclSend(
@@ -221,6 +224,7 @@ ncclResult_t ncclAllToAllv_impl(const void *sendbuff, const size_t sendcounts[],
         stream));
   }
   NCCLCHECK(ncclGroupEnd());
+  if (!mscclIsCaller()) Recorder::instance().skip(false);
   return ncclSuccess;
 }
 
@@ -283,6 +287,7 @@ ncclResult_t ncclGather_impl(const void* sendbuff, void* recvbuff, size_t sendco
   if (sendcount == 0) return ncclSuccess;
   int rank;
   NCCLCHECK(ncclCommUserRank(comm, &rank));
+  if (!mscclIsCaller()) Recorder::instance().skip(true);
   NCCLCHECK(ncclGroupStart());
   if (rank == root) {
     for (int r=0; r<nRanks; r++)
@@ -290,6 +295,7 @@ ncclResult_t ncclGather_impl(const void* sendbuff, void* recvbuff, size_t sendco
   }
   NCCLCHECK(ncclSend(sendbuff, sendcount, datatype, root, comm, stream));
   NCCLCHECK(ncclGroupEnd());
+  if (!mscclIsCaller()) Recorder::instance().skip(false);
   return ncclSuccess;
 }
 
@@ -372,6 +378,7 @@ ncclResult_t ncclScatter_impl(const void* sendbuff, void* recvbuff, size_t recvc
   if (recvcount == 0) return ncclSuccess;
   int rank;
   NCCLCHECK(ncclCommUserRank(comm, &rank));
+  if (!mscclIsCaller()) Recorder::instance().skip(true);
   NCCLCHECK(ncclGroupStart());
   if (rank == root) {
     for (int r=0; r<nRanks; r++)
@@ -379,6 +386,7 @@ ncclResult_t ncclScatter_impl(const void* sendbuff, void* recvbuff, size_t recvc
   }
   NCCLCHECK(ncclRecv(recvbuff, recvcount, datatype, root, comm, stream));
   NCCLCHECK(ncclGroupEnd());
+  if (!mscclIsCaller()) Recorder::instance().skip(false);
   return ncclSuccess;
 }
 
--- a/src/group.cc
+++ b/src/group.cc
@@ -98,7 +98,10 @@ ncclResult_t ncclAsyncJobComplete(struct ncclAsyncJob* job) {
 
 NCCL_API(ncclResult_t, ncclGroupStart);
 ncclResult_t ncclGroupStart_impl() {
-  NCCLCHECK(Recorder::instance().record(rrGroupStart, ncclGroupDepth));
+  if (!mscclIsCaller())
+  {
+    NCCLCHECK(Recorder::instance().record(rrGroupStart, ncclGroupDepth));
+  }
   ncclResult_t ret = ncclSuccess;
   NVTX3_FUNC_RANGE_IN(nccl_domain);
 
@@ -117,7 +120,10 @@ ncclResult_t ncclGroupStartInternal() {
 
 NCCL_API(ncclResult_t, ncclGroupEnd);
 ncclResult_t ncclGroupEnd_impl() {
-  NCCLCHECK(Recorder::instance().record(rrGroupEnd, ncclGroupDepth));
+  if (!mscclIsCaller())
+  {
+    NCCLCHECK(Recorder::instance().record(rrGroupEnd, ncclGroupDepth));
+  }
   ncclResult_t ret = ncclSuccess;
   NVTX3_FUNC_RANGE_IN(nccl_domain);
   NCCLCHECKGOTO(ncclGroupEndInternal(), ret, exit);
@@ -128,7 +134,10 @@ exit:
 
 NCCL_API(ncclResult_t, ncclGroupSimulateEnd, ncclSimInfo_t* simInfo);
 ncclResult_t ncclGroupSimulateEnd(ncclSimInfo_t* simInfo) {
-  Recorder::instance().record(ncclGroupDepth, simInfo);
+  if (!mscclIsCaller())
+  {
+    Recorder::instance().record(ncclGroupDepth, simInfo);
+  }
   ncclResult_t ret = ncclSuccess;
   NVTX3_FUNC_RANGE_IN(nccl_domain);
   NCCLCHECKGOTO(ncclGroupEndInternal(simInfo), ret, exit);
--- a/src/include/recorder.h
+++ b/src/include/recorder.h
@@ -86,9 +86,13 @@ struct rcclApiCall {
   double                timestamp = -1;
   unsigned long long    graphID = 0;
   int                   graphCaptured = -1;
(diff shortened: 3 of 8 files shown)


src/misc/recorder.cc
```diff
@@ -527,6 +547,7 @@ void Recorder::record(rcclCall_t type, int size, int rank, ncclUniqueId* commId,
     outputFile.write(buffer, len);
     outputFile.flush();
   }
+  // else flush
```
corey-derochie-amd: Is this correct?

src/misc/recorder.cc
```diff
@@ -538,16 +559,22 @@ void Recorder::record(ncclComm_t* comms, int ndev, const int* devlist)
...
+    {
+      outputFile << ", devlist : [";
+      for (int i = 0; i < call.root - 1; i++)
+        outputFile << devlist[i] << ", ";
+      outputFile << devlist[call.root - 1] << "]";
+      outputFile.flush();
+    } else {
+      outputFile.write((char*)devlist, sizeof(int) * ndev);
+      outputFile.flush();
+    }
```
corey-derochie-amd: Is this correct?
```suggestion
    } else {
      outputFile.write((char*)devlist, sizeof(int) * ndev);
    }
    outputFile.flush();
```

test/_RecorderTests.cpp
```diff
@@ -38,16 +38,17 @@ namespace RcclUnitTesting
...
+    std::string filename = "/tmp/test." + std::to_string(pid) + "." + std::string(entry) + ".json";
+    std::ifstream fp(filename);
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     parseJsonEntry(entry, calls);
     int result = memcmp((char*)&calls[0]+4, (char*)&call+4, sizeof(rccl::rcclApiCall)-4);
     fp.close(); // care that recorder is not designed to anticipate fp closing before destructor
-    remove(filename.c_str());
+    //remove(filename.c_str());
```
corey-derochie-amd: Suggest removing commented lines.
```suggestion
```

test/_RecorderTests.cpp
```diff
@@ -38,16 +38,17 @@ namespace RcclUnitTesting
...
     //parse the outfile
-    std::string filename = "test" + std::to_string(pid) + ".json";
-    std::ifstream fp("/tmp/test" + std::to_string(pid) + ".json");
+    std::string filename = "/tmp/test." + std::to_string(pid) + "." + std::string(entry) + ".json";
+    std::ifstream fp(filename);
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     parseJsonEntry(entry, calls);
     int result = memcmp((char*)&calls[0]+4, (char*)&call+4, sizeof(rccl::rcclApiCall)-4);
```
corey-derochie-amd: Seems you are trying to skip comparison of the `pid` field? Writing in this way is hard to understand and is brittle against changes to the struct. First suggestion, just copy the `pid` field from `calls[0]` into `call` and directly compare.
```suggestion
    call.pid = calls[0].pid; // pid will be different, do not compare original values
    int result = memcmp(&calls[0], &call, sizeof(rccl::rcclApiCall));
```

corey-derochie-amd: Alternative to make it a little easier to comprehend.
```suggestion
    // compare only the fields after the pid field
    int result = memcmp(&(calls[0].pid)+1, &(call.pid)+1, sizeof(rccl::rcclApiCall)-sizeof(call.pid));
```

src/misc/recorder.cc
```diff
@@ -501,6 +520,7 @@ void Recorder::record(int groupDepth, ncclSimInfo_t *siminfo)
                        siminfo->size, siminfo->magic, siminfo->version, siminfo->estimatedTime, call.timestamp);
     outputFile.write(buffer, len);
   } // no tid for groupCall
+  // else flush
```
corey-derochie-amd: Not an "else" case?
```suggestion
```

src/misc/recorder.cc
corey-derochie-amd: Why are you commenting "else flush" here?

src/collectives.cc
```diff
@@ -372,13 +378,15 @@ ncclResult_t ncclScatter_impl(const void* sendbuff, void* recvbuff, size_t recvc
...
   NCCLCHECK(ncclCommUserRank(comm, &rank));
+  if (!mscclIsCaller()) Recorder::instance().skip(true);
   NCCLCHECK(ncclGroupStart());
   if (rank == root) {
     for (int r=0; r<nRanks; r++)
       NCCLCHECK(ncclSend(((char*)sendbuff)+r*rankOffset, recvcount, datatype, r, comm, stream));
   }
   NCCLCHECK(ncclRecv(recvbuff, recvcount, datatype, root, comm, stream));
   NCCLCHECK(ncclGroupEnd());
+  if (!mscclIsCaller()) Recorder::instance().skip(false);
```
corey-derochie-amd, line 391: Can `mscclIsCaller()` be true here? I think the function should have returned above if so, so will this ever skip?

src/group.cc
```diff
@@ -98,7 +98,10 @@ ncclResult_t ncclAsyncJobComplete(struct ncclAsyncJob* job) {
 
 NCCL_API(ncclResult_t, ncclGroupStart);
 ncclResult_t ncclGroupStart_impl() {
-  NCCLCHECK(Recorder::instance().record(rrGroupStart, ncclGroupDepth));
+  if (!mscclIsCaller())
+  {
+    NCCLCHECK(Recorder::instance().record(rrGroupStart, ncclGroupDepth));
+  }
```
corey-derochie-amd, line 104: Could you call `skip(true)` whenever you enter an MSCCL call, and then you wouldn't need this check duplicated everywhere?

test/_RecorderTests.cpp
AtlantaPepsi: I would need this line eventually. Currently the file is in /tmp, and we lack the mechanism to close fp before destruction of the process, and removing the file like such might create problems for still alive `Recorder` instance. I'll implement a recorder API that reads envvar on demand and open/close fp accordingly in future

src/misc/recorder.cc
AtlantaPepsi: This is also a space holder for future todo, atm siminfo (and other irregular NCCL structs) are not exported in binary log nor replayed

corey-derochie-amd: Fair enough, might want to add "todo" into the comment.

test/_RecorderTests.cpp
corey-derochie-amd: Might want to add "todo" to this commented line then.

I have an idea for how to handle the envvar issue in `Recorder`, and make the tests a little easier to write. Don't reference the envvar in `Recorder`, just make it a settable argument in the constructor. Then only read the envvar in the singleton function and pass it into the constructor. Tests should construct their own `Recorder` and not use the singleton.

src/collectives.cc
AtlantaPepsi: I think it can, if MSCCL is deemed available and enqueued first time when the collective is called, but later inside MSCCL lifecycle it's decided that the kernel cannot be executed, and it falls back to the same NCCL call but with `mscclIsCaller`. In that case, the second (fallback) collective should not be recorded, but also the send/recv under the hood also be skipped

src/group.cc
AtlantaPepsi: thats a good idea, ill make add this change

corey-derochie-amd: You may want to consider making `skip` a reference count stack instead of a bool, in case the code enters a skip region multiple times.

corey-derochie-amd: We talked and decided to leave as is for now.

corey-derochie-amd (approved):

tools/RcclReplayer/Makefile
```diff
@@ -6,7 +6,7 @@ INCLUDES = -I$(MPI_DIR)/include -I$(RCCL_DIR)/include -I$(RCCL_DIR)/hipify/src/i
 LDFLAGS  = -L$(MPI_DIR)/lib -L$(RCCL_DIR) -lmpi -lrccl
 
 main: rcclReplayer.cpp
-	$(ROCM_DIR)/bin/hipcc rcclReplayer.cpp -O1 -g -o rcclReplayer $(INCLUDES) $(LDFLAGS)
+	$(ROCM_DIR)/bin/hipcc -std=c++20 rcclReplayer.cpp -O0 -g -o rcclReplayer $(INCLUDES) $(LDFLAGS)
```
alex-breslow-amd, line 9: Typically, we use -O1 -g for debug builds of RCCL.  I'm wondering if the same makes sense here.  However, -O0 should be fine I guess.

gilbertlee-amd (approved):

test/_RecorderTests.cpp
```diff
@@ -38,16 +38,18 @@ namespace RcclUnitTesting
...
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     fp.getline(entry, 4096);
     parseJsonEntry(entry, calls);
-    int result = memcmp((char*)&calls[0]+4, (char*)&call+4, sizeof(rccl::rcclApiCall)-4);
+    // compare only the fields after the pid field
+    int result = memcmp(&(calls[0].pid)+1, &(call.pid)+1, sizeof(rccl::rcclApiCall)-sizeof(call.pid));
     fp.close(); // care that recorder is not designed to anticipate fp closing before destructor
-    remove(filename.c_str());
+    //remove(filename.c_str()); TODO: after implement fp closing mechanism via envvar, remove file after test case
```
alex-breslow-amd, line 54: Can you add TODO(NTID)?  Or an associated Jira ticket?  We should be tracking commented code that needs to be fixed.

src/misc/recorder.cc
```diff
@@ -538,15 +559,20 @@ void Recorder::record(ncclComm_t* comms, int ndev, const int* devlist)
...
-      outputFile << devlist[i] << ", ";
-    outputFile << devlist[call.root - 1] << "]";
+    if (output_json)
+    {
+      outputFile << ", devlist : [";
+      for (int i = 0; i < call.root - 1; i++)
+        outputFile << devlist[i] << ", ";
+      outputFile << devlist[call.root - 1] << "]";
+    } else {
+      outputFile.write((char*)devlist, sizeof(int) * ndev);
```
alex-breslow-amd, line 576: Prefer static_cast to c-style casting.

src/misc/recorder.cc
```diff
@@ -108,6 +115,14 @@ Recorder& Recorder::instance()
   return _instance;
 }
 
+void Recorder::skip(bool b)
```
alex-breslow-amd, line 120: This works, but I think a try/catch block might be cleaner and more robust.  No need to change anything.

src/include/recorder.h
```diff
@@ -88,17 +88,21 @@ struct rcclApiCall {
...
+  size_t                sendPtrExtent = 0;
+  size_t                recvPtrExtent = 0;
   size_t                count = 0;
   ncclDataType_t        datatype;
   ncclRedOp_t           op;
   int                   root = -1;
   int                   nRanks = -1;
-  ncclComm_t            comm;
-  hipStream_t           stream;
+  ncclComm_t            comm = NULL;
```
alex-breslow-amd, line 106: Prefer nullptr over NULL

src/group.cc
```diff
@@ -117,7 +120,10 @@ ncclResult_t ncclGroupStartInternal() {
 
 NCCL_API(ncclResult_t, ncclGroupEnd);
 ncclResult_t ncclGroupEnd_impl() {
-  NCCLCHECK(Recorder::instance().record(rrGroupEnd, ncclGroupDepth));
+  if (!mscclIsCaller())
```
alex-breslow-amd, line 123: So we don't plan on supporting MSCCL?  That's fine.  Just curious.

alex-breslow-amd (approved):
Looks okay to me.

src/group.cc
AtlantaPepsi: Not at the current stage, the main goal is to accurately log what RCCL API users called for replay purpose. MSCCL internally could fall back to the RCCL collective kernels again, we don't want to pollute logs with these underlying calls

src/misc/recorder.cc
AtlantaPepsi: Will clean up code style in future amendments

PR #1731, opened 2025-06-05
[BUILD] Enable LL128 on gfx950

nileshnegi:
## Details

**Work item:** Internal

**What were the changes?**  
Enable LL128 protocol on `gfx950` multi-node.

**How was the outcome achieved?**  
Enable LL128 device-code generation for `gfx950` in `src/device/generate.py`

## Approval Checklist
___Do not approve until these items are satisfied.___
- [x] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #1734, opened 2025-06-06
[DEVICE] Adding ability to choose unroll factor at runtime

nileshnegi:
## Details

**Work item:** Internal

**What were the changes?**  
- Enable runtime selection of unroll factor with new RCCL env. variable: `RCCL_UNROLL_FACTOR`.
- By default, RCCL generates device code with unroll=1, 2, 4.
- Defined presets for unroll depending on GPU target and no. of nodes.
- Adds unroll as a build-time parameter to `ONLY_FUNCS` for debugging.

`gfx950` performs better when we use `unroll=1` for single-node and `unroll=2` for multi-node.

- `src/device/generate.py` was modified to generate function table entries for multiple unrolls.
- `src/init.cc` was modified to define presets for unroll

**Additional Documentation:**  
The compilation and linking time for default builds increases as we generate triple the number of kernels.

## Approval Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 * Added support for `unroll=1` in device-code generation to improve performance.
 * Set a default of 112 channels for a single node with `8 * gfx950`.
 * Enabled LL128 protocol on `gfx950`.
-* Adding ability to choose unroll factor at runtime via `RCCL_UNROLL_FACTOR`.  This can be set at runtime to 1, 2 or 4.  This change currently increases compilation and linking time due to triple the number of kernels generated.
+* Adding ability to choose unroll factor at runtime via `RCCL_UNROLL_FACTOR`.  This can be set at runtime to 1, 2, or 4.  This change currently increases compilation and linking time because it triples the number of kernels generated.
 * Added MSCCL support for AllGather multinode gfx942/gfx950 (i.e., 16 and 32 GPUs). To enable, set the environment variable `RCCL_MSCCL_FORCE_ENABLE=1`. Max message size for MSCCL AllGather usage is `12292 * sizeof(datatype) * nGPUs`.
 * Thread thresholds for LL/LL128 are selected in Tuning Models for the MI300X. This impacts the number of channels used for AG and RS. Channel tuning model is bypassed if `NCCL_THREAD_THRESHOLDS`, `NCCL_MIN_NCHANNELS', or 'NCCL_MAX_NCHANNELS` are set.
 * Multi-node tuning for AllGather, AllReduce, and ReduceScatter that leverages LL/LL64/LL128 protocol to use nontemporal vector load/store for tunable message size ranges.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,10 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 ### Added
 
 * Added new GPU target `gfx950`.
-* Added support for `unroll=1` in device-code generation to improve performance,
-* Set a default of 112 channels for a single node with `8 * gfx950`,
+* Added support for `unroll=1` in device-code generation to improve performance.
+* Set a default of 112 channels for a single node with `8 * gfx950`.
 * Enabled LL128 protocol on `gfx950`.
+* Adding ability to choose unroll factor at runtime via `RCCL_UNROLL_FACTOR`.  This can be set at runtime to 1, 2 or 4.  This change currently increases compilation and linking time due to triple the number of kernels generated.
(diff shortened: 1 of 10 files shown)


nileshnegi:
Currently a draft, as we expect changes after https://github.com/ROCm/rccl/pull/1731 is merged.

src/device/generate.py
```diff
@@ -70,16 +75,27 @@ def paste(sep, *args):
...
+unroll_and_onlyFuncs = sys.argv[5:]
+func_pattern = "AllGather|AllReduce|AllToAllPivot|Broadcast|Reduce|ReduceScatter|SendRecv"
+
+# check if UNROLL and/or ONLY_FUNC are defined
+if unroll_and_onlyFuncs:
+  arg = unroll_and_onlyFuncs[0]
+
+  # Set UNROLL only if digit and part of `all_unroll`
+  if arg.isdigit() and arg in all_unroll:
+    all_unroll = [arg]
```
corey-derochie-amd: Alternative to modifying `all_unroll` is to have unroll as part of the `ONLY_FUNCS` pattern. It is already supported.

src/enqueue.cc
```diff
@@ -28,30 +28,30 @@
...
+  int numKernels = sizeof(rcclKernelTable) / sizeof(rcclKernelTable[0]);
+  int firstKernel = 0;
 #endif
 
+  // Check if the requested unroll exists
+  for (int kernelIdx = 0; kernelIdx < numKernels; kernelIdx++) {
+    if (rcclKernelTable[firstKernel + kernelIdx].unroll == unroll) {
+      return rcclKernelTable[firstKernel + kernelIdx].funcPtr;
+    }
+  }
```
corey-derochie-amd, line 50: This lookup can probably be done in an O(1) way, but it probably doesn't matter?

src/device/generate.py
nileshnegi: trying...

nileshnegi:
PR https://github.com/ROCm/rccl/pull/1731 has been merged, and conflicts resolved, so ready for review.

CHANGELOG.md
```diff
@@ -14,9 +14,10 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 ### Added
 
 * Added new GPU target `gfx950`.
-* Added support for `unroll=1` in device-code generation to improve performance,
-* Set a default of 112 channels for a single node with `8 * gfx950`,
+* Added support for `unroll=1` in device-code generation to improve performance.
+* Set a default of 112 channels for a single node with `8 * gfx950`.
 * Enabled LL128 protocol on `gfx950`.
+* Adding ability to choose unroll factor at runtime via `RCCL_UNROLL_FACTOR`.  This can be set at runtime to 1, 2 or 4.  This change currently increases compilation and linking time due to triple the number of kernels generated.
```
```suggestion
* Adding ability to choose unroll factor at runtime via `RCCL_UNROLL_FACTOR`.  This can be set at runtime to 1, 2, or 4.  This change currently increases compilation and linking time because it triples the number of kernels generated.
```

amd-jnovotny (approved):
Minor rewording suggestions.

CHANGELOG.md
corey-derochie-amd: The Oxford comma...

src/device/common.h
```diff
@@ -572,21 +572,7 @@ __device__ __forceinline__ void ncclKernelMain(struct ncclDevKernelArgs const* a
...
-        ncclDevFuncTable_4[ncclShmem.funcId]();
-#else
-      if (COLL_UNROLL == 1)
-        NCCL_CALL_FUNCTIONS_1(ncclShmem.funcId);
-      else if (COLL_UNROLL == 2)
-        NCCL_CALL_FUNCTIONS_2(ncclShmem.funcId);
-      else
-        NCCL_CALL_FUNCTIONS_4(ncclShmem.funcId);
-#endif
+      NCCL_CALL_FUNCTIONS<COLL_UNROLL>(ncclShmem.funcId);
```
alex-breslow-amd, line 575: I like this.  This is cleaner.

corey-derochie-amd: Does this remove indirect function call capability? Is that intended?

nileshnegi: no, this IFC logic is moved to `src/device/generate.py` https://github.com/ROCm/rccl/pull/1734/files#diff-515d9cbe4978a14447d8387010a20f6aada1352d180a7b6cc30d0c37a6cc491fR332-R340

nileshnegi: do you mean the overloaded call for IFC?

alex-breslow-amd:
I guess a general comment I had (preexists your diff) is whether we should try to remove actual lines of code from generate.py (i.e., stuff that we pipe to out()) and instead just gate functionality with ```#if defined``` statements so that all C++ code exists in C++ files and the generator just switches different stuff on by ```#define```'ing different macros.  I think this might make long-term maintainability easier (not 100% sure since code riddled with #define statements can be a pain to read too).  I'm not expecting you to make this change in this diff, but I think it's something we should discuss as a team.

src/device/common.h
corey-derochie-amd: I missed that, generate.py is hard to read.

corey-derochie-amd (approved):

src/enqueue.cc
nileshnegi: keeping as-is for now.

PR #1752, opened 2025-06-19
passing down NET_OPTIONAL_RECV_COMPLETION hint to n/w plugin to enable optimizations

sarat-k:
n/w plugin can decide which RDMA operation to use to send data to the peer based on this hint. using WRITEs is more efficient in completions are not needed on the other side.

### DIFF
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -1360,8 +1360,14 @@ static ncclResult_t sendProxyProgress(struct ncclProxyState* proxyState, struct 
             // Coverity complains about the size here as pointing to an out-of-scope temporary.  Which is nonsense,
             // since size is a plain integer.
             // coverity[use_invalid:FALSE]
-            NCCLCHECK(proxyState->ncclNet->isend(resources->netSendComm, buff, size, resources->tpRank, sub->sendMhandle, sub, sub->requests+buffSlot));
-            if (sub->requests[buffSlot] != NULL) {
+            void **requestPtr = sub->requests+buffSlot;
+            // for LL/LL128 protocols, completion event for write operation is not needed on the receiver side as
+            // the LL flags are actively polled to detect if full data is received or not, so this hint can be used
+            // by network plugin to optimize the transport for LL/LL128
+            bool ignoreCompletion = ncclParamNetOptionalRecvCompletion() && ((args->protocol == NCCL_PROTO_LL128) || (args->protocol == NCCL_PROTO_LL));
+            if (ignoreCompletion) *requestPtr = (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION;
+            NCCLCHECK(proxyState->ncclNet->isend(resources->netSendComm, buff, size, resources->tpRank, sub->sendMhandle, sub, requestPtr));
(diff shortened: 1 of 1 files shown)


src/transport/net.cc
```diff
@@ -1326,8 +1326,11 @@ static ncclResult_t sendProxyProgress(struct ncclProxyState* proxyState, struct
             // Coverity complains about the size here as pointing to an out-of-scope temporary.  Which is nonsense,
             // since size is a plain integer.
             // coverity[use_invalid:FALSE]
-            NCCLCHECK(proxyState->ncclNet->isend(resources->netSendComm, buff, size, resources->tpRank, sub->sendMhandle, sub->requests+buffSlot));
-            if (sub->requests[buffSlot] != NULL) {
+            void **requestPtr = sub->requests+buffSlot;
```
nileshnegi, line 1363: Please add some comments here, so that it helps for future syncs.

sarat-k: updated with the comments

vijasrin (approved):
lgtm

src/transport/net.cc
```diff
@@ -1360,8 +1360,14 @@ static ncclResult_t sendProxyProgress(struct ncclProxyState* proxyState, struct
             // Coverity complains about the size here as pointing to an out-of-scope temporary.  Which is nonsense,
             // since size is a plain integer.
             // coverity[use_invalid:FALSE]
-            NCCLCHECK(proxyState->ncclNet->isend(resources->netSendComm, buff, size, resources->tpRank, sub->sendMhandle, sub, sub->requests+buffSlot));
-            if (sub->requests[buffSlot] != NULL) {
+            void **requestPtr = sub->requests+buffSlot;
+            // for LL/LL128 protocols, completion event for write operation is not needed on the receiver side as
+            // the LL flags are actively polled to detect if full data is received or not, so this hint can be used
```
mustafabar, line 1365: For multi-node, how do you know that the RDMA write is completed without some completion event? The polled flag is set by the CPU proxy thread. 
CC: @nileshnegi

sarat-k: are you asking about data sender side or receiver side ? sender side there is always completion event posted by the driver to the app .. receiver side depends on whether WRITE or WRITE_IMM is used by RDMA

mustafabar: I understand the motivation now and makes sense for LL/LL128. However, is this targeting some other plugin implementation besides net_ib.c? The isend implementation ignores the input contents of that pointer and overwrite it with fifoReqs. If your plugin never checks the incoming value, the hint is lost. https://github.com/ROCm/rccl/blob/c28d3d26a3e64b4ca1a8858ae5d9e4f16c1ad7db/src/transport/net_ib.cc#L2159
Can you pls verify if the actual completion is skipped? I could have missed something

sarat-k: Yes, this just a hint to the plugin and plugin can choose to ignore if there is no benefit to specific NIC's datapath. AMD's AINIC is programmable and the corresponding ANP plugin makes use of this for better performance

mustafabar (approved):

PR #1756, opened 2025-06-20
[DEVICE] Enable PAT algo for RCCL 1ppn

BertanDogancay:
## Details

**What were the changes?**  
Initial enablement of PAT algo for RCCL.
This PR does not cover tuning for PAT algo. There will be a separate PR for tuning work.

PAT algo came with NCCL 2.23 but was disabled this whole time. We're now enabling it since there are use cases for it.

**Additional Documentation:**  
Use `NCCL_PAT_ENABLE=1` to enable it in runtime.

NOTE: This addition of new PAT algo will increase RCCL build time by ~20% since we're generating more device functions now.

## Approval Checklist


src/include/collectives.h
```diff
@@ -402,7 +402,7 @@ class RingBCAlgorithm : public RingAlgorithm {
 #endif
 
 // Need a power of two to ensure it divides by parallelFactor (which is also a power of two)
-#define NCCL_PAT_NWORKERS 512
+#define NCCL_PAT_NWORKERS 128
```
nusislam, line 405: Why was this changed?

src/device/all_gather.h
```diff
@@ -220,7 +220,6 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_RING, NCCL_PROTO_LL128
 template<typename T, typename RedOp>
 struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE> {
   __device__ __forceinline__ void run(int tid, int nthreads, struct ncclDevWorkColl* work) {
-#if __CUDA_ARCH__ >= 600
```
corey-derochie-amd, line 223: NCCL has this restriction because it probably relies on certain CUDA functionality. Do we have an equivalent HIP version we should check against?

src/include/collectives.h
```diff
@@ -424,7 +424,7 @@ struct ncclPatPeer {
     int connStepSize;
 };
 
-#define NCCL_SHMEM_PAT_STEPS 32
+#define NCCL_SHMEM_PAT_STEPS 64
```
nusislam: What is the reason of this change?

src/device/all_gather.h
```diff
@@ -242,8 +241,8 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
       int step = 0;
       while (1) {
         struct ncclPatStep* ps = shmem->patSteps+(step%NCCL_SHMEM_PAT_STEPS);
-        cuda::atomic_ref<int, cuda::thread_scope_block> poll(ps->flags);
```
corey-derochie-amd, line 245: No `atomic_ref` in HIP?

src/graph/tuning.cc
```diff
@@ -141,18 +141,18 @@ struct tuningModel {
 static struct tuningModel tuning_model_0 {
   .hwLat = {
     /* NVLINK */
-    { /* Tree (LL/LL128/Simple)*/ { 0.8, 1.4, 2.5 }, /* Ring (LL/LL128/Simple)*/ { 0.8, 2.2, 3.6 }, /* CollNetDirect (Simple)*/ { 0.0, 0.0, 0.8 }, /* CollNetChain (Simple)*/ { 0.0, 0.0, 1.4 }, /* NVLS */ { 0, 0, 0 }, /* NVLS Tree */ { 0, 0, 0 } },
+    { /* Tree (LL/LL128/Simple)*/ { 0.8, 1.4, 2.5 }, /* Ring (LL/LL128/Simple)*/ { 0.8, 2.2, 3.6 }, /* CollNetDirect (Simple)*/ { 0.0, 0.0, 0.8 }, /* CollNetChain (Simple)*/ { 0.0, 0.0, 1.4 }, /* NVLS */ { 0, 0, 0 }, /* NVLS Tree */ { 0, 0, 0 }, /* PAT */ { 0, 0, 3.6} },
     /* PCI */
-    { /* Tree (LL/LL128/Simple)*/ { 2.2, 2.2, 5.7 }, /* Ring (LL/LL128/Simple)*/ { 2.2, 2.2, 5.7 }, /* CollNetDirect (Simple)*/ { 0.0, 0.0, 5.7 }, /* CollNetChain (Simple)*/ { 0.0, 0.0, 5.7 }, /* NVLS */ { 0, 0, 0 }, /* NVLS Tree */ { 0, 0, 0 } },
+    { /* Tree (LL/LL128/Simple)*/ { 2.2, 2.2, 5.7 }, /* Ring (LL/LL128/Simple)*/ { 2.2, 2.2, 5.7 }, /* CollNetDirect (Simple)*/ { 0.0, 0.0, 5.7 }, /* CollNetChain (Simple)*/ { 0.0, 0.0, 5.7 }, /* NVLS */ { 0, 0, 0 }, /* NVLS Tree */ { 0, 0, 0 }, /* PAT */ { 0, 0, 5.7} },
```
nusislam, line 147: How did you select these values for PAT?

src/device/all_gather.h
BertanDogancay: I checked but couldn't find an equivalent API from HIP for this.

src/include/collectives.h
BertanDogancay: RCCL has `MAX_NTHREADS=256` so any number above that won't work. I didn't set it to 256 because I experienced some hangs with it. Still investigating.

src/graph/tuning.cc
BertanDogancay: All tuning models except `tuning_model_0` is set to 0 for PAT algo since RCCL chooses this tuning model by default when it can't match it to anything (1 rank/node in this case). PAT also only supports SIMPLE protocol so values for LL and LL128 are set to 0 as well.  And we only care about second index for `bwRatio` since PAT needs multiple nodes.

The `non-zero` values don't have actual meanings they were just copied from RING algo for initial enablement purposes. I think there is a script to select these values which needs to be updated for PAT.

src/device/all_gather.h
```diff
@@ -242,8 +241,8 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
...
-        while (poll.load(cuda::memory_order_acquire) != 0) pollCount++; // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
+        int* poll = &ps->flags;
+        while (__atomic_load_n(poll, __ATOMIC_ACQUIRE) != 0) pollCount++; // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
```
wenkaidu: Looks like the poll is "block" scope. I suspect block scope atomic can work? i.e. __hip_atomic_load(poll, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_WORKGROUP)

src/include/device.h
```diff
@@ -693,48 +693,61 @@ extern int const ncclDevFuncRowToId[];
...
+    int nAlgos = 2;
     if (coll == ncclFuncAllGather) {
-      row += proto;
+      int algo1 = algo == NCCL_ALGO_RING ? 0 :
+                /*algo == NCCL_ALGO_PAT*/ 1;
+      row += algo1 * NCCL_NUM_PROTOCOLS + proto;
       break;
     }
-    row += NCCL_NUM_PROTOCOLS;
+    row += nAlgos * NCCL_NUM_PROTOCOLS;
```
mustafabar, line 704: Are all protocols supported for PAT?

src/graph/tuning.cc
```diff
@@ -379,10 +379,12 @@ static const double perChMaxTreeBws[][3] = {
...
 
-NCCL_PARAM(PatEnable, "PAT_ENABLE", 2);
+NCCL_PARAM(PatEnable, "PAT_ENABLE", 0);
 static int ncclPatEnable(struct ncclComm* comm) {
   int patEnable = ncclParamPatEnable();
+#if !defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__)
   if (comm->minCompCap < 60) return 0; // Need SM60 or higher for CUDA atomics
+#endif
   if (patEnable != 2) return patEnable;
   if (comm->nNodes != comm->nRanks) return 0; // PAT only supports 1 GPU per node
```
mustafabar: Looks like setting it to 1 will force-enable it. 
If PAT is only supported for 1 GPU per node, we should not force-enable it in that scenario. 
Pls modify this part accordingly.

mustafabar (commented):
Added some reviews below

src/include/device.h
BertanDogancay: No, only `SIMPLE` protocol is supported at the moment for `PAT` algo, but we generate the kernels for `LL` and `LL128` during build regardless. NCCL does this the same way. I believe this is for code simplicity. Otherwise, `ncclDevFuncId()` will be messy.

Currently proto support for PAT is handled in runtime.

nusislam:
@BertanDogancay - Is there any message size threshold for PAT enablement?

BertanDogancay:
> @BertanDogancay - Is there any message size threshold for PAT enablement?

PAT is used for small to medium message sizes. It switches to RING algo after 2M for both `AllGather` and `ReduceScatter`.  I think there is no hardcoded threshold value, this is decided in tuning.cc

src/device/all_gather.h
BertanDogancay: Implemented Wenkai's suggestion.

src/include/collectives.h
BertanDogancay: The computation loop will never run when setting this to 256 which will cause a hang. PAT requires a pow2 number of workers, so 128 should be correct.

src/include/collectives.h
BertanDogancay: This was to make it safer and give more time to the worker threads before the steps are overwritten by the compute threads. I reverted this back to the original value after finding out that it's a synchronization issue.

src/device/prims_simple.h
```diff
@@ -92,7 +92,7 @@ class Primitives<
   }
 
   inline __device__ void patBarrier() {
-    barrier();
+    __syncthreads();
```
nusislam: Is this change ok in terms of correctness?

nusislam:
Please change the PR description also..since the force enable option has been removed.

src/device/prims_simple.h
BertanDogancay: no it wasn't. I added new barrier implementation for PAT now.

wenkaidu (approved):

src/device/all_gather.h
BertanDogancay: I couldn't find since when hip atomic operation APIs exist but they're old. We should be okay without the version check.

BertanDogancay: The same API is used in `primitives.h` as well and we don't have version check for that too.

src/graph/tuning.cc
```diff
@@ -421,11 +421,12 @@ static const double perChMaxTreeBws[][3] = {
   /* Blackwell (N1/N2/N4) */ {2*38.7, 2*41.4, 2*36.0},
 };
 
-NCCL_PARAM(PatEnable, "PAT_ENABLE", 2);
+NCCL_PARAM(PatEnable, "PAT_ENABLE", 0);
 static int ncclPatEnable(struct ncclComm* comm) {
-  int patEnable = ncclParamPatEnable();
+  if (!ncclParamPatEnable()) return 0;
```
corey-derochie-amd, line 426: Other than disabling by default, why change how this flag is interpreted?

corey-derochie-amd, line 426: In the original way, the user could either explicitly disable by setting to 0, explicitly enable by setting to 1, or allow it to selectively disable based on rank count and `netDeviceType` by setting it to 2. We should maintain this behaviour.

```suggestion
  int patEnable = ncclParamPatEnable();
```

src/graph/tuning.cc
```diff
@@ -421,11 +421,12 @@ static const double perChMaxTreeBws[][3] = {
...
+#if !defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__)
   if (comm->minCompCap < 60) return 0; // Need SM60 or higher for CUDA atomics
-  if (patEnable != 2) return patEnable;
+#endif
```
corey-derochie-amd, line 429:
```suggestion
#endif
  if (patEnable != 2) return patEnable;
```

corey-derochie-amd (changes requested):

src/graph/tuning.cc
BertanDogancay: The reason we removed the original force enable logic here is because with `NCCL_PAT_ENABLE=1` it will still use `RING` algo when > 1ppn, and if we force `NCCL_ALGO=PAT` with it, we get `no algo/proto available for function` error. Since it defaults to ring and PAT is currently only supported for 1ppn, it is better to remove this.

corey-derochie-amd: How does this work in NCCL?

BertanDogancay: This force enable flag is only to experiment with > 1ppn I assume. We can bring back this when we tune RCCL to pick up PAT algo when > 1ppn.

BertanDogancay: Right now, there is no use case for this force enable flag and current implementation should be fine.

corey-derochie-amd (approved):

PR #1772, opened 2025-06-25
add warning if workFIFO is not available after multiple retries.

thananon:
## Details
Follow up from #1722. We are still not out of the woods for high iteration alltoall. We should not just hang when work FIFO exhaustion occur. This commit prints out warning when RCCL get stuck waiting. 

image

**What were the changes?**  
Print out warnings after multiple retries to get workFifo.

Provide user warnings and mitigation.

Count and print.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/enqueue.cc
+++ b/src/enqueue.cc
@@ -1187,6 +1187,8 @@ static ncclResult_t scheduleP2pTasksToPlan(
 // Spin until its safe to increase comm->workFifoProduced to desiredProduced.
 static void waitWorkFifoAvailable(struct ncclComm* comm, uint32_t desiredProduced) {
   bool hasRoom = (desiredProduced - comm->workFifoConsumedLeast) <= comm->workFifoBytes;
+  uint64_t count = 0;
+  int warned = 0;
   if (hasRoom) return;
   while (true) {
     // We have to poll for notifications from device.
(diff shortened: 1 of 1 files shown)


src/enqueue.cc
```diff
@@ -1225,6 +1227,17 @@ static void waitWorkFifoAvailable(struct ncclComm* comm, uint32_t desiredProduce
...
     sched_yield();
+
+    /* Warn if we get stuck waiting for workFifo. */
+    count++;
+    if (warned == 0 && count == 100000 && comm->rank == 0) {
+      warned = 1;
+      WARN("Waiting for work FIFO to become available."
+           "Work fifo exhaustion can happen in large scale/high iteration count of alltoall."
+           "In order to increase work FIFO size, set NCCL_WORK_FIFO_BYTES to higher number"
+           " (current: %ld). \n\nRCCL continues to retry...", comm->workFifoBytes);
```
corey-derochie-amd:
```suggestion
      WARN("Waiting for work FIFO to become available. "
           "Work fifo exhaustion can happen in large scale/high iteration count of alltoall. "
           "In order to increase work FIFO size, set NCCL_WORK_FIFO_BYTES to higher number (current: %ld).\n\n"
           "RCCL continues to retry...", comm->workFifoBytes);
```
:)

src/enqueue.cc
```diff
@@ -1225,6 +1227,17 @@ static void waitWorkFifoAvailable(struct ncclComm* comm, uint32_t desiredProduce
     hasRoom = (desiredProduced - comm->workFifoConsumedLeast) <= comm->workFifoBytes;
     if (hasRoom) break;
     sched_yield();
+
+    /* Warn if we get stuck waiting for workFifo. */
+    count++;
+    if (warned == 0 && count == 100000 && comm->rank == 0) {
```
corey-derochie-amd, line 1233: Probably not important, but should this `count` limit be configurable?

src/enqueue.cc
corey-derochie-amd: Or did you intend for each line to be a separate line in the output?

corey-derochie-amd (approved):

src/enqueue.cc
thananon: I was playing around with the numbers but this should show up as soon as we think we are stuck. I think it should be fine as is but adding an env is also fine.

BertanDogancay (approved):

PR #1774, opened 2025-06-26
Code coverage improvements - src/transport/p2p.cc

atulkulk:
Added new unit tests to cover test cases that were not covered by the functional tests for src/transport/p2p.cc file.

**What were the changes?**  
Addition of new unit tests

New unit tests were written

| **p2p.cc** | Initial | 74.19% (23/31)    | 53.16% (387/728) | 36.12% (522/1445) | 26.37% (202/766) |
|                 | Updated | 87.10% (27/31)    | 67.44% (493/731) | 46.40% (689/1485) | 34.66% (269/776) |

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -732,10 +732,40 @@ if (BUILD_TESTS)
     set(TEST_NONSTATIC_SOURCE_FILES
       ${HIPIFY_SRC_DIR}/misc/alt_rsmi.cc
       ${HIPIFY_SRC_DIR}/transport/shm.cc
+      ${HIPIFY_SRC_DIR}/transport/p2p.cc
     )
+
+    set(EXCLUDE_STATIC_FILE "${CMAKE_SOURCE_DIR}/tools/scripts/exclude_static_list.txt")
+    # Read the exclude list file into a CMake variable
+    file(READ "${EXCLUDE_STATIC_FILE}" EXCLUDE_STATIC_CONTENTS)
+    string(REPLACE "\n" ";" EXCLUDE_STATIC_LINES "${EXCLUDE_STATIC_CONTENTS}")
+    # Create a mapping from full/relative filename to exclude list
+    unset(EXCLUDE_MAP)
+    foreach(line ${EXCLUDE_STATIC_LINES})
+      if(line MATCHES "^([a-zA-Z0-9_./-]+):([a-zA-Z0-9_,]*)")
+        set(fname "${CMAKE_MATCH_1}")
+        set(exlist "${CMAKE_MATCH_2}")
+        # Map both the basename and the full/relative path for flexibility
+        get_filename_component(basename "${fname}" NAME)
+        set(EXCLUDE_MAP_${fname} "${exlist}")
+        set(EXCLUDE_MAP_${basename} "${exlist}")
+      endif()
+    endforeach()
+
+    # Now, for each file, get the exclude list and pass to the script
     # Create a custom command to backup the original files and remove static
     # Always run replace script on hipified files, but preserve original backups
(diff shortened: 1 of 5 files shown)


src/transport/p2p.cc
```diff
@@ -1176,4 +1176,5 @@ static void initCeOperation() {
     }
     init = 1;
   }
-}
\ No newline at end of file
+}
+
```
thananon: Is this intentional?

test/common/StandaloneUtils.hpp
```diff
@@ -25,7 +25,7 @@
     }                                                           \
 } while(0)
 
-#define MAX_STACK_SIZE 570
+#define MAX_STACK_SIZE 640
```
thananon: Do we have larger stacksize now?

atulkulk: Yes, it came out to be around `584` while I was working on the tests. Added additional buffer (`640`) for future increase (just in case).

src/transport/p2p.cc
atulkulk: No, I will revert this change.

thananon (approved):

venksubramd:
Must like in PR 1689, the test file here is duplicating the structs in the .cc file. I would like to discuss this further and find ways to avoid this. Once this becomes a practice, besides violating the DRY principle, we will have to put efforts to keep the structs in sync and any resulting code changes. It would be highly beneficial not to duplicate the structs and focus on testing the functions that are public and visible, using only data structures those functions need.

We should also avoid redeclaring the functions that are present in the p2p.h. I am sure some dependency or linking frustration may have lead to this, but we should put effort to resolve those issues instead of duplicating the declarations. If we go down this path we are really making the test suite brittle for future changes. In the best case we would get errors during compilation, but in the worst case we may have runtime issues that are not easy to detect, due to incompatibility to change in the data structure, and will become a time sink.

We should focus on making the tests for functions declared in p2p.h more robust.

thananon (approved):

venksubramd:
TEST_F(P2pTests, P2pAllocateShareableBuffer)
Please avoid multiple independent tests in one test. If one of the independent tests fail, what follows will not get a chance to run. Unless the asserts are related to results from one call, the asserts have to be in separate tests.

TEST_F(P2pTests, P2pFreeShareableBuffer) {
  // Test 1: All valid parameters - should succeed
The function does not care about the parameters, it succeeds always. 🙂

Do we need tests for 
ncclP2pImportShareableBuffer
ncclIpcLocalRegisterBuffer
ncclIpcGraphRegisterBuffer
ncclIpcDeregBuffer
?

As a separate task we need to revisit the issue of duplicating the structures from code into tests.

atulkulk:
> TEST_F(P2pTests, P2pAllocateShareableBuffer) Please avoid multiple independent tests in one test. If one of the independent tests fail, what follows will not get a chance to run. Unless the asserts are related to results from one call, the asserts have to be in separate tests.

Addressed above issue.

> TEST_F(P2pTests, P2pFreeShareableBuffer) { // Test 1: All valid parameters - should succeed The function does not care about the parameters, it succeeds always. 🙂
> 
> Do we need tests for ncclP2pImportShareableBuffer ncclIpcLocalRegisterBuffer ncclIpcGraphRegisterBuffer ncclIpcDeregBuffer ?

I will need to look into these but some of them are already covered with the existing tests. I will create a separate PR if any new tests can be added for the above functions.

> As a separate task we need to revisit the issue of duplicating the structures from code into tests.

rahulvaidya20 (approved):

PR #1780, opened 2025-06-28
[INIT] Fix fallback for unsupported user-specified runtime unroll factor

nileshnegi:
## Details

**Work item:**
Internal (SWDEV-540559)

**What were the changes?**  
- Fix the fallback mechanism when the user specifies an unsupported runtime unroll factor.
- Move `commSetUnrollFactor()` to `src/rccl_wrap.cc`

Currently, if the user specifies an unsupported unroll factor, RCCL would default to unroll=1. This is not necessarily the optimal choice for all GPU targets (e.g., `gfx942`, where the recommended unroll value is 2).

Expose `rcclKernelTable` to check for supported unroll factors during initialization.

## Approval Checklist

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -94,27 +94,33 @@ ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
   hipDeviceProp_t devProp;
   CUDACHECK(hipGetDeviceProperties(&devProp, comm->cudaDev));
 
-  //If RCCL runtime param is set, it will override defaults
+  //If RCCL runtime param is set, it will override defaults, if supported
   if (rcclParamUnrollFactor() != 0) {
-    comm->unroll = rcclParamUnrollFactor();
-    INFO(NCCL_INIT, "RCCL Unroll Factor (user-defined): %d", comm->unroll);
-  }
-  else {
-    if (IsArchMatch(devProp.gcnArchName, "gfx950")) {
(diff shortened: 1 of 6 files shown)


thananon (approved):

src/init.cc
```diff
@@ -94,27 +94,37 @@ ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
...
+      comm->unroll = 2;
   }
+  else if((IsArchMatch(devProp.gcnArchName, "gfx908")) ||
+          (IsArchMatch(devProp.gcnArchName, "gfx942") && devProp.multiProcessorCount > 80))
+    //on MI300X and gfx908, use unroll=2
+    comm->unroll = 2;
+  else
+    comm->unroll = 4;
+  INFO(NCCL_INIT, "RCCL Unroll Factor (pre-set): %d", comm->unroll);
   return ncclSuccess;
```
gilbertlee-amd: Let's avoid putting this in init.cc.  
We should introduce a new file under misc which captures all this logic, so that people know where it happens.

gilbertlee-amd (changes requested):

src/init.cc
mustafabar: @nileshnegi  - we already have a file src/rccl_wrap.cc with RCCL-only logic. Can this be moved there?

corey-derochie-amd (approved):

gilbertlee-amd (approved):

PR #1783, opened 2025-07-01
Code Coverage Unit Tests for comm.h

ishkool:
## Details

**Work item:** _"Internal", Added new unit tests to cover test cases that were not covered by the functional tests for src/include/comm.h file.

**What were the changes?**  
Addition of new unit tests

**Added 1 test for uncovered function, the functional code coverage is 100% (8/8) from previous 87.5%(7/8)**

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/test/CommTest.cpp
+++ b/test/CommTest.cpp
@@ -1,3 +1,8 @@
+/*************************************************************************
+ * Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
 #include "comm.h"
 #include "TestBed.hpp"
 
--- a/test/CommTest.cpp
+++ b/test/CommTest.cpp
@@ -13,7 +13,7 @@ namespace RcclUnitTesting
 	ncclTaskCollSorter* me_ptr = new ncclTaskCollSorter;
 	me_ptr->head = nullptr;
 
-	assert(true == ncclTaskCollSorterEmpty(me_ptr));
+	ASSERT_EQ(ncclTaskCollSorterEmpty(me_ptr), true);
 	delete me_ptr;
 
 	INFO("[CommTest] Completed the test\n");
--- a/test/CommTest.cpp
+++ b/test/CommTest.cpp
@@ -3,8 +3,8 @@
  *
  * See LICENSE.txt for license information
  ************************************************************************/
+#include "gtest/gtest.h"
 #include "comm.h"
-#include "TestBed.hpp"
 
 
 
@@ -12,18 +12,12 @@ namespace RcclUnitTesting
 {
   TEST(CommTest, Sorter)
   {
-	TestBed testBed;
-	INFO("[CommTest] starting the test with %d GPUs\n", testBed.numDevicesAvailable);
 	// Configuration
 	ncclTaskCollSorter* me_ptr = new ncclTaskCollSorter;
 	me_ptr->head = nullptr;
 
 	ASSERT_EQ(ncclTaskCollSorterEmpty(me_ptr), true);
 	delete me_ptr;
-
-	INFO("[CommTest] Completed the test\n");
-	testBed.Finalize();
-
   }
 }
 
--- a/test/CommTest.cpp
+++ b/test/CommTest.cpp
@@ -1,0 +1,27 @@
+#include "comm.h"
+#include "TestBed.hpp"
+
+
+
+namespace RcclUnitTesting
+{
+  TEST(CommTest, Sorter)
+  {
+	TestBed testBed;
+	INFO("[CommTest] starting the test with %d GPUs\n", testBed.numDevicesAvailable);
+	// Configuration
+	ncclTaskCollSorter* me_ptr = new ncclTaskCollSorter;
+	me_ptr->head = nullptr;
+
+	assert(true == ncclTaskCollSorterEmpty(me_ptr));
+	delete me_ptr;
+
+	INFO("[CommTest] Completed the test\n");
+	testBed.Finalize();
+
+  }
+}
+
+
+
+
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -91,7 +91,6 @@ if(BUILD_TESTS)
     AllToAllTests.cpp
     AllToAllVTests.cpp
     BroadcastTests.cpp
-    CommTest.cpp
     GatherTests.cpp
     GroupCallTests.cpp
     NonBlockingTests.cpp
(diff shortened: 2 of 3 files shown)


venksubramd:
The test file currently has one test that checks the behavior of the ncclTaskCollSorterEmpty function. The test may be simplified. Instead of creating the instance of ncclTaskCollSorter using new and deleting it, we may create it on the stack and pass a pointer to it to the ncclTaskCollSorterEmpty function.

The ncclTaskCollSorterEmpty function does not handle the case of the pointer to ncclTaskCollSorter being empty. Should this function be returning an error rather than crashing in such case?

Do we need tests for
ncclTaskCollSorterInsert
ncclTaskCollSorterDequeueAll
ncclCommPollCallbacks
ncclCommPollEventCallbacks
ncclCommIntraBarrierIn
ncclCommIntraBarrierOut
ncclUserRedOpMangle
?

ishkool:
> The test file currently has one test that checks the behavior of the ncclTaskCollSorterEmpty function. The test may be simplified. Instead of creating the instance of ncclTaskCollSorter using new and deleting it, we may create it on the stack and pass a pointer to it to the ncclTaskCollSorterEmpty function.
> 
> The ncclTaskCollSorterEmpty function does not handle the case of the pointer to ncclTaskCollSorter being empty. Should this function be returning an error rather than crashing in such case?
> 
> Do we need tests for ncclTaskCollSorterInsert ncclTaskCollSorterDequeueAll ncclCommPollCallbacks ncclCommPollEventCallbacks ncclCommIntraBarrierIn ncclCommIntraBarrierOut ncclUserRedOpMangle ?

1) Some structs are huge and cause a stack overflow if directly declared on the stack, I have witnessed it several times hence I have adopted the heap route. Since compilation takes a lot of time hence I'm following this for everything while making sure I free up the memory afterwards.
2) No error checking is required since ncclTaskCollSorterEmpty just checks if the pointer is a nullptr or not and returns a bool value accordingly.
3) The other mentions functions are already being called/covered and hence no tests are required for those. This is mostly for code coverage activity.

Thank you.

atulkulk:
> > The test file currently has one test that checks the behavior of the ncclTaskCollSorterEmpty function. The test may be simplified. Instead of creating the instance of ncclTaskCollSorter using new and deleting it, we may create it on the stack and pass a pointer to it to the ncclTaskCollSorterEmpty function.
> > The ncclTaskCollSorterEmpty function does not handle the case of the pointer to ncclTaskCollSorter being empty. Should this function be returning an error rather than crashing in such case?
> > Do we need tests for ncclTaskCollSorterInsert ncclTaskCollSorterDequeueAll ncclCommPollCallbacks ncclCommPollEventCallbacks ncclCommIntraBarrierIn ncclCommIntraBarrierOut ncclUserRedOpMangle ?
> 
> 1. Some structs are huge and cause a stack overflow if directly declared on the stack, I have witnessed it several times hence I have adopted the heap route. Since compilation takes a lot of time hence I'm following this for everything while making sure I free up the memory afterwards.
> 2. No error checking is required since ncclTaskCollSorterEmpty just checks if the pointer is a nullptr or not and returns a bool value accordingly.
> 3. The other mentions functions are already being called/covered and hence no tests are required for those. This is mostly for code coverage activity.

@ishkool Let me add some clarification on point 3, the goal here is not just code coverage activity but to have meaningful tests which will help us find any regressions or bugs early in the development cycle. Code coverage metrics does help us in identifying which code paths were taken and which weren't.

> Thank you.

test/CommTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+{
+  TEST(CommTest, Sorter)
+  {
+	TestBed testBed;
+	INFO("[CommTest] starting the test with %d GPUs\n", testBed.numDevicesAvailable);
+	// Configuration
+	ncclTaskCollSorter* me_ptr = new ncclTaskCollSorter;
+	me_ptr->head = nullptr;
+
+	assert(true == ncclTaskCollSorterEmpty(me_ptr));
```
atulkulk: Use GTest API for equality checks.

venksubramd:
@ishkool

> 1. Some structs are huge and cause a stack overflow if directly declared on the stack, I have witnessed it several times hence I have adopted the heap route. Since compilation takes a lot of time hence I'm following this for everything while making sure I free up the memory afterwards.

Thanks you for providing the reason. It may be better to treat this as an exception than a norm. We can keep the tests less noisy and simple by keeping things on the stack and only use the heap when really necessary. Clarity and cleanliness over convenience is something that may be useful to consider. If we are running into heap issues most of the time, then the option to use new/malloc and delete may be reasonable.

ishkool:
> > > The test file currently has one test that checks the behavior of the ncclTaskCollSorterEmpty function. The test may be simplified. Instead of creating the instance of ncclTaskCollSorter using new and deleting it, we may create it on the stack and pass a pointer to it to the ncclTaskCollSorterEmpty function.
> > > The ncclTaskCollSorterEmpty function does not handle the case of the pointer to ncclTaskCollSorter being empty. Should this function be returning an error rather than crashing in such case?
> > > Do we need tests for ncclTaskCollSorterInsert ncclTaskCollSorterDequeueAll ncclCommPollCallbacks ncclCommPollEventCallbacks ncclCommIntraBarrierIn ncclCommIntraBarrierOut ncclUserRedOpMangle ?
> > 
> > 
> > 
> > 1. Some structs are huge and cause a stack overflow if directly declared on the stack, I have witnessed it several times hence I have adopted the heap route. Since compilation takes a lot of time hence I'm following this for everything while making sure I free up the memory afterwards.
> > 2. No error checking is required since ncclTaskCollSorterEmpty just checks if the pointer is a nullptr or not and returns a bool value accordingly.
> > 3. The other mentions functions are already being called/covered and hence no tests are required for those. This is mostly for code coverage activity.
> 
> @ishkool Let me add some clarification on point 3, the goal here is not just code coverage activity but to have meaningful tests which will help us find any regressions or bugs early in the development cycle. Code coverage metrics does help us in identifying which code paths were taken and which weren't.

@atulkulk Agreed but these tests are already being covered so no point in writing redundant tests. 

> 
> > Thank you.

rahulvaidya20 (approved):
Looks okay for now. We can also add another test for non-nullptr case or other edge cases later to check if the function exits gracefully.

venksubramd:
Looks good for now. As @rahulvaidya20 has suggested we should consider those additional tests later. OK to close.

PJAvinash (approved):

Nikhil-Nunna (approved):

PR #1789, opened 2025-07-02
Improving build time by removing the gfx11xx and host code from rccl_float8.h

mberenjk:
## Details

SWDEV-539929
**What were the changes?**  
Removing the host side and gfx11 from OCP type check.

The build time overhead is due to those specific architectures when the build includes all architectures (i.e., without -l).

We achieved the build time reduction by excluding certain architectures from the build.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/include/rccl_float8.h
+++ b/src/include/rccl_float8.h
@@ -24,7 +24,7 @@
 #define ROCBLAS_FLOAT8_H
 
 #include <stdint.h>
-//#include <hip/hip_version.h>
+#include <hip/hip_version.h>
 
 #if __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
 /*! \brief Struct to represent a 8 bit floating-point number. */
--- a/src/include/rccl_float8.h
+++ b/src/include/rccl_float8.h
@@ -24,6 +24,7 @@
 #define ROCBLAS_FLOAT8_H
 
 #include <stdint.h>
+//#include <hip/hip_version.h>
 
 #if __cplusplus < 201103L || (!defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__))
 /*! \brief Struct to represent a 8 bit floating-point number. */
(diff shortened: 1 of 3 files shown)


wenkaidu (approved):

src/include/rccl_float8.h
```diff
@@ -24,7 +24,7 @@
 #define ROCBLAS_FLOAT8_H
 
 #include <stdint.h>
-#include <hip/hip_version.h>
+//#include <hip/hip_version.h>
```
thananon: Why comment but not removal?

src/include/rccl_float8.h
```diff
@@ -40,50 +40,17 @@ typedef struct
...
-
-inline __host__ __device__ float operator*(rccl_float8 a, float b)
-{
-    return float(a) * float(b);
-}
-
-inline __host__ __device__ float operator*(rccl_bfloat8 a, float b)
-{
-    return float(a) * float(b);
-}
```
thananon, line 86: Why are these no longer needed? I thought we only remove for host?

test/common/CollectiveArgs.cpp
```diff
@@ -201,10 +201,10 @@ namespace RcclUnitTesting
         case ncclUint32:     ss << scalarsPerRank.U4[this->globalRank]; break;
         case ncclInt64:      ss << scalarsPerRank.I8[this->globalRank]; break;
         case ncclUint64:     ss << scalarsPerRank.U8[this->globalRank]; break;
-        case ncclFloat8e4m3: ss << scalarsPerRank.F1[this->globalRank]; break;
+        case ncclFloat8e4m3: ss << (float)scalarsPerRank.F1[this->globalRank]; break;
         case ncclFloat32:    ss << scalarsPerRank.F4[this->globalRank]; break;
```
thananon, line 205: Why is this cast needed? Is this for build time or just general correctness?

src/include/rccl_float8.h
mberenjk: Since there is no product operation defined for FP8—neither in hip_fp8.h nor in our local FP8 header—the computation is always cast to FP32 for multiplication. I modified the code to handle the product operation directly, which applies to both host and device code. There's no need to define separate functions for this anymore.

test/common/CollectiveArgs.cpp
mberenjk: This change is necessary to eliminate all host-side dependencies on the FP8 header. It does not affect the build time.

thananon (approved):

PR #1732, opened 2025-06-06
Add proxyTrace

dmwu:
## Details
This feature tracks the proxy events and status of each send/recv op. ProxyTrace keeps a fixed number of active ops in host mem and dumps the status of each op when the program crashes or hangs.

**What were the changes?**  
Added
src/include/proxy_trace/proxy_trace.h
src/misc/proxy_trace/proxy_trace.cc
test/proxy_trace/ProxyTraceUnitTests.cpp

**Implementation Notes**
1. RTR  and RTS counters are not updated for now
2. To decouple with Pytorch Distributed (PTD), we dump the proxy trace in commFree(). So it works for non-PTD cases as well. In PTD's case, watchdog timeout will also call into commFree.
3. Set RCCL_ENABLE_PROXY_TRACE=1 to enable this feature in runtime
4. Only works for multi-node scenarios with network activities. For single-node, traces are empty

Modified init.cc proxy.h, proxy.cc, transport/net.cc to enabling proxy tracing

**How was the outcome achieved?**  
1. UT tests -- Passing:
./build/debug/test/rccl-UnitTests --gtest_filter=ProxyTraceTestFixture*
[==========] Running 6 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 6 tests from ProxyTraceTestFixture
[ RUN      ] ProxyTraceTestFixture.nonEmptySingleton
[       OK ] ProxyTraceTestFixture.nonEmptySingleton (0 ms)
[ RUN      ] ProxyTraceTestFixture.addTraceOp
[       OK ] ProxyTraceTestFixture.addTraceOp (0 ms)
[ RUN      ] ProxyTraceTestFixture.getMapSizeMB
[       OK ] ProxyTraceTestFixture.getMapSizeMB (0 ms)
[ RUN      ] ProxyTraceTestFixture.updateTraceOp
[       OK ] ProxyTraceTestFixture.updateTraceOp (0 ms)
[ RUN      ] ProxyTraceTestFixture.updateTraceOp2
[       OK ] ProxyTraceTestFixture.updateTraceOp2 (0 ms)
[ RUN      ] ProxyTraceTestFixture.memoryReclaim
[       OK ] ProxyTraceTestFixture.memoryReclaim (0 ms)
[----------] 6 tests from ProxyTraceTestFixture (0 ms total)
[----------] Global test environment tear-down
[==========] 6 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 6 tests.

[...]

PR #1791, opened 2025-07-07
[BUILD] Use fmt-header instead of libfmt

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Replaced libfmt.so dependency with fmt-header

https://github.com/ROCm/rccl/pull/1732 introduced a RCCL dependency on `libfmt.so`. This PR switches to using fmt-header.

## Approval Checklist


thananon (approved):

BertanDogancay (approved):

src/include/proxy_trace/proxy_trace.h
```diff
@@ -10,6 +10,9 @@
 #include <chrono>
 #include <cstdint>
 #include <deque>
+#ifndef FMT_HEADER_ONLY
+#define FMT_HEADER_ONLY 1
```
mustafabar, line 14: What is this for? Can you add a comment if this is needed?

gpinkert (approved):
Thanks for this. You mentioned build time differences about this change, did you find the difference neglible?

src/include/proxy_trace/proxy_trace.h
nileshnegi: FMT docs advise using this `FMT_HEADER_ONLY` macro to enable header-only mode. However, if we do not guard this, it throws a pre-definition warning. I thought we could keep this guard if some systems have an older FMT that does not pre-include this macro.

nileshnegi:
> Thanks for this. You mentioned build time differences about this change, did you find the difference neglible?

yep, no difference in RCCL build time.

PR #1792, opened 2025-07-07
[GRAPH] Pass rank instead of busId due to a change in an internal function signature

BertanDogancay:
## Details

SWDEV-541381

**What were the changes?**  
Use rank id instead of busId in the `ncclTopoCheckGdr` to find closest NIC. 

RCCL can't find a path for pattern and chooses the far NIC, which drops performance in some cases. This is seen after the 2.24 sync.

Updated `ncclTopoCheckGdr` function to use rank id.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/graph/search.cc
+++ b/src/graph/search.cc
@@ -696,7 +696,7 @@ ncclResult_t ncclTopoSearchRecNet(struct ncclTopoSystem* system, struct ncclTopo
             if (paths[i].count <= paths[f].count) {
               // prefer GPU direct RDMA
               enum ncclTopoGdrMode useGdr;
-              NCCLCHECK(ncclTopoCheckGdr(system, system->nodes[GPU].nodes[i].id, net->id, 0, &useGdr));
+              NCCLCHECK(ncclTopoCheckGdr(system, system->nodes[GPU].nodes[i].gpu.rank, net->id, 0, &useGdr));
               if (paths[i].count < paths[f].count || (paths[i].count == paths[f].count && !f_gdr && useGdr)) {
                 f = i;
                 f_gdr = useGdr;


thananon (approved):

nileshnegi (approved):

src/graph/search.cc
```diff
@@ -696,7 +696,7 @@ ncclResult_t ncclTopoSearchRecNet(struct ncclTopoSystem* system, struct ncclTopo
             if (paths[i].count <= paths[f].count) {
               // prefer GPU direct RDMA
               enum ncclTopoGdrMode useGdr;
-              NCCLCHECK(ncclTopoCheckGdr(system, system->nodes[GPU].nodes[i].id, net->id, 0, &useGdr));
+              NCCLCHECK(ncclTopoCheckGdr(system, system->nodes[GPU].nodes[i].gpu.rank, net->id, 0, &useGdr));
```
mustafabar, line 699: It doesn't make sense to me to compare the rank id to NIC bus ID. This will break in many cases. Need to check why this failed for the particular pattern rather than changing the comparison terms

mustafabar (changes requested):

src/graph/search.cc
BertanDogancay: It's the same case in other parts of the code. Ex: https://github.com/ROCm/rccl/blob/develop/src/graph/paths.cc#L773

This change is from NCCL 2.24. If we don't want to compare rank id to NIC bus id, we would have to change other places its being used as well. Otherwise, we encounter can't find path for pattern error.

mustafabar: Looks like the implementation of the underlying function has changed to map rank to index using `ncclTopoRankToIndex`. With this modification, passing rankId is a must. Please update the PR title to something like: "[GRAPH] Pass rank instead of busID due to a change in an internal function signature"

BertanDogancay: Updated the PR title.

mustafabar (approved):

PR #1799, opened 2025-07-10
Force enable proto and/or algo after model selection

mustafabar:
## Details

**Work item:** _Internal._

**What were the changes?**  
Enhance debugging for the algo-proto suboptimal selection.

The original logic does not only replace the protocol of default algo/proto choices; instead, it runs the math model using only the specified protocol (i.e., in `NCCL_PROTO`) to determine the best algorithm. As a result, the selected algorithm may change — even though the intention was only to evaluate the impact of changing the protocol. The same behavior applies when overriding the algorithm: it fixes the algorithm but still allows the protocol to vary, potentially masking the intended effect of overriding only the algorithm.

- Add override function after all algo/proto changing code 
- Add two env `RCCL_OVERRIDE_PROTO` and `RCCL_OVERRIDE_ALGO`.

**Additional Documentation:**  
Errors are trapped but ignored to prevent failures from this use

## Approval Checklist


src/rccl_wrap.cc
```diff
@@ -76,6 +76,65 @@ void rcclUpdateCollectiveProtocol(struct ncclComm* comm, size_t const& nBytes, s
...
+    static int protoVal = NCCL_PROTO_UNDEF;
+    if(protoVal == NCCL_PROTO_UNDEF && protoOverrideEnv) {
+      if(rcclGetAlgoProtoIndex(protoOverrideEnv, ncclProtoStr, NCCL_NUM_PROTOCOLS, protoVal) != ncclSuccess) {
+        validInput = false;
+        return ncclInvalidUsage;
+      }
+    }
+    if(protoVal > NCCL_PROTO_UNDEF) {
+      if(table[info->algorithm][protoVal] == NCCL_ALGO_PROTO_IGNORE) {
+        WARN("Failed to force algorithm for function %s with datatype %s.%s", ncclFuncToString(info->func), ncclDatatypeToString(info->datatype), protoOverrideEnv);
```
nileshnegi: is it better to phrase it like: `"Failed to force protocol %s for function %s with datatype %s as it is not supported", protoOverrideEnv, ncclFuncToString(info->func), ncclDatatypeToString(info->datatype)`?

also, should we make this message more explicit for protocols like `LL128`, which are not supported on single node?

mustafabar: For the LL128 note, I think it will just not work. Better to keep it generic.
Modified the message based on your feedback

nusislam:
With this PR, if one wants to set a protocol without changing the algo, he/she needs to set both `NCCL_PROTO` and `RCCL_OVERRIDE_PROTO`?

mustafabar:
> With this PR, if one wants to set a protocol without changing the algo, he/she needs to set both `NCCL_PROTO` and `RCCL_OVERRIDE_PROTO`?
@nusislam  `RCCL_OVERRIDE_PROTO` is enough

src/rccl_wrap.cc
```diff
@@ -77,6 +77,65 @@ void rcclUpdateCollectiveProtocol(struct ncclComm* comm, size_t const& nBytes, s
...
+    if (!failedProtoWarn) {
+      WARN("Invalid algo or protocol string passed %s", envStr);
+      failedProtoWarn = true;
+      return ncclInvalidUsage;
+    }
+  }
+  return ncclInvalidUsage;
+}
+
+ncclResult_t rcclOverrideAlgoOrProto(const char* ncclAlgoStr[], const char* ncclProtoStr[], float table[][NCCL_NUM_PROTOCOLS],  struct ncclTaskColl* info) {
```
atulkulk: I think it would it better to have this into two separate functions, one for proto and the other one for algo in terms of single responsibility.

mustafabar: @atulkulk, done

atulkulk (approved):

PedramAlizadeh (approved):

PR #1801, opened 2025-07-11
Compiler warnings fix 2

PJAvinash:
## Details
Minimal changes to code to address as many compiler warnings as possible

**Work item:** "Internal"
**What were the changes?**  
The following files are specific rccl 

- src/graph/rome_models.cc
- src/misc/msccl/*
- src/misc/alt_rsmi.cc
- src/misc/recorder.cc
- src/misc/proxy_trace/proxy_trace.cc

Hence, changes to these files do now show up during sync from upstream. Any changes to other files should have an indication that it is 'compiler warning' fix. (proxy.cc has very minimal change that doesn't require explicit instruction)

To reduce compiler warnings during builds

Addressing compiler warnings for all the supported architectures without modifying functionality 

**Additional Documentation:**  
- 95% + Compiler warnings are addressed
- warnings from ext-src are not addressed in this PR
- 'hipCtxGetCurrent' is marked as  deprecated, This is not addressed in this PR
- An implicit NULL to FALSE(bool) conversion in Primitives<T, RedOp, FanSymmetric<1>, 0, Proto, 0, isNetOffload> prims , in allgather.h is not addressed

### DIFF
--- a/src/device/all_gather.h
+++ b/src/device/all_gather.h
@@ -242,7 +242,9 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
       while (1) {
         struct ncclPatStep* ps = shmem->patSteps+(step%NCCL_SHMEM_PAT_STEPS);
         int* poll = &ps->flags;
-        while (__hip_atomic_load(poll, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_WORKGROUP) != 0) //pollCount++ ;//unused variable compiler waring // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
+        while (__hip_atomic_load(poll, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_WORKGROUP) != 0) { 
+          //pollCount++ ;//unused variable compiler warning // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
+        }
         patAlgo.getNextOp(ps);
         int last = ps->last;
         step++;
--- a/src/device/all_gather.h
+++ b/src/device/all_gather.h
@@ -16,7 +16,10 @@ namespace {
 #else
   __device__ __attribute__((noinline)) void runRing(int tid, int nthreads, struct ncclDevWorkColl* work) {
 #endif
+#if defined(ENABLE_NPKIT)
     const int bid = ncclShmem.channelId - work->channelLo;
+    int npKitCtxIdx = bid; // unused variable - compiler warning
+#endif
     ncclRing *ring = &ncclShmem.channel.ring;
     const int *ringRanks = ring->userRanks;
     const int nranks = ncclShmem.comm.nRanks;
@@ -27,9 +30,6 @@ namespace {
     int nelem;
     int rankDest;
 
-#if defined(ENABLE_NPKIT)
-    int npKitCtxIdx = bid;
-#endif
 
 #if defined(ENABLE_NPKIT) && defined(ENABLE_NPKIT_EVENT_TIME_SYNC_CPU)
     if (tid == 0) {
@@ -228,7 +228,7 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
 
     static constexpr int nworkers = NCCL_PAT_NWORKERS;
     struct ncclPatShmem* shmem = (struct ncclPatShmem*)ncclScratchForWarp(0);
-    uint64_t pollCount = 0;
+    // uint64_t pollCount = 0; unused variable - compiler warning
     __syncthreads(); // Don't start using shared mem until everyone arrives
     for (int i=tid; i<NCCL_SHMEM_PAT_STEPS; i+=nthreads) shmem->patSteps[i].flags = 0;
     if (tid == 0) shmem->localAccSize = 0;
@@ -237,12 +237,12 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
 
     if (tid == nworkers) { // Algo computation thread
       PatAGAlgorithm<T> patAlgo(chunkCount*sizeof(T), NCCL_STEPS, NCCL_PAT_NWORKERS/WARP_SIZE, channelOffset, channelOffset + channelCount, count, chunkCount, rank, nranks);
-      int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor();
+      //int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor(); // unused variable compiler warning
       int step = 0;
       while (1) {
         struct ncclPatStep* ps = shmem->patSteps+(step%NCCL_SHMEM_PAT_STEPS);
         int* poll = &ps->flags;
-        while (__hip_atomic_load(poll, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_WORKGROUP) != 0) pollCount++; // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
+        while (__hip_atomic_load(poll, __ATOMIC_ACQUIRE, __HIP_MEMORY_SCOPE_WORKGROUP) != 0) //pollCount++ ;//unused variable compiler waring // Wait for workers to be done with step 'step-NCCL_SHMEM_PAT_STEPS'
         patAlgo.getNextOp(ps);
         int last = ps->last;
         step++;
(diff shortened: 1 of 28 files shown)


src/graph/paths.cc
```diff
@@ -286,14 +286,17 @@ ncclResult_t ncclTopoCheckP2p(struct ncclComm* comm, struct ncclTopoSystem* syst
     // GPU not found, we can't use p2p.
     return ncclSuccess;
   }
-
+  #if !defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__)
```
nileshnegi, line 289: why is this guard removed?

src/misc/proxy_trace/proxy_trace.cc
```diff
@@ -148,8 +148,8 @@ std::string facebook_rccl::ProxyTrace::dump(uint64_t commHash) {
       sortedDumpStrMap[traceKey.str()] = proxyOpMap.second.str();
     }
   }
```
nileshnegi, line 149: instead of all these changes, maybe we should consider moving to C++ 17?

src/graph/paths.cc
thananon: its added no?

thananon (approved):

src/graph/paths.cc
nileshnegi: my bad

src/misc/proxy_trace/proxy_trace.cc
PJAvinash: I can update the PR once we have the concrete decision on minimum C++ version that we would like to support for RCCL

corey-derochie-amd: I believe we are currently on C++14 in compliance with NCCL. We should be sure that no code semantically changes if we move to C++17. I don't see a problem, especially if we add C++17 code to RCCL portions only, but I don't want to have to worry about NCCL code in that...

thananon: Lets keep it this way for now and get this in.

src/device/all_gather.h
```diff
@@ -237,12 +237,14 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
 
     if (tid == nworkers) { // Algo computation thread
       PatAGAlgorithm<T> patAlgo(chunkCount*sizeof(T), NCCL_STEPS, NCCL_PAT_NWORKERS/WARP_SIZE, channelOffset, channelOffset + channelCount, count, chunkCount, rank, nranks);
-      int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor();
+      //int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor(); // unused variable compiler warning
```
corey-derochie-amd: This line also sets `shmem->parallelFactor`, so commenting this removes that side effect.

corey-derochie-amd (changes requested):

src/device/all_gather.h
```diff
@@ -228,7 +228,7 @@ struct RunWorkColl<ncclFuncAllGather, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SIMPLE
 
     static constexpr int nworkers = NCCL_PAT_NWORKERS;
     struct ncclPatShmem* shmem = (struct ncclPatShmem*)ncclScratchForWarp(0);
-    uint64_t pollCount = 0;
+    // uint64_t pollCount = 0; unused variable - compiler warning
```
corey-derochie-amd: An alternative would be this, which would remove the need for changes below:

```suggestion
    uint64_t pollCount = 0;
    (void)pollCount; // unused variable - compiler warning
```

corey-derochie-amd: Any thoughts on which style we should prefer?

src/device/prims_ll.h
```diff
@@ -148,7 +148,7 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload>:
   __device__ uint64_t readLL(int offset, int i) {
     union ncclLLFifoLine* src = recvPtr(i) + offset;
     uint32_t flag = recvFlag(i);
-    uint32_t data1, flag1, data2, flag2;
+    //uint32_t data1, flag1, data2, flag2; // unused variable - compiler warning
```
corey-derochie-amd: Since these are used in one of the #if/#else branches, you can remove the warning in the other branch like this:
```
(void)data1; (void)flag1; (void)data2; (void)flag2; // unused variable - compiler warning
```

src/device/prims_ll.h
```diff
@@ -855,8 +856,8 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload>:
       int eltInLine = EltPerLine < nelem ? EltPerLine : nelem;
 
       DataLoader dl;
-      ncclLLFifoLine line[MaxRecv];
-      uint64_t data, peerData;
+      // ncclLLFifoLine line[MaxRecv];//unused variable - compiler warning
+      uint64_t data;                  //unused variable - compiler warning
```
corey-derochie-amd: Might consider this so we see what was removed.
```suggestion
      uint64_t data/*, peerData*/;    //unused variable - compiler warning
```

src/device/prims_ll128.h
```diff
@@ -507,10 +507,9 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL128, P2p, isNetOffload>:
       bool ipcReg = false, bool netReg = false, int stepSize_ = 0
     ):
     redOp(redOpArg),
-    tid(tid), nthreads(nthreads), wid(tid%WARP_SIZE), warp(tid/WARP_SIZE),
-    warpInBlock(threadIdx.x/WARP_SIZE),
-    flagThread((tid%4)==3), group(group),
-    stepSize(ncclShmem.comm.buffSizes[NCCL_PROTO_LL128]/NCCL_STEPS/sizeof(uint64_t)) {
```
corey-derochie-amd, line 527: Same suggestion about commenting.

src/device/reduce_scatter.h
```diff
@@ -187,12 +187,14 @@ struct RunWorkColl<ncclFuncReduceScatter, T, RedOp, NCCL_ALGO_PAT, NCCL_PROTO_SI
 
     if (tid == nworkers) { // Algo computation thread
       PatRSAlgorithm<T> patAlgo(chunkCount*sizeof(T), NCCL_STEPS, NCCL_PAT_NWORKERS/WARP_SIZE, channelOffset, channelOffset + channelCount, count, chunkCount, rank, nranks);
-      int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor();
+      //int parallelFactor = shmem->parallelFactor = patAlgo.getParallelFactor(); //unused variable - compiler warning
```
corey-derochie-amd: Same problem with `shmem->parallelFactor`.

corey-derochie-amd (changes requested):

src/init.cc
```diff
@@ -1871,9 +1871,9 @@ static ncclResult_t commGetSplitInfo(struct ncclComm* comm, struct ncclComm* par
 static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   struct ncclCommInitRankAsyncJob* job = (struct ncclCommInitRankAsyncJob*)job_;
   ncclComm_t comm = job->comm;
-#ifdef ENABLE_MSCCLPP
-  ncclUniqueId origUniqueId = *job->commId;
-#endif
+// #ifdef ENABLE_MSCCLPP
+//   ncclUniqueId origUniqueId = *job->commId;
+// #endif
```
corey-derochie-amd: We should double check that this unused variable isn't actually a bug, this was originally here for a reason. If not, these lines can just be removed, not commented.

src/misc/proxy_trace/proxy_trace.cc
corey-derochie-amd: Let's remember to revisit this after 2.27 goes in, as I think we can safely switch then.

src/transport/coll_net.cc
```diff
@@ -523,8 +523,8 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   NCCLCHECK(sharedBuffersInit(connection->collNet, resources->useGdr, &mapMem->gpuPtr, &mapMem->cpuPtr, &mapMem->size));
   NCCL_NET_MAP_ADD_POINTER(map, 1, resources->useGdr ? 1 : 0, mapMem->size, buffs[NCCL_PROTO_SIMPLE]);
 
-  int dmabuf_fd = -1;
 #if CUDA_VERSION >= 11070
```
corey-derochie-amd: Could use `(void)` here to avoid all the changes below.

src/device/prims_ll128.h
PJAvinash: Change here is to match the initialization order. @corey-derochie-amd  Please resolve the comment if it is good.

src/init.cc
PJAvinash: origUniqueId symbol doesn't appear anywhere else in this file.

src/transport/coll_net.cc
PJAvinash: This is addressed in the new commit

corey-derochie-amd (approved):

BertanDogancay (approved):

PR #1817, opened 2025-07-17
Code Coverage: Infiniband unit tests

ishkool:
## Details

**Work item:** _"Internal"

**What were the changes?**  
 Adding tests for infiniband network functions.

For improving the code coverage.
**Added 12 tests for uncovered function, the functional code coverage is 97.6% (84/86) from previous 83%(72/86)**

**Additional Documentation:**  

## Approval Checklist


venksubramd:
TEST(IbTest, UpdateGidIndex)
Please look into this test to see if we can add some asserts and also if it will be useful to split this test into multiple tests.

TEST(IbTest, GidAddrPrefix)
We have an assert for true, do we need a test for assert of false return value? Also, do we need more to test the actual behavior of this function?

TEST(IbTest, AddrRange)
Do we need to check the value assigned to mask? Other behavior of this function?

TEST(IbTest, Devftl)
This test looks like sufficient for this function.

TEST(IbTest, EnvAddrFamily)
Do we need more tests for other returned values for the res, like AF_INET6?

TEST(IbTest, GetGidAddrFamily)
Do we need more tests for other returned values for the res?

TEST(IbTest, ConfiguredGid)
Do we need a test for the false return value?

TEST(IbTest, LinkLocalGid)
Please include assert so test indicates fail if unsuccessful.

TEST(IbTest, ValidGid)
Please include assert so test indicates fail if unsuccessful.

TEST(IbTest, RoceVer)
Do we need to verify behavior of this function in addition to checking the return value?
What about situations that return something other than ncclSuccess, like error for instance?

test/CMakeLists.txt
```diff
@@ -56,63 +90,97 @@ if(BUILD_TESTS)
...
-  target_include_directories(rccl-UnitTests PRIVATE ${PROJECT_BINARY_DIR}/hipify/src/include)        # for rccl_bfloat16.h
-  target_include_directories(rccl-UnitTests PRIVATE ${PROJECT_BINARY_DIR}/hipify/src/include/plugin) # for recorder tests
+  set(RCCL_TEST_EXECUTABLES rccl-UnitTests)
+
+  # Create rccl-UnitTestsFixtures binary if ROCm version is 4.6.0 or greater
+  # and build type is Debug
+  if (ROCM_VERSION VERSION_GREATER_EQUAL "60400" AND CMAKE_BUILD_TYPE MATCHES "Debug")
+
+    set(TEST_FIXTURE_SOURCE_FILES
+	    #AltRsmiTests.cpp
```
atulkulk: Why these are commented out?

test/IbTest.cpp
```diff
@@ -0,0 +1,375 @@
...
+ncclResult_t ncclIbRoceGetVersionNum(const char* deviceName, int portNum, int gidIndex, int* version);
+void* envIbAddrRange(sa_family_t af, int* mask);
+bool matchGidAddrPrefix(sa_family_t af, void* prefix, int prefixlen, union ibv_gid* gid);
+ncclResult_t ncclUpdateGidIndex(struct ibv_context* context, uint8_t portNum, sa_family_t af, void* prefix, int prefixlen, int roceVer, int gidIndexCandidate, int* gidIndex);
+
+namespace RcclUnitTesting
+{
+  TEST(IbTest, UpdateGidIndex)
+  {
+    INFO("[IbTest] Test begin \n");
```
atulkulk: Why INFO is being used everywhere here? Where are the input/output checks here in the tests? i.e. expected v/s actual results. The goal here is just not to make a function call to get the coverage but to check the behavior of the function by having meaningful tests.

test/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ cmake_minimum_required(VERSION 2.8.12)
 if(BUILD_TESTS)
```
atulkulk, line 6: Rebase your branch with the RCCL's develop branch.

atulkulk (changes requested):

test/CMakeLists.txt
ishkool: Was not using it so just wanted to reduce number of variables, will keep the previous ones as is then.

test/IbTest.cpp
ishkool: INFO prints will only come if we have set the export UT_VERBOSE=1 flag which is the part of google test framework, so it is fine.
Regarding the expected vs actual output, most of these functions are querying functions so we get back what is configued by the system, I'll check with you on that I don't know what to do here.

ishkool:
> TEST(IbTest, UpdateGidIndex) Please look into this test to see if we can add some asserts and also if it will be useful to split this test into multiple tests.

Actually, all functions used by this test have their corresponding tests in this file I'll add asserts when they are called by this function.

> TEST(IbTest, GidAddrPrefix) We have an assert for true, do we need a test for assert of false return value? Also, do we need more to test the actual behavior of this function?

Function will return false if perfix does not match, we are providing dummy value such that it matches hence the assert on true. Also function does most of work while matching so that is targeted as well.
> TEST(IbTest, AddrRange) Do we need to check the value assigned to mask? Other behavior of this function?

Not needed we just want the pointer to not be null, mask is based on the input.
> TEST(IbTest, Devftl) This test looks like sufficient for this function.

Yes cause it just does an atomic addition operation over a struct variable.
> TEST(IbTest, EnvAddrFamily) Do we need more tests for other returned values for the res, like AF_INET6?

Thanks I'll check if ipv6 is supported which most probably will be and will modify accordingly.
Update: It's an env variable which defaults to ipv4 so not changing anything here.
> TEST(IbTest, GetGidAddrFamily) Do we need more tests for other returned values for the res?

It just returns which prortocal we use 2 is for AF_INET IPV4 , I think we support ipv4 and ipv6 qill check that.
> TEST(IbTest, ConfiguredGid) Do we need a test for the false return value?

I don't think so, what you want to configure should be successful unless inputs are bad, and we are already using correct inps in the dummy vals.
> TEST(IbTest, LinkLocalGid) Please include assert so test indicates fail if unsuccessful.

Thanks, will do.
> TEST(IbTest, ValidGid) Please include assert so test indicates fail if unsuccessful.

Thanks, will do.

> TEST(IbTest, RoceVer) Do we need to verify behavior of this function in addition to checking the return value? What about situations that return something other than ncclSuccess, like error for instance?

If error is returned then Roce is not supported maybe.

test/IbTests.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+  INFO(NCCL_LOG_INFO, "init val %u \n", res);
+  res = validGid(&gid);
+  INFO(NCCL_LOG_INFO, "[IbTests] final val %u \n", res);
+  ASSERT_EQ(res, true);
+
+  INFO(NCCL_LOG_INFO, "[IbTests] test complete\n");
+}
+
+TEST(IbTests, RoceVer) {
```
rahulvaidya20, line 404: This test will likely fail on systems where IB cards/drivers are not present. Need to check if `/sys/class/infiniband/mlx5` exists before asserting it to true/false. It will be good if we can add another test for cases where `ncclIbRoceGetVersionNum` function returns `ncclRemoteError`

test/IbTests.cpp
```diff
@@ -0,0 +1,422 @@
...
+           "expected value. Try running with export "
+           "NCCL_IB_ADDR_RANGE=192.168.1.1/24 \n",
+           env);
+    }
+  }
+}
+
+TEST(IbTests,
+     AddrRangeIPV6) // Should be run with 'export
+                    // NCCL_IB_ADDR_RANGE=2001:0db8:85a3::/64' for ipv6 test
```
rahulvaidya20, line 154: Would it be better to set this env var instead?

test/IbTests.cpp
```diff
@@ -0,0 +1,422 @@
...
+
+  bool result_ipv6 = matchGidAddrPrefix(AF_INET6, &prefix2, 64, &gid2);
+  INFO(NCCL_LOG_INFO, "Match result ipv6: %s\n",
+       result_ipv6 ? "true" : "false");
+  ASSERT_EQ(result_ipv6, true);
+  INFO(NCCL_LOG_INFO, "[IbTests] Test complete \n");
+}
+
+TEST(IbTests, AddrRangeIPV4) // Should be run with 'export
+                             // NCCL_IB_ADDR_RANGE=192.168.1.1/24' for ipv4 test
```
rahulvaidya20, line 123: Would it be better to set this env var instead?

rahulvaidya20 (commented):
Some of the tests seem to only validate "happy path" scenarios. We also need to add more tests to cover error paths and other edge cases to make sure the functions don't break. This can be addressed in a separate PR.

test/IbTests.cpp
ishkool: Not needed, don't want to set the env just for a test, you can check with @atulkulk regarding this, thanks.

test/IbTests.cpp
ishkool: Not needed, don't want to set the env just for a test, you can check with @atulkulk regarding this, thanks.

test/IbTests.cpp
ishkool: That is the point tests should fail on a system which does not have IB cards :)

rahulvaidya20: Rather than letting the test fail on non-IB systems, it would make more sense to check if the function correctly handles this type of scenario and exits gracefully.

ishkool: You can add it later.

venksubramd:
We could create another PR for the cases that @rahulvaidya20 has suggested and close this one.

test/IbTests.cpp
```diff
@@ -0,0 +1,422 @@
...
+                                     int gidIndex, int *version);
+void *envIbAddrRange(sa_family_t af, int *mask);
+bool matchGidAddrPrefix(sa_family_t af, void *prefix, int prefixlen,
+                        union ibv_gid *gid);
+ncclResult_t ncclUpdateGidIndex(struct ibv_context *context, uint8_t portNum,
+                                sa_family_t af, void *prefix, int prefixlen,
+                                int roceVer, int gidIndexCandidate,
+                                int *gidIndex);
+
+int compareStrings(const char *str1, const char *str2) {
```
PJAvinash, line 82: What happens if one of the args is NULL  ?

PJAvinash (approved):

Nikhil-Nunna (approved):

test/IbTests.cpp
rahulvaidya20: Need to cover these types of edge cases in a separate PR.

atulkulk:
Closing out this PR. Tests will be re-worked.

PR #1818, opened 2025-07-17
Code Coverage: Proxy.cc tests

ishkool:
## Details

**Work item:** "Internal"
**What were the changes?**  
Tests for proxy.cc file

To improve code coverage.
**Added 6 tests for uncovered function, the functional code coverage is 92.7% (51/55) from previous 81.82%(45/55)**

**Additional Documentation:**  

## Approval Checklist


venksubramd:
TEST(ProxyTest, getOpIndex)
We are asserting for return value to be ncclSuccess. It will be useful to check for the values being set by the getOpIndex function.

Also, a test for situation when the return is not ncclSuccess?

TEST(ProxyTest, printProxyOp)
It will be useful to assert values other than there returned ncclSuccess. Having said that, this function merely prints and there may not be much value testing that unless we need to truly verify what gets written out.

TEST(ProxyTest, dumpProxyState)
The dumpProxyState is a long function. We will need tests to verify its behavior. The assert in this test only checks for the result of success.

TEST(ProxyTest, ncclProxyCallBlockingUDS)
The asserts checks a range of value being returned. It does not assert any of the expected behavior of the function being tested, however.

The next few tests similarly assert a range of return values and do not focus on the behavior of the function being tested.

TEST(ProxyTest, Index2)
Is this necessary?

ishkool:
> getOpIndex

ishkool:
> TEST(ProxyTest, getOpIndex) We are asserting for return value to be ncclSuccess. It will be useful to check for the values being set by the getOpIndex function.
> Also, a test for situation when the return is not ncclSuccess?

True, will add checks for that, thanks.
> TEST(ProxyTest, printProxyOp) It will be useful to assert values other than there returned ncclSuccess. Having said that, this function merely prints and there may not be much value testing that unless we need to truly verify what gets written out.

Not sure how you'll test that, cause I see the prints coming in the terminal/stderr
> TEST(ProxyTest, dumpProxyState) The dumpProxyState is a long function. We will need tests to verify its behavior. The assert in this test only checks for the result of success.

This is a parent function calling many child functions all of which have separate tests in this file itself, so not needed.
> TEST(ProxyTest, ncclProxyCallBlockingUDS) The asserts checks a range of value being returned. It does not assert any of the expected behavior of the function being tested, however.
>
> The next few tests similarly assert a range of return values and do not focus on the behavior of the function being tested.

Yep cause they need mocking of underlying socket functions, which I'm yet to do, will do that.
> TEST(ProxyTest, Index2) Is this necessary?

Nope will remove it, was part of debugging.

test/ProxyTests.cpp
```diff
@@ -0,0 +1,395 @@
...
+TEST(ProxyTests,
+     getOpIndex) { // Tests what is the index of the pool being passed within
+                   // the known valid pools in state ptr
+  INFO(NCCL_LOG_INFO, "[ProxyTests] Test Start \n");
+
+  // Init Dummy structs
+  struct ncclProxyArgs *pool_ptr = new ncclProxyArgs;
+  struct ncclProxyPool *pools_ptr = new ncclProxyPool;
+  struct ncclProxyPool *pools2_ptr = new ncclProxyPool;
+  struct ncclProxyProgressState *state_ptr = new ncclProxyProgressState;
```
rahulvaidya20, line 78: Please delete these allocated structs at the end to avoid memory leaks.

test/ProxyTests.cpp
```diff
@@ -0,0 +1,395 @@
...
+TEST(ProxyTests, printProxyOp) {
+  INFO(NCCL_LOG_INFO, "[ProxyTests] Test Start \n");
+  // Init Dummy structs
+
+  struct ncclProxyArgs *pool_ptr = new ncclProxyArgs;
+
+  struct ncclProxyPool *pools_ptr = new ncclProxyPool;
+  struct ncclProxyPool *pools2_ptr = new ncclProxyPool;
+
+  struct ncclProxyProgressState *state_ptr = new ncclProxyProgressState;
```
rahulvaidya20, line 122: Same as above

rahulvaidya20 (changes requested):
Please address memory leak issues. Some tests only check for return values. We need a separate PR to make these tests more meaningful and check for core functionality.

test/ProxyTests.cpp
ishkool: Done, now please get this merged

test/ProxyTests.cpp
ishkool: Done, now please get this merged

venksubramd:
OK to close this after we create PRs for the additional tasks.

test/ProxyTests.cpp
```diff
@@ -0,0 +1,433 @@
...
+#include <sched.h>
+#include <sys/syscall.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#define NCCL_MAX_OPS (2048)
+#define OP_INDEX(op) ((op) ? (op) - state->pools->elems : -1)
+#define OP_SEEN 0x100000
+
```
PJAvinash, line 31: In the below functions read-only args (const ptr) and non read-only args can be indicated for better readability.

PJAvinash (approved):

Nikhil-Nunna (approved):

rahulvaidya20 (approved):

PR #1820, opened 2025-07-17
Fix segfault when libibverbs returns 0 device.

thananon:
Fix: SWDEV-543816

## Details

**Work item:** : SWDEV-543816

**What were the changes?**  
Check for NULL pointer before calling free.

In some configuration where libibverbs is available but no device, it returns 0 device count but the device list (NULL) is still passed to free. This PR checks for NULL pointer.

One line if-statement.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/misc/ibvwrap.cc
+++ b/src/misc/ibvwrap.cc
@@ -107,6 +107,7 @@ ncclResult_t wrap_ibv_get_device_list(struct ibv_device ***ret, int *num_devices
 }
 
 ncclResult_t wrap_ibv_free_device_list(struct ibv_device **list) {
+  if (list == nullptr) return ncclSuccess;
   IBV_PASSTHRU(ibvSymbols, ibv_internal_free_device_list, ibv_internal_free_device_list(list));
 }
 


rahulvaidya20 (approved):

src/misc/ibvwrap.cc
```diff
@@ -107,6 +107,7 @@ ncclResult_t wrap_ibv_get_device_list(struct ibv_device ***ret, int *num_devices
 }
 
 ncclResult_t wrap_ibv_free_device_list(struct ibv_device **list) {
+  if (list == NULL) return ncclSuccess;
```
nileshnegi: is it better/safer to use `nullptr`?

thananon: It seems like we use `NULL` in many other places in this file so I went with it. I can change to `nullptr`. What do you think?

corey-derochie-amd: We can use modern code for new lines.

nileshnegi (approved):

PedramAlizadeh (approved):

src/misc/ibvwrap.cc
```diff
@@ -107,6 +107,7 @@ ncclResult_t wrap_ibv_get_device_list(struct ibv_device ***ret, int *num_devices
 }
 
 ncclResult_t wrap_ibv_free_device_list(struct ibv_device **list) {
+  if (list == nullptr) return ncclSuccess;
```
mustafabar, line 110: Maybe don't call on a nullptr? 
No strong opinion (i.e. this is a valid fix imo), but if you free a nullptr in C there's no graceful exit like here

PJAvinash: ibv_free_device_list() should be called even when there are zero devices on the system, if ibv_get_device_list() called before. Else it will have memory leak of 8 bytes.

thananon: @nileshnegi voiced the same concern as @mustafabar but my rationale is, the alternative is to check at `wrap_ibv_free_device_list()` call, which you suggested. I think putting this here is more fool proof.

corey-derochie-amd: In general I'd prefer @mustafabar 's suggestion as it's more semantically correct. In this case, I think it's fine to make `wrap_ibv_free_device_list` nop on a null list because it's just a free function.

PR #1821, opened 2025-07-21
Add Unit Test for bitops.h

peizhang56:
## Details
Add Unit Test For bitopts.h

**What were the changes?**  
1. Changed UnitTest Camke to honor code coverage compiler flags
2. Added UnitTest for bitops.h to improve the coverage to  95.65% (22/23)

The motivation is to improve Unit Test Coverage

Change in CMakeLists.txt file.
Added Unit Test

**Additional Documentation:**  
If we don't honor the code coverage flags in the CMakeLists.txt, the code coverage report will not be generated for the bitopts.h

## Approval Checklist

### DIFF
--- a/test/BitOpsTests.cpp
+++ b/test/BitOpsTests.cpp
@@ -1,0 +1,225 @@
+#include "bitops.h"
+#include "gtest/gtest.h"
+
+class BitOpsTestsFixture : public ::testing::Test {
+public:
+    BitOpsTestsFixture() = default;
+    ~BitOpsTestsFixture() override = default;
+protected:
+    int m_size{0};
+    uint32_t m_u32Val{2};
+    uint8_t m_u8Val{2};
+    std::vector<uint32_t> m_u32Vec{2, 4, 8, 16, 32, 64, 128, 256};
+    std::vector<float> m_floatVec{2.0f};
+};
+
+
+TEST_F(BitOpsTestsFixture, DIVUP_Tests) {
+    EXPECT_EQ(DIVUP(0, 8), 0);
+    EXPECT_EQ(DIVUP(1, 8), 1);
+    EXPECT_EQ(DIVUP(7, 8), 1);
+    EXPECT_EQ(DIVUP(8, 8), 1);
+    EXPECT_EQ(DIVUP(9, 8), 2);
+}
+
+TEST_F(BitOpsTestsFixture, ROUNDUP_Tests) {
+    EXPECT_EQ(ROUNDUP(0, 8), 0);
+    EXPECT_EQ(ROUNDUP(1, 8), 8);
+    EXPECT_EQ(ROUNDUP(7, 8), 8);
+    EXPECT_EQ(ROUNDUP(8, 8), 8);
+    EXPECT_EQ(ROUNDUP(9, 8), 16);
+    EXPECT_EQ(ROUNDUP(15, 8), 16);
+    EXPECT_EQ(ROUNDUP(16, 8), 16);
+    EXPECT_EQ(ROUNDUP(17, 8), 24);
+}
+
+TEST_F(BitOpsTestsFixture, ALIGN_POWER_Tests) {
+    EXPECT_EQ(ALIGN_POWER(7, 8), 8);
+    EXPECT_EQ(ALIGN_POWER(8, 8), 8);
+    EXPECT_EQ(ALIGN_POWER(9, 8), 16);
+}
+
+TEST_F(BitOpsTestsFixture, ALIGN_SIZE_Tests) {
+    m_size = 1;
+    ALIGN_SIZE(m_size, 8);
+    EXPECT_EQ(m_size, 8);
+    m_size = 7;
+    ALIGN_SIZE(m_size, 8);
+    EXPECT_EQ(m_size, 8);
+}
+
+TEST_F(BitOpsTestsFixture, divUp_Tests) {
+    EXPECT_EQ(divUp(0, 8), 0);
+    EXPECT_EQ(divUp(1, 8), 1);
+    EXPECT_EQ(divUp(7, 8), 1);
+    EXPECT_EQ(divUp(8, 8), 1);
+    EXPECT_EQ(divUp(9, 8), 2);
+    EXPECT_EQ(divUp(15, 8), 2);
+    EXPECT_EQ(divUp(16, 8), 2);
+    EXPECT_EQ(divUp(17, 8), 3);
+}
+
+TEST_F(BitOpsTestsFixture, roundUp_Tests) {
+    EXPECT_EQ(roundUp(0, 8), 0);
+    EXPECT_EQ(roundUp(1, 8), 8);
+    EXPECT_EQ(roundUp(7, 8), 8);
+    EXPECT_EQ(roundUp(8, 8), 8);
+    EXPECT_EQ(roundUp(9, 8), 16);
+    EXPECT_EQ(roundUp(15, 8), 16);
+    EXPECT_EQ(roundUp(16, 8), 16);
+    EXPECT_EQ(roundUp(17, 8), 24);
+}
+
+TEST_F(BitOpsTestsFixture, roundDown_Tests) {
+    EXPECT_EQ(roundDown(0, 8), 0);
+    EXPECT_EQ(roundDown(1, 8), 0);
+    EXPECT_EQ(roundDown(7, 8), 0);
+    EXPECT_EQ(roundDown(8, 8), 8);
+    EXPECT_EQ(roundDown(9, 8), 8);
+    EXPECT_EQ(roundDown(15, 8), 8);
+    EXPECT_EQ(roundDown(16, 8), 16);
+    EXPECT_EQ(roundDown(17, 8), 16);
+}
+
+TEST_F(BitOpsTestsFixture, alignUp_Tests) {
+    EXPECT_EQ(alignUp(0, 8), 0);
+    EXPECT_EQ(alignUp(1, 8), 8);
+    EXPECT_EQ(alignUp(7, 8), 8);
+    EXPECT_EQ(alignUp(8, 8), 8);
+    EXPECT_EQ(alignUp(9, 8), 16);
+    EXPECT_EQ(alignUp(15, 8), 16);
+    EXPECT_EQ(alignUp(16, 8), 16);
+    EXPECT_EQ(alignUp(17, 8), 24);
+}
+
+
+TEST_F(BitOpsTestsFixture, u32fp8MaxValue) {
+    EXPECT_EQ(u32fp8MaxValue(), 0xf0000000);
+}
+
+TEST_F(BitOpsTestsFixture, u32fp8Encode) {
+    uint8_t expected = 2;
+    EXPECT_EQ(u32fp8Encode(this->m_u32Val), expected);
+}
+
+TEST_F(BitOpsTestsFixture, u32fp8Decode) {
+    uint32_t expected = 2;
+    EXPECT_EQ(u32fp8Decode(this->m_u8Val), expected);
+}
+
+TEST_F(BitOpsTestsFixture, getHash) {
+    uint64_t expected = 0xa4495d05731e3337;
+    auto ret = getHash(this->m_u32Vec.data(), this->m_u32Vec.size() * sizeof(uint32_t));
+    EXPECT_EQ(ret, expected);
+
+    expected = 0xedbaa57e84d6dbaa;
+    ret = getHash(this->m_floatVec.data());
+    EXPECT_EQ(ret, expected);
+}
+
+template <typename T>
+class BitOpsTemplateAllIntTestsFixture : public testing::Test {
+public:
+    BitOpsTemplateAllIntTestsFixture() = default;
+    ~BitOpsTemplateAllIntTestsFixture() override = default;
+protected:
+    T m_countOneBitVal{3};
+    T m_firstOneBitVal{3};
+    T m_popFirstOneBitVal{3};
+    T m_log2DownVal{0};
+    T m_log2UpVal{0};
+    T m_power2UpVal{0};
+    T m_power2DownVal{1};
+};
+
+using BitOpsAllIntTypes = ::testing::Types<int, unsigned int, long, unsigned long, long long, unsigned long long>;
+
+TYPED_TEST_SUITE(BitOpsTemplateAllIntTestsFixture, BitOpsAllIntTypes);
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, BitOpsCountOneBits) {
+    EXPECT_EQ(countOneBits(this->m_countOneBitVal), 2);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, BitOpsFirstOneBits) {
+    EXPECT_EQ(firstOneBit(this->m_firstOneBitVal), 0);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, BitOpsPopFirstOneBits) {
+    EXPECT_EQ(popFirstOneBit(&(this->m_popFirstOneBitVal)), 0);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, log2Down) {
+    EXPECT_EQ(log2Down(this->m_log2DownVal), -1);
+    this->m_log2DownVal = 1;
+    EXPECT_EQ(log2Down(this->m_log2DownVal), 0);
+    this->m_log2DownVal = 2;
+    EXPECT_EQ(log2Down(this->m_log2DownVal), 1);
+}
+
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, log2Up) {
+    EXPECT_EQ(log2Up(this->m_log2UpVal), 0);
+    this->m_log2UpVal = 1;
+    EXPECT_EQ(log2Up(this->m_log2UpVal), 0);
+    this->m_log2UpVal = 2;
+    EXPECT_EQ(log2Up(this->m_log2UpVal), 1);
+    this->m_log2UpVal = 3;
+    EXPECT_EQ(log2Up(this->m_log2UpVal), 2);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, pow2Up){
+    EXPECT_EQ(pow2Up(this->m_power2UpVal), 1);
+    this->m_power2UpVal = 1;
+    EXPECT_EQ(pow2Up(this->m_power2UpVal), 1);
+    this->m_power2UpVal = 2;
+    EXPECT_EQ(pow2Up(this->m_power2UpVal), 2);
+    this->m_power2UpVal = 3;
+    EXPECT_EQ(pow2Up(this->m_power2UpVal), 4);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, pow2Down){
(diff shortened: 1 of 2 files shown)


peizhang56:
Screenshot 2025-07-21 100955

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
+#include "bitops.h"
```
demandal25, line 7: copyright?

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include "gtest/gtest.h"
```
demandal25, line 8: @atulkulk what's the order of inclusions are we following

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+protected:
+    int m_size{0};
+    uint32_t m_u32Val{2};
+    uint8_t m_u8Val{2};
+    std::vector<uint32_t> m_u32Vec{2, 4, 8, 16, 32, 64, 128, 256};
+    std::vector<float> m_floatVec{2.0f};
+};
+
+
+TEST_F(BitOpsTestsFixture, DIVUP_Tests) {
```
demandal25: why do these need any fixture?

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
+#include "bitops.h"
+#include "gtest/gtest.h"
+
+class BitOpsTestsFixture : public ::testing::Test {
+public:
+    BitOpsTestsFixture() = default;
+    ~BitOpsTestsFixture() override = default;
```
demandal25: w/o any setup, why do we need a fixture? They can be just `TEST` I think. 
Second, assemble the ones that actually need the fixture, and put the fixture just before them

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+protected:
+    int m_size{0};
+    uint32_t m_u32Val{2};
+    uint8_t m_u8Val{2};
+    std::vector<uint32_t> m_u32Vec{2, 4, 8, 16, 32, 64, 128, 256};
+    std::vector<float> m_floatVec{2.0f};
```
demandal25: @atulkulk  what is our naming convention for the member variables?

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+TEST_F(BitOpsTestsFixture, ROUNDUP_Tests) {
+    EXPECT_EQ(ROUNDUP(0, 8), 0);
+    EXPECT_EQ(ROUNDUP(1, 8), 8);
+    EXPECT_EQ(ROUNDUP(7, 8), 8);
+    EXPECT_EQ(ROUNDUP(8, 8), 8);
+    EXPECT_EQ(ROUNDUP(9, 8), 16);
+    EXPECT_EQ(ROUNDUP(15, 8), 16);
+    EXPECT_EQ(ROUNDUP(16, 8), 16);
```
demandal25, line 28: the definition from gtest is actually `EXPECT_EQ(expected_value, actual_value)`, so need to reverse the params I think

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+}
+
+TEST_F(BitOpsTestsFixture, ALIGN_POWER_Tests) {
+    EXPECT_EQ(ALIGN_POWER(7, 8), 8);
+    EXPECT_EQ(ALIGN_POWER(8, 8), 8);
+    EXPECT_EQ(ALIGN_POWER(9, 8), 16);
+}
+
+TEST_F(BitOpsTestsFixture, ALIGN_SIZE_Tests) {
+    m_size = 1;
```
demandal25: why does it have to be a member variable? just a local variable should suffice, I think.

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+    EXPECT_EQ(alignUp(17, 8), 24);
+}
+
+
+TEST_F(BitOpsTestsFixture, u32fp8MaxValue) {
+    EXPECT_EQ(u32fp8MaxValue(), 0xf0000000);
+}
+
+TEST_F(BitOpsTestsFixture, u32fp8Encode) {
+    uint8_t expected = 2;
```
demandal25: why are you suddenly declaring `expected` variable here? and not before?

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+    EXPECT_EQ(alignUp(7, 8), 8);
+    EXPECT_EQ(alignUp(8, 8), 8);
+    EXPECT_EQ(alignUp(9, 8), 16);
+    EXPECT_EQ(alignUp(15, 8), 16);
+    EXPECT_EQ(alignUp(16, 8), 16);
+    EXPECT_EQ(alignUp(17, 8), 24);
+}
+
+
+TEST_F(BitOpsTestsFixture, u32fp8MaxValue) {
```
demandal25: based on previous naming you used, here and below!
```suggestion
TEST_F(BitOpsTestsFixture, u32fp8MaxValue_Tests) {
```

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+    T m_log2UpVal{0};
+    T m_power2UpVal{0};
+    T m_power2DownVal{1};
+};
+
+using BitOpsAllIntTypes = ::testing::Types<int, unsigned int, long, unsigned long, long long, unsigned long long>;
+
+TYPED_TEST_SUITE(BitOpsTemplateAllIntTestsFixture, BitOpsAllIntTypes);
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, BitOpsCountOneBits) {
```
demandal25: again, naming style `_Tests`

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,225 @@
...
+    T m_countOneBitVal{3};
+    T m_firstOneBitVal{3};
+    T m_popFirstOneBitVal{3};
+    T m_log2DownVal{0};
+    T m_log2UpVal{0};
+    T m_power2UpVal{0};
+    T m_power2DownVal{1};
+};
+
+using BitOpsAllIntTypes = ::testing::Types<int, unsigned int, long, unsigned long, long long, unsigned long long>;
```
demandal25: nice :)

demandal25 (commented):
A quick pass!

test/BitOpsTests.cpp
peizhang56: Include should be something like this
https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
I would say gtest/gmock should come later than the major code

test/BitOpsTests.cpp
peizhang56: I think it is a matter of style. 
Most of the cases, TestFixture is helping to removing duplicate setups.
But since this is a simple header file. It should not matter to use TEST_F/TEST. 
Unless rccl is very picky on writing unit test

test/BitOpsTests.cpp
peizhang56: Overall TestFixture helps to declare common variables/setups that can be shared across the test suite.
Here, since we are not doing too many repeated test cases, so it works fine if we don't use TestFixtures. 
Since this is a Unit Test, it does not bother to me if TestFixtures are used. As most of the cases, we should use them to avoid duplicated UT code.

test/BitOpsTests.cpp
peizhang56: Refactored

test/BitOpsTests.cpp
peizhang56: refactored to match the style

test/BitOpsTests.cpp
peizhang56: Refactored.

test/BitOpsTests.cpp
peizhang56: refactored. Because this test requires the type to be uint8_t.

test/BitOpsTests.cpp
peizhang56: Refactored to match the existing style

test/BitOpsTests.cpp
peizhang56: While modern versions of Google Test (gtest) treat the arguments to EXPECT_EQ symmetrically (meaning the order doesn't affect the comparison itself), historically, it was conventional to use EXPECT_EQ(expected, actual).

peizhang56: https://google.github.io/googletest/reference/assertions.html#EXPECT_EQ
It is val1, val2 now, not expected, actual

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,243 @@
+/*************************************************************************
+ * Copyright (c) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include "bitops.h"
+#include "gtest/gtest.h"
+
+namespace RcclUnitTesting
```
demandal25, line 10: @atulkulk  qq, I see this namespace in other test files as well. Why do we need a namespace for the test files? They are not bundled as part of the release. I don't see where we need to avoid and naming collision!

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+    EXPECT_EQ(pow2Up(this->valX_), 1);
+    this->valX_ = 1;
+    EXPECT_EQ(pow2Up(this->valX_), 1);
+    this->valX_ = 2;
+    EXPECT_EQ(pow2Up(this->valX_), 2);
+    this->valX_ = 3;
+    EXPECT_EQ(pow2Up(this->valX_), 4);
+}
+
+TYPED_TEST(BitOpsTemplateAllIntTestsFixture, pow2Down){
```
demandal25:
```suggestion
TYPED_TEST(BitOpsTemplateAllIntTestsFixture, pow2DownSucceess){
```

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+    this->valX_ = 2;
+    EXPECT_EQ(pow2Down(this->valX_), 2);
+    this->valX_ = 3;
+    EXPECT_EQ(pow2Down(this->valX_), 2);
+}
+
+template <typename T>
+class BitOpsTemplateUnsignedTestsFixture : public testing::Test {
+public:
+    BitOpsTemplateUnsignedTestsFixture() = default;
```
demandal25: we don't need to explicitly mention the default ones...right?

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+protected:
+    T reverseSubBits_{2};
+    T reverseBits_{2}; // 0b10
+};
+
+using BitOpsUnsignedTypes = ::testing::Types<unsigned int, unsigned long, unsigned long long>;
+
+TYPED_TEST_SUITE(BitOpsTemplateUnsignedTestsFixture, BitOpsUnsignedTypes);
+
+TYPED_TEST(BitOpsTemplateUnsignedTestsFixture, reverseSubBits) {
```
demandal25:
```suggestion
TYPED_TEST(BitOpsTemplateUnsignedTestsFixture, reverseSubBitsSuccess) {
```
below one too!

test/BitOpsTests.cpp
```diff
@@ -0,0 +1,243 @@
...
+    EXPECT_EQ(ROUNDUP(1, 8), 8);
+    EXPECT_EQ(ROUNDUP(7, 8), 8);
+    EXPECT_EQ(ROUNDUP(8, 8), 8);
+    EXPECT_EQ(ROUNDUP(9, 8), 16);
+    EXPECT_EQ(ROUNDUP(15, 8), 16);
+    EXPECT_EQ(ROUNDUP(16, 8), 16);
+    EXPECT_EQ(ROUNDUP(17, 8), 24);
+}
+
+TEST(ALIGN_POWER, ALIGN_POWERSuccess) {
```
demandal25, line 32: nit: so the `ALIGN_POWERSuccess` part is actually a verb as per the gtest docs, so probably better to name it like `ALIGN_POWERSucceeds`

demandal25 (commented):
final pass!

test/BitOpsTests.cpp
demandal25: I usually prefer to use test fixture only if needed. No extra work! It's C++ :)

test/BitOpsTests.cpp
demandal25: yes, I follow this too, but would like to confirm with Atul that we are following it throughout RCCL

test/BitOpsTests.cpp
peizhang56: I think the name is self explanatory

test/BitOpsTests.cpp
peizhang56: namespace should be fine to add this. Is there any downside to add namespace?

test/BitOpsTests.cpp
peizhang56: No, compiler will add the one for me if not defined. I usually don't mind if it is Unit Test

test/BitOpsTests.cpp
peizhang56: I usually don't mind if it is Unit Test.

test/BitOpsTests.cpp
demandal25: nit: I would recommend removing the unnecessary code unless they really help to understand the code.

test/BitOpsTests.cpp
demandal25: nit: again, I usually don't recommend adding anything if that is unnecessary, less code, less error!

peizhang56: namespace is designed for customized scopes. 
I am not against adding user customized scopes in UT. I usually don't do this. But I don't think existing rccl code did anything  unreasonable.

test/BitOpsTests.cpp
peizhang56: True, I think syntax-wise, it should not matter. I also think there is no right or wrong in this case. Compiler will understand it anyways. 
I put it there to make it in parity with destructor. But I removed it anyways.

demandal25 (approved):

test/BitOpsTests.cpp
atulkulk: The scope differentiates that the tests were written specifically for RCCL.

test/BitOpsTests.cpp
atulkulk: Following Google style guide for RCCL UT is fine.

test/CMakeLists.txt
```diff
@@ -9,6 +9,11 @@ if(BUILD_TESTS)
 
   message("Building rccl unit tests (Installed in /test/rccl-UnitTests)")
 
+  if (ENABLE_CODE_COVERAGE)
```
atulkulk, line 12: Instead of adding the flags again, how about just inheriting those flags from the main rccl target? The options will be in sync with the main library for the test binaries. Something like below which can go after line 121 in the for loop:

```
  get_target_property(RCCL_COMPILE_OPTIONS rccl COMPILE_OPTIONS)
  if(RCCL_COMPILE_OPTIONS)
    target_compile_options(${target} PRIVATE ${RCCL_COMPILE_OPTIONS})
  endif()

  get_target_property(RCCL_INTERFACE_COMPILE_OPTIONS rccl INTERFACE_COMPILE_OPTIONS)
  if(RCCL_INTERFACE_COMPILE_OPTIONS)
    target_compile_options(${target} PRIVATE ${RCCL_INTERFACE_COMPILE_OPTIONS})
  endif()
```

peizhang56: This won't work. 
I am not sure why, but it did not generate the expected coverage. I believe the header was rebuilt without cov flags
image

peizhang56: @atulkulk I tried a couple of times, not sure why, but looks bitops.h could be coming from hipify from the unit test, adding the flags to the target did not work.
However, the current way should not matter, as it only enables when coverage flags is on. Should not affect other builds

atulkulk (approved):
Looks good to me. I would wait for the CI to complete before merging.

venksubramd:
The tests use EXPECT_ instead of ASSERT_. This allows us to place multiple independent checks in one test. With EXPECT, even if one check fails, the following checks will run. In the case of ASSERT, the failure terminates the test right away.

From the above point of view, using EXPECT here is good instead of using ASSERT. Also, in this particular test suite, the tests are fairly straightforward. So, using EXPECT works really well here, nice job. However, it is better not to use EXPECT and write multiple tests to keep independent checks in separate tests for more complex tests. The tests here are nice, but we should use caution from overly using multiple EXPECTs in tests.

TEST(DIVUP, DIVUPSuccess) {
Do we need to care about 0 denominator? If we have a specific expectations about the behavior of the function when denominator is 0, a test for that will ensure that behavior is maintained in the future.

TYPED_TEST(BitOpsTemplateAllIntTestsFixture, divUpSuccess) {
Same as above, do we need to capture the expected behavior for 0 denominator?

Do we need tests for alignDown?

Do we need more tests for countOneBits?

Do we need more tests for firstOneBit?

Do we need tests for u32fpEncode? This function is being called from u32fp8Decode which has a test, but wondering if we need tests directly on u32fpEncode?

Likewise, do we need tests for u32fpDecode?

The eatHash functions are being called from getHash. Do we need tests directly on eatHash? Probably not, but another pair of eyes looking into it would be good.

Thank you.

peizhang56:
@venksubramd 
Thx for your comment.
1. I agree with you, but I would prefer EXPECT_ for easy test cases, because it helped to run all the tests. Running CI/CD pipeline in rccl takes long time, we want to run as many tests as we want in that launch. Otherwise, imagine fixing 1 failure, then rebuild, then another one comes up.
2. I tried the divide by zero test cases. I wanted to use ASSERT_DEATH, EXPECT_THROW, however, looks it is not causing any crash or exception. But compiler warning shows undefined behavior on divide by 0. I think we don't want a test for undefined behavior.
3. alignDown added with a couple more, right now it is 100% coverage with 144 conditions
4. countOneBits is missing 1 condition because in my env, sizeof(unsigned long) and sizeof(unsigned long long) are all 8, so it cannot cover all conditions because in this case, these are unreachable code. I also tried to add __ int128, but static_assert will block it from compiling. See image below for the unreachable code.
5. firstOneBit: the same as above
6. u32fpEncode/u32fpDecode/eatHash added. Originally, I thought 80% was the goal, but yes, let's make it 100%
Screenshot 2025-07-25 163606
image

venksubramd:
@peizhang56 

Thank you, I agree with your comments.

Sorry if I am not looking correctly, but I do not see the tests for 

| 6. u32fpEncode/u32fpDecode/eatHash added. Originally, I thought 80% was the goal, but yes, let's make it 100%

Could you please check if those have been committed?

Thanks.

peizhang56:
> @peizhang56
> 
> Thank you, I agree with your comments.
> 
> Sorry if I am not looking correctly, but I do not see the tests for
> 
> | 6. u32fpEncode/u32fpDecode/eatHash added. Originally, I thought 80% was the goal, but yes, let's make it 100%
> 
> Could you please check if those have been committed?
> 
> Thanks.

Yes, they were in this commit. 
image

venksubramd:
@peizhang56

Thank you. I see those now.

Looks good. We can move to close this.

rahulvaidya20 (approved):
LGTM

PR #1828, opened 2025-07-28
Optimize alltoall for 64 GPUs and above for gfx942

mustafabar:
## Details

**Work item:** _Internal_

**What were the changes?**  
Tune large scale alltoall for MI300x by increasing the P2P chunk size alongside PXN enablement

Observed that the NET P2P chunk size increase to 512K is effective but only from 64 GPUs **and with PXN enabled** 

Overwrite existing defaults from 64 GPUs and within RCCL only path

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/include/rccl_common.h
+++ b/src/include/rccl_common.h
@@ -41,6 +41,9 @@ typedef enum RcclTunableColls {
 #define RCCL_PROTOCOL_FACTOR_IDX 2
 #define RCCL_PROTOCOL_THREAD_THRESHOLD_IDX 3
 
+#define RCCL_VALUE_UNSET -2
+#define RCCL_VALUE_INVALID -1
+
 #ifdef RCCL_EXPOSE_STATIC
 #define RCCL_STATIC_EXPOSE_CHECK()
 #else
@@ -78,7 +81,8 @@ void rcclUpdateThreadThreshold(struct ncclComm* comm, size_t const& nBytes, stru
 ncclResult_t rcclGetAlgoInfo(struct ncclComm* comm, ncclFunc_t coll, uint64_t count, ncclDataType_t dataType,
                              int collNetSupport, int nvlsSupport, int numPipeOps,
                              int* algo, int* protocol, int* maxChannels);
-
+void rcclSetPxn(struct ncclComm* comm,  int& rcclPxnDisable);
+void rcclSetP2pNetChunkSize(struct ncclComm* comm,  int& rcclP2pNetChunkSize);
 ncclResult_t rcclFuncMaxSendRecvCount(ncclFunc_t func, int nRanks, size_t count, size_t& maxCount);
 ncclResult_t commSetUnrollFactor(struct ncclComm* comm);
 #endif
--- a/src/include/rccl_common.h
+++ b/src/include/rccl_common.h
@@ -41,8 +41,10 @@ typedef enum RcclTunableColls {
 #define RCCL_PROTOCOL_FACTOR_IDX 2
 #define RCCL_PROTOCOL_THREAD_THRESHOLD_IDX 3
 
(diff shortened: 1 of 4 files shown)


wenkaidu (approved):

src/rccl_wrap.cc
```diff
@@ -120,6 +120,31 @@ ncclResult_t rcclGetAlgoInfo(struct ncclComm* comm, ncclFunc_t coll, uint64_t co
   return ncclSuccess;
 }
 
+void rcclSetPxn(struct ncclComm* comm,  int& rcclPxnDisable) {
+  static int pxnDisable = RCCL_VALUE_UNSET;
+  if(pxnDisable == RCCL_VALUE_UNSET) {
+    const char *inputStr = getenv("NCCL_PXN_DISABLE");
```
alex-breslow-amd, line 126: Question here.  Why not use the builtin NCCL_PARAM logic since NCCL_PXN_DISABLE is already defined that way?  The way it currently is if I'm reading this right, even setting it 0 will disable where a value of 0 should mean it is enabled.  This seems like a potential footgun.

Let me know if I am misunderstanding something.

src/rccl_wrap.cc
```diff
@@ -120,6 +120,31 @@ ncclResult_t rcclGetAlgoInfo(struct ncclComm* comm, ncclFunc_t coll, uint64_t co
...
+    pxnDisable = (comm->nRanks >= 64)? 0 : 1;
+    INFO(NCCL_INIT, "RCCL PXN set as %s", !pxnDisable? "enabled" : "disabled");
+  }
+  rcclPxnDisable = pxnDisable;
+}
+
+void rcclSetP2pNetChunkSize(struct ncclComm* comm,  int& rcclP2pNetChunkSize) {
+  static int p2pNetChunkSize = RCCL_VALUE_UNSET;
+  if(p2pNetChunkSize == RCCL_VALUE_UNSET) {
+    const char *inputStr = getenv("NCCL_P2P_NET_CHUNKSIZE");
```
alex-breslow-amd, line 140: Same comment here.

src/include/rccl_common.h
```diff
@@ -41,6 +41,9 @@ typedef enum RcclTunableColls {
 #define RCCL_PROTOCOL_FACTOR_IDX 2
 #define RCCL_PROTOCOL_THREAD_THRESHOLD_IDX 3
 
+#define RCCL_VALUE_UNSET -2
+#define RCCL_VALUE_INVALID -1
+
```
alex-breslow-amd: I'm wondering if this should instead be defined as a enum struct.  We don't do this in the rest of the code, but it prevents these file local parameters from leaking into the global namespace.  This is a broader comment that extends beyond this diff.  It's a problem in general with RCCL/NCCL.

src/rccl_wrap.cc
mustafabar: I need to get communicator's level information before changing the value. NCCL_PARAM uses compile time constexpr to set default.
Regarding the second point on 0 disabling PXN, not sure I understand the problem. Did you catch a bug? from 64 GPUs, I want it to enable PXN at line 131 (and that happens by setting the value to 0 due to current semantics of double negation)

src/include/rccl_common.h
mustafabar: good point, will do

alex-breslow-amd (approved):

src/rccl_wrap.cc
dmwu: I think we should avoid double negation as much as possible to improve code readability, e.g., it's more straightforward to see pxnEnable=true, rather than implying that through pxnDisable=false

PR #1841, opened 2025-08-02
changes for hugepage backed host buffer for larger allocations

bmital:
## Details

**What were the changes?**  
changes are related to host buffer allocation being backed by hugepages instead of 4K pages

With model training (at higher values of PP and TP) the number of page table entries could explode due to the high number of 4K page backed MRs. Today RCCL uses 4K pages for all MRs that are allocated on the host memory using the hipHostMalloc() call. There are several 4MB host buffers (one per channel per comm) allocated and with a pagesize of 4K, this would result in 1K PTEs. A change was made to do the allocation via hugepage mmap calls and register using hipHostRegister followed by hipHostGetDevicePointer to get the pointer to the huge page backed buffer. This did not give us any benefit for RCCL sweep test but cut the PTE consumption for model run by quite a bit. Here's the comparison :

With current allocation:
```
--------------------------
MR Count: 548
Total 4K Pages: 53052
Total 2MB Pages: 314
Total PTE Entries: 53366
```
 
With Hugepage backed allocation (PP=TP=4):
```
----------------------
MR Count: 548
Total 4K Pages: 3900
Total 2MB Pages: 410
Total PTE Entries: 4310
```
 
Note: If the hugepage allocation fails then the allocation will fall back to hipHostMalloc()

A change was made to do the allocation via hugepage mmap calls and register using hipHostRegister followed by hipHostGetDevicePointer to get the pointer to the huge page backed buffer. 

**Additional Documentation:**  
The RCCL sweep tests for all collectives are at par with baseline numbers, after this change (as tested so far on 2-node setup with 8 nics/gpus each).

## Approval Checklist

### DIFF
--- a/src/include/alloc.h
+++ b/src/include/alloc.h
@@ -18,6 +18,10 @@
 #include <stdlib.h>
 #include <string.h>
 #include "rccl_vars.h"
+#include <unordered_map>
+#include <mutex>
+
+#define RCCL_HP_MIN_SIZE 2097152
 
 #if CUDART_VERSION >= 11030
 #include <cuda.h>
@@ -30,6 +34,9 @@ template<typename T>
 constexpr size_t ncclSizeOfT() { return sizeof(T); }
 template<>
 constexpr size_t ncclSizeOfT<void>() { return 1; }
+
+extern std::unordered_map<void*, size_t> hugepageAllocs;
+extern std::mutex hugepageAllocsMutex;
 
 #if CUDART_VERSION >= 12020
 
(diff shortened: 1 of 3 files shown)


bmital:
@nileshnegi @wenkaidu updated the details above to better explain the nature of the changes in the PR.

src/include/alloc.h
```diff
@@ -136,12 +135,64 @@ ncclResult_t ncclCudaHostCallocDebug(T** ptr, size_t nelem, const char *filefunc
...
+      if (size < 2097152) {
+        WARN("small size : forcing back to hipHostMalloc");
+        CUDACHECKGOTO(hipHostMalloc(ptr, size, cudaHostAllocMapped), result, finish);
+      } else {
+	      // Hugepage allocation via mmap
+	      void* hostPtr = mmap(NULL, size, PROT_READ | PROT_WRITE,
+			           MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB, -1, 0);
+	      if (hostPtr == MAP_FAILED) {
+		      WARN("Hugepage allocation failed. Falling back to hipHostMalloc");
+		      CUDACHECKGOTO(hipHostMalloc(ptr, size, cudaHostAllocMapped), result, finish);
```
wenkaidu: need hipHostMallocUncached to work on gfx950

src/include/alloc.h
```diff
@@ -136,12 +135,64 @@ ncclResult_t ncclCudaHostCallocDebug(T** ptr, size_t nelem, const char *filefunc
...
+#else
+      CUDACHECKGOTO(hipExtMallocWithFlags((void**)ptr, size, hipDeviceMallocFinegrained), result, finish);
+#endif
+    } else {
+      if (size < 2097152) {
+        WARN("small size : forcing back to hipHostMalloc");
+        CUDACHECKGOTO(hipHostMalloc(ptr, size, cudaHostAllocMapped), result, finish);
+      } else {
+	      // Hugepage allocation via mmap
+	      void* hostPtr = mmap(NULL, size, PROT_READ | PROT_WRITE,
```
wenkaidu, line 163: This path cannot work on gfx950, need hipHostMallocUncached flag

src/include/alloc.h
```diff
@@ -136,12 +135,64 @@ ncclResult_t ncclCudaHostCallocDebug(T** ptr, size_t nelem, const char *filefunc
...
+			           MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB, -1, 0);
+	      if (hostPtr == MAP_FAILED) {
+		      WARN("Hugepage allocation failed. Falling back to hipHostMalloc");
+		      CUDACHECKGOTO(hipHostMalloc(ptr, size, cudaHostAllocMapped), result, finish);
+	      } else {
+		      memset(hostPtr, 0, size);
+		      CUDACHECKGOTO(hipHostRegister(hostPtr, size, hipHostRegisterMapped), result, finish);
```
wenkaidu: need hipHostMallocUncached to work on gfx950

alex-breslow-amd:
This is just a general comment, but @bmital, do we have any performance data for this?  On paper, this should be good, but there are some subtleties of using huge pages on NUMA systems.  From your comment in the description, it appears that there is no performance uplift from doing this.  I just want to verify that is indeed correct.

bmital:
> This is just a general comment, but @bmital, do we have any performance data for this? On paper, this should be good, but there are some subtleties of using huge pages on NUMA systems. From your comment in the description, it appears that there is no performance uplift from doing this. I just want to verify that is indeed correct.

hi @alex-breslow-amd agree with you. The gain here is not in performance rather in the PTE entry savings which can be a table entry allocated on the nic or the host (based on implementation) The PTE entry savings are considerable for llama model runs at higher scale of parallelism (as contrasted above in the PR description : data was collected with TP=PP=4 and num_iterations=10).

src/include/alloc.h
```diff
@@ -136,12 +135,83 @@ ncclResult_t ncclCudaHostCallocDebug(T** ptr, size_t nelem, const char *filefunc
   return result;
 }
 
+template <typename T>
+ncclResult_t ncclCudaHostCallocDebug2(T** ptr, size_t nelem, int hp_request, const char *filefunc, int line) {
```
vijasrin: move this inside an env var and merge with ncclCudaHostCallocDebug

src/include/alloc.h
```diff
@@ -136,12 +135,83 @@ ncclResult_t ncclCudaHostCallocDebug(T** ptr, size_t nelem, const char *filefunc
...
+  if (nelem > 0) {
+    if (managed) {
+#if defined(HIP_UNCACHED_MEMORY)
+      CUDACHECKGOTO(hipExtMallocWithFlags((void**)ptr, size, hipDeviceMallocUncached), result, finish);
+#else
+      CUDACHECKGOTO(hipExtMallocWithFlags((void**)ptr, size, hipDeviceMallocFinegrained), result, finish);
+#endif
+    } else {
+      if (hp_request) {
+	      if (size < 2097152) {
```
vijasrin: add a #define for 2MB size

src/transport/net.cc
```diff
@@ -1153,7 +1160,11 @@ static ncclResult_t sendProxyFree(struct ncclProxyConnection* connection, struct
     }
     struct connectMapMem* mems = resources->map.mems;
     if (resources->map.sameProcess) {
-      NCCLCHECK(ncclCudaHostFree(mems[NCCL_NET_MAP_HOSTMEM].cpuPtr));
+      if (((mems[NCCL_NET_MAP_HOSTMEM].size)*(sizeof(int))) < 2097152) {
```
vijasrin: we can keep the changes localized to alloc.h with some bookeeping for mmap and munmap

src/include/alloc.h
bmital: done

src/include/alloc.h
bmital: done

src/transport/net.cc
bmital: done

src/include/alloc.h
bmital: done

src/include/alloc.h
bmital: done

src/include/alloc.h
```diff
@@ -105,43 +112,97 @@ static inline ncclResult_t ncclCuMemHostFree(void* ptr) {
...
 
-static inline ncclResult_t ncclCudaHostFree(void* ptr) {
+
+static inline ncclResult_t ncclCudaHostFree(void* ptr, size_t alloc_size=0) {
+  if (alloc_size >= 0) { 
+    std::lock_guard<std::mutex> lock(hugepageAllocsMutex);
+    for (auto &kv : hugepageAllocs) INFO(NCCL_ALLOC, "Hugepage alloc ptr %p size %zu", kv.first, kv.second);
+    auto it = hugepageAllocs.find(ptr);
+    if (it != hugepageAllocs.end()) {
+      INFO(NCCL_ALLOC, "%s:%d Cuda Host HugePage unmap size %ld pointer %p app_tracked_size %ld", __FILE__, __LINE__, it->second, ptr, alloc_size);
```
vijasrin: Call hipHostUnregister also ?
https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___memory.html#ga4c9e1810b9f5858d36c4d28c91c86924

bmital: good point. Done

vijasrin (approved):
lgtm

wenkaidu (approved):

PR #1857, opened 2025-08-12
Have ncclDevFuncId use 64-Bit keyed map with field packing

mustafabar:
## Details

**Work item:** Internal

**What were the changes?**  
Refactored `ncclDevFuncId` to use a 64-bit key and an `std::unordered_map` (`ncclDevFuncNameToId`) instead of row-based indexing; updated `generate.py` to emit the map and added 4-bit field packing with per-collective special cases.

- To decouple lookup from fragile row math, simplify maintenance when adding/removing functions or customizations. This reduces risk from off-by-row errors.
- **This will facilitate pipelining only specific datatypes (e.g. `bfloat16`), which is WIP**
- Look-up is at least 2x faster than the existing scheme on average

- Introduced 64-bit keys packing (coll, algo, proto, devRedOp, type) into 4-bit fields (`RCCL_FUNC_ID_MASK=0xF`, `RCCL_<FIELD>_SHIFT`). 
- Added special handling: `Broadcast`: key uses (coll | (proto<<8)). `SendRecv`, `AllToAllPivot`: key uses just coll. 
- Replaced `ncclDevFuncRowToId` with a generated `unordered_map<uint64_t,int>` and changed `device.h` to consume it. 
- `generate.py` now asserts ≤16 values per field, computes indices, builds keys, and emits the unordered_map with comments to document bit layout.

**Additional Documentation:**  
- Keep all_colls/all_redops/all_tys/all_protos/all_algos ordering in generate.py in sync with enums in the code; the generator enforces capacity (4-bit) limits. 
- If more than 16 values are ever needed for a field, bump mask/shift and the generator (and `device.h`) accordingly.

## Approval Checklist


src/include/device.h
```diff
@@ -687,74 +693,42 @@ inline bool ncclNvlsSupported(int devRedOp, int type) {
...
+          ((uint64_t)(proto    & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT*2));
+  } else if (coll == ncclFuncSendRecv || coll == ncclFuncAllToAllPivot) {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK));
+  } else {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK)) |
+          ((uint64_t)(algo     & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT))   |
+          ((uint64_t)(proto    & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT*2)) |
+          ((uint64_t)(devRedOp & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT*3)) |
+          ((uint64_t)(type     & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT*4));
+  }
```
alex-breslow-amd: I have a concern about these hardcoded shifts.  Maybe it would be better to make these variables like RCCL_PROTO_SHIFT, RCCL_ALGO_SHIFT, RCCL_DEVREDOP_SHIFT, and RCCL_TYPE_SHIFT.  Additionally, I think you could probably encapsulate these shifting and masking operations as part of a macro or templated C++ function like the following:
#define compute_key(coll, algo, proto, devRedOp, type) (lines 712 to 716 go here)
That would eliminate the all of the if then else logic too.  Some arguments would just be zero.

src/include/device.h
```diff
@@ -687,74 +693,42 @@ inline bool ncclNvlsSupported(int devRedOp, int type) {
...
+  auto it = ncclDevFuncNameToId.find(key);
```
alex-breslow-amd, line 722: I have some concerns about how compact this really is and how expensive this added hashing logic will be.  I'd recommend benchmarking this and checking if the amount of register spills increases from this.

src/include/device.h
```diff
@@ -687,74 +693,42 @@ inline bool ncclNvlsSupported(int devRedOp, int type) {
...
+  // Pack 4-bit fields from right (LSB) to left in order:
+  // coll, algo, proto, devRedOp, type
+  // This logic must be in sync with the key generation logic in generate.py
+  if (coll == ncclFuncBroadcast) {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK)) |
+          ((uint64_t)(proto    & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT*2));
+  } else if (coll == ncclFuncSendRecv || coll == ncclFuncAllToAllPivot) {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK));
+  } else {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK)) |
```
alex-breslow-amd: A general question here is why not just have an added layer of indirection.  For instance, if the dimension for broadcast is only coll x proto, why not just index into a broadcast table.  For allToAllPivot/sendRecv, why not just index into an allToAllPivot Table, everything else can index into the big table I guess.  That would get rid of needing to have a hash map.  The key would just be the table index.

alex-breslow-amd: Note, that if you use a pointer to a pointer, this should just work.  First pointer determines which table to index into, second pointer points to the front of the table.  That would get rid of this control flow logic at the cost of being a bit more opaque.

src/include/device.h
```diff
@@ -687,74 +693,42 @@ inline bool ncclNvlsSupported(int devRedOp, int type) {
...
+          ((uint64_t)(algo     & RCCL_FUNC_ID_MASK) << (RCCL_FUNC_ID_SHIFT))   |
```
alex-breslow-amd: I'm wondering if the bitfield builtins in C/C++ might work here.  Specifically, you can have some like the following:
union ConversionUnion{
    struct Foo as_foo;
    uint64_t as_u64;
};

// This syntax might not be exactly right.
struct Foo{
    int: 4 algo;
    int: 4 proto;
    int: 4 devRedOp;
    int: 4 type;
};

By using as_u64, this gets rid of having to do the explicit shifting yourself.  The compiler does it.  You can use a static_assert regarding the size of ConversionUnion if you are concerned about it being greater than 4 bytes or 8 bytes in size.

src/include/device.h
mustafabar: Sure, I can do that

src/include/device.h
mustafabar: Note this is called on the host, however, based on my benchmarks, it is at least 2x faster than row index calc approach

src/include/device.h
mustafabar: That's possible, but not sure if it is worth it. It's a tradeoff between generality and table size. 
The hashtable look-up is amortized constant, and the corner cases may change in the future as we add more algorithms and may require other fields to be covered. 
As a matter of fact, we are debating where we should support protocols for send/recv. What I agree with though is that we reserve redop for non reduction collectives, which we may want to change

src/include/device.h
mustafabar: On the macro point, we still need the if-statements to set the ignored fields to zero. If we just pass the fields as is for bcast, for example, we can have issues when we get non-zero inputs for ignored fields

src/include/device.h
mustafabar: Thanks for the suggestion. I considered bit-fields, but since their layout is implementation-defined and can vary with compiler, ABI, and endianness, I opted for explicit shifts. This guarantees the key encoding stays in sync with generate.py [python generator] across platforms. We could explore bit-fields if we lock the compiler/ABI and add strict CI layout checks, but for now I think the shift-based packing is safer.

alex-breslow-amd (approved):
Thanks.  LGTM.

src/include/device.h
```diff
@@ -687,74 +697,42 @@ inline bool ncclNvlsSupported(int devRedOp, int type) {
...
+          ((uint64_t)(proto    & RCCL_FUNC_ID_MASK) << RCCL_PROTO_SHIFT);
+  } else if (coll == ncclFuncSendRecv || coll == ncclFuncAllToAllPivot) {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK) << RCCL_COLL_SHIFT );
+  } else {
+    key = ((uint64_t)(coll     & RCCL_FUNC_ID_MASK) << RCCL_COLL_SHIFT ) |
+          ((uint64_t)(algo     & RCCL_FUNC_ID_MASK) << RCCL_ALGO_SHIFT ) |
+          ((uint64_t)(proto    & RCCL_FUNC_ID_MASK) << RCCL_PROTO_SHIFT) |
+          ((uint64_t)(devRedOp & RCCL_FUNC_ID_MASK) << RCCL_REDOP_SHIFT) |
+          ((uint64_t)(type     & RCCL_FUNC_ID_MASK) << RCCL_DTYPE_SHIFT);
+  }
```
alex-breslow-amd, line 721: How critical is it for these to use only 4 bits?  Why not just use 8 bits for each field?  If 4 bits is required, then you can use a struct like this (note, I haven't tested this so there may be a few syntax mistakes):
```
struct Key{
  uint32_t: 4  coll {0};
  uint32_t: 4 algo {0};
  uint32_t: 4 proto {0};
  uint32_t: 4 devRedOp {0};
  uint32_t: 4 type {0};
  
  uint32_t as_u32(){
    return (coll << RCCL_COLL_SHIFT) | (algo << RCCL_ALGO_SHIFT) | (proto << RCCL_PROTO_SHIFT) | (devRedOp << RCCL_REDOP_SHIFT) | (type << RCCL_DTYPE_SHIFT));
  }
};

static_assert(sizeof(Key) == 4, "Key type must be 4 bytes");
// I think this is the right syntax.  You can also do per field.
uint32_t key = (Key {coll, algo, proto, devRedOp, type}).as_u32();
```
Note, here the compiler does the masking operation for you.  

I think you can take this one step further by using a union datatype:
```
union KeyUnion{
  Key as_key;
  uint32_t as_u32;
};

static_assert(sizeof(KeyUnion) == 4, "KeyUnion must be 4 bytes");
```
Then you can even compute the key implicitly without defining the shifts, no need for the as_u32 function in Key.

I only mention this because I think this code is easier to read and maintain.

PedramAlizadeh (approved):

src/include/device.h
mustafabar: @alex-breslow-amd Using 4 bits is to make us accommodate as many fields as possible within 64 bits before we run out of field space. It is just a defensive mechanism to enable future extension as more fields are added. With 4 bits, we get twice as much field capacity despite the limitation to 16 enumerations. The limitation is guarded in the Python code and will fail assertion if violated.

Regarding `struct Key` which you suggested, it looks good, and using 32 bits may be ok for now, but may soon run out and require changes in both geneate.py and device.h. There could be trade-offs on the look-up time of 32-bit vs 64 bit keys, but not sure if it is worth it given that the code is not on the performance critical path. I have this experimental commit here. I can add cherry-pick it here ? any thoughts? https://github.com/mustafabar/rccl/commit/4a879b5d58f1b96078c4eaedc56a07ccb8025fa3 

Regarding Union type, I am a bit worried because the code is generated in Python with LSB shifting mimicking little-Endian. So if the code is built on big-Endian machine, the Union type logic will break.

alex-breslow-amd: I don't have any strong opinions here about this.  The Python concerns you mentioned weren't ones I was considering.  No objections from me regarding merging.

PR #1858, opened 2025-08-12
device: optimize threadfence for ll64 protocol

nusislam:
## Details
The `postSend()` function in LL64 protocol uses a threadfence() that adds up to 5% overhead for medium message sizes. This PR minimizes this overhead by up to 2.5% on MI300X.
  
**What were the changes?**  
Instead of `__threadfence()`, `__threadfence_block()` was used along with `asm volatile("s_waitcnt lgkmcnt(0) vmcnt(0)")`;

Minimize the LL64 protocol overhead.

**Additional Documentation:**  
Need to evaluate on MI350.

## Approval Checklist

### DIFF
--- a/src/device/prims_ll128.h
+++ b/src/device/prims_ll128.h
@@ -138,13 +138,13 @@ private:
     if (recvConnHeadPtr) STORE(recvConnHeadPtr, recvConnHead += 1);
   }
   inline __device__ void postSend() {
+    __atomic_signal_fence(__ATOMIC_SEQ_CST);	  
     asm volatile("s_waitcnt lgkmcnt(0) vmcnt(0)"); 	  
+    __atomic_signal_fence(__ATOMIC_SEQ_CST);
+
     if (sendConnTailPtr) {
 #if __CUDA_ARCH__ >= 900
       __threadfence_system();
-#else
-      //__threadfence();
-      __threadfence_block();
 #endif
       STORE((unsigned long long *)sendConnTailPtr, sendConnTail += 1);
     }
--- a/src/device/prims_ll128.h
+++ b/src/device/prims_ll128.h
@@ -138,11 +138,13 @@ private:
     if (recvConnHeadPtr) STORE(recvConnHeadPtr, recvConnHead += 1);
   }
   inline __device__ void postSend() {
+    asm volatile("s_waitcnt lgkmcnt(0) vmcnt(0)"); 	  
     if (sendConnTailPtr) {
 #if __CUDA_ARCH__ >= 900
       __threadfence_system();
 #else
-      __threadfence();
+      //__threadfence();
+      __threadfence_block();
 #endif
       STORE((unsigned long long *)sendConnTailPtr, sendConnTail += 1);
     }


wenkaidu (approved):

src/device/prims_ll128.h
```diff
@@ -138,11 +138,13 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL128, P2p, isNetOffload>:
...
   }
   inline __device__ void postSend() {
+    asm volatile("s_waitcnt lgkmcnt(0) vmcnt(0)"); 	  
     if (sendConnTailPtr) {
 #if __CUDA_ARCH__ >= 900
       __threadfence_system();
 #else
-      __threadfence();
+      //__threadfence();
+      __threadfence_block();
```
alex-breslow-amd: You don't need this extra __threadfence_block.  This gets compiled to s_waitcnt lgkmcnt(0).  You've already done the wait on this above.  You just need to prevent reordering around the asm volatile call.  I'd recommend putting the asm volatile call between two __atomic_signal_fence(__ATOMIC_SEQ_CST) calls to prevent raising and lowering.  This is what I do in PR1830.  It's a bit pedantic, but I think this will make it more correct.

nusislam: I tried using `__atomic_signal_fence(__ATOMIC_SEQ_CST)` instead of `__threadfence_block()`. Performance wise, both look similar. @wenkaidu - any thoughts on which option to use?

wenkaidu: I think __atomic_signal_fence(__ATOMIC_SEQ_CST)  maybe better as it doesn't generate any code.

alex-breslow-amd (approved):
Thank you for making this change.  Since lgkmcnt and vmcnt are registers per wavefront, this is sufficient.

src/device/prims_ll128.h
```diff
@@ -138,11 +138,13 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL128, P2p, isNetOffload>:
     if (recvConnHeadPtr) STORE(recvConnHeadPtr, recvConnHead += 1);
   }
   inline __device__ void postSend() {
+    __atomic_signal_fence(__ATOMIC_SEQ_CST);	  
+    asm volatile("s_waitcnt lgkmcnt(0) vmcnt(0)");
```
alex-breslow-amd, line 142: @nusislam, please note that this s_waitcnt likely won't work on subsequent GPU generations beyond GFX950.  We'll need to add some additional logic here.  Let's discuss offline since I don't think those details are publicly available.

alex-breslow-amd: This unfortunately won't work with user registered buffers on the receiver that aren't marked as uncacheable on the sender.  I will add a Jira item to track since it also impacts PR1830.  We discussed offline.

nusislam: ll128 is not used with UBR.

PedramAlizadeh (approved):

src/device/prims_ll128.h
dmwu: @nusislam  Can you provide more context on why ll128 is not used in UBR? is this something hardcoded or may change by users through ENVs

nusislam: LL128 protocol code does not keep track of UBR (e.g., `regUsed`, `netRegUsed` are there in `Simple`: https://github.com/ROCm/rccl/blob/develop/src/device/prims_simple.h#L816, not in LL128). Moreover, UBR is currently disabled in RCCL.

PR #1866, opened 2025-08-19
Removing "Could not find any local path from gpu X to net." warning

thananon:
## Details

**Work item:** SWDEV-546101

**What were the changes?**  
Removing the warning introduced by NCCL 2.26 that caused confusion to the users.

This warning caused so many false positive when debugging issues.

ifdefs

## Approval Checklist

### DIFF
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -1545,10 +1545,12 @@ ncclResult_t ncclTopoGetLocalNet(struct ncclTopoSystem* system, int rank, int ch
   int localNets[NCCL_TOPO_MAX_NODES];
   int localNetCount;
   NCCLCHECK(ncclTopoGetLocal(system, GPU, gpu, NET, localNets, &localNetCount, NULL));
+#if !defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__)
   if (localNetCount==0) {
     WARN("Could not find any local path from gpu %d to net.", gpu);
     return ncclInternalError;
   }
+#endif
 
   int localGpus[NCCL_TOPO_MAX_NODES];
   int localGpuCount;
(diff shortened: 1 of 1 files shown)


alex-breslow-amd (approved):

src/graph/topo.cc
```diff
@@ -1545,10 +1545,12 @@ ncclResult_t ncclTopoGetLocalNet(struct ncclTopoSystem* system, int rank, int ch
   int localNets[NCCL_TOPO_MAX_NODES];
   int localNetCount;
   NCCLCHECK(ncclTopoGetLocal(system, GPU, gpu, NET, localNets, &localNetCount, NULL));
+#if !defined(__HIP_PLATFORM_AMD__) && !defined(__HIPCC__)
```
alex-breslow-amd: Maybe use De Morgan's law:
```
#if !(defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__))

```

I think this is cleaner, but up to you.  Thinking about it more, maybe just checking if we're on a system with NVIDIA GPUs is simpler, but this is fine. :-}

thananon: I picked this one from existing code to be consistent.

alex-breslow-amd (approved):

src/graph/topo.cc
alex-breslow-amd: A more general comment is if this is safe to disable.

mustafabar (approved):

src/graph/topo.cc
corey-derochie-amd: Yeah, this isn't just a warning change, it's a functional change, right? Why is it valid to proceed? Is that just the case on AMD?

thananon: hmm maybe the guard needs to be only for the print statement.

thananon:
Updated to only guard the WARN statement.

PR #1871, opened 2025-08-20
[BUILD] Populate host_table entries only for 1 unroll

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Modify generator.py for host_table entries to be agnostic of the no. of supported unroll factors.

host_table entries are a mapping of device functions based on collective, algorithm, protocol, redop, and datatype.
For GPU targets that support multiple unrolls, e.g., `gfx950` (or) for non-local builds, the current implementation will have duplicate entries in host_table.cpp.

Bring back the sweep, which is limited to the first set of valid functions, ordered by unroll.

## Approval Checklist

### DIFF
--- a/src/device/generate.py
+++ b/src/device/generate.py
@@ -267,6 +267,7 @@ def equivalent_primary(coll, algo, proto, redop, ty, acc, unroll):
   return (coll, algo, proto, redop, ty, acc, unroll)
 
 # Order rows are enumerated must match formula of `ncclDevFuncId()`:
+# outermost loop should be for unroll factor; refer to host_table section
 def enumerate_func_rows():
   for unroll in all_unroll:
     for acc in use_acc:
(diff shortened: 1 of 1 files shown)


src/device/generate.py
```diff
@@ -474,7 +474,7 @@ def custom_sort_key(fn):
   out("//   bits 16-19:  ty index\n")
   out("#include <unordered_map>\n")
   out("extern std::unordered_map<uint64_t, int> ncclDevFuncNameToId = {\n")
-  for fn in func_rows:
+  for fn in func_rows[:len(func_rows)//len(all_unroll)]:
```
mustafabar, line 482: pls add a comment in place here why this division is needed. Also in the `enumerate_func_rows` indicate that unroll has to be at top level.

mustafabar (changes requested):

wenkaidu (approved):

mustafabar (approved):

PR #1876, opened 2025-08-22
[TheRock CI] Adding single node tests for RCCL

geomin12:
Adding single node tests, the same tests we use in TheRock

Working here: https://github.com/ROCm/rccl/actions/runs/17255114232

### DIFF
--- a/.github/workflows/therock-test-packages-single-node.yml
+++ b/.github/workflows/therock-test-packages-single-node.yml
@@ -1,4 +1,4 @@
-name: TheRock Test Packages
+name: TheRock Test Packages single-node
 
 on:
   workflow_call:
@@ -9,8 +9,6 @@ on:
         type: string
       artifact_run_id:
         type: string
-      test_type:
-        type: string
   workflow_dispatch:
     inputs:
       amdgpu_families:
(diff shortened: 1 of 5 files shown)


nileshnegi (approved):

.github/workflows/therock-test-packages.yml
```diff
@@ -50,9 +58,54 @@ jobs:
...
+          ARTIFACT_RUN_ID: ${{ env.ARTIFACT_RUN_ID }}
+          AMDGPU_FAMILIES: ${{ inputs.amdgpu_families }}
+          OUTPUT_ARTIFACTS_DIR: ${{ env.OUTPUT_ARTIFACTS_DIR }}
+          VENV_DIR: ${{ env.VENV_DIR }}
+          FETCH_ARTIFACT_ARGS: "--rccl --tests"
+          IS_PR_FROM_FORK: ${{ github.event.pull_request.head.repo.fork }}
+
+      # MPI install is required until https://github.com/ROCm/TheRock/issues/1284 is resolved
+      - name: Install dependencies
+        run: sudo apt install openmpi-bin openmpi-common libopenmpi-dev -y
```
marbre: We've created a Docker image which has MPI pre-installed. Feels like we should rather rebuild tests with MPI disabled or limit to only rccl unittestes (excluding rccl-tests).

marbre (changes requested):

.github/workflows/therock-test-packages-single-node.yml
```diff
@@ -60,6 +60,8 @@ jobs:
 
       - name: Test
         timeout-minutes: 15
+        # Currently, TheRock CI always builds with MPI enabled which causes RCCL correctness test failures
+        # TODO (geomin12): rebuild rccl-tests without MPI to enable RCCL correctness tests
```
marbre:
```suggestion
        # Currently, TheRock CI in RCCL always builds with MPI-supported enabled which causes the
        # RCCL correctness tests to fail on the mi325 runners which don't have MPI pre-installed.
        # TODO (geomin12): Rebuild rccl-tests without MPI to enable RCCL correctness tests.
```

marbre (approved):
LGTM, assuming the newly added workflow passes.

PR #1895, opened 2025-09-05
Added new tests for rccl_wrap - rcclOverrideProtocol, rcclOverrideAlgorithm

Kapil-Shyam-Pawar:
## Details

**Work item:** _"Internal"_

**What were the changes?**  
Tests for `rcclOverrideProtocol` and `rcclOverrideAlgorithm` as an extension to tests for rccl_wrap.cc - https://github.com/ROCm/rccl/pull/1694

The tests were added to cover function `rcclOverrideProtocol` and `rcclOverrideAlgorithm`

12 tests are added to cover all use cases for rcclOverrideProtocol, rcclOverrideAlgorithm, and rcclGetAlgoProtoIndex

**Additional Documentation:**  

The following command(s) can be used to run the tests below:
-`RcclOverrideProtocol_NoOverride`
-`RcclOverrideAlgorithm_NoOverride`
```
./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverride*
```

The following command(s) can be used to run the tests below:
-`RcclOverrideProtocol_UnsupportedOverride`
-`RcclOverrideAlgorithm_UnsupportedOverride`
-`RcclOverrideProtocol_ValidOverride`
-`RcclOverrideAlgorithm_ValidOverride`
-`RcclOverrideProtocol_ValidOverridePersists`
-`RcclOverrideAlgorithm_ValidOverridePersists`
```
// Run RcclOverrideProtocol_UnsupportedOverride, RcclOverrideProtocol_ValidOverride, and RcclOverrideProtocol_ValidOverridePersists
RCCL_OVERRIDE_PROTO=Simple ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverrideProtocol*

// Run RcclOverrideAlgorithm_UnsupportedOverride, RcclOverrideAlgorithm_ValidOverride, RcclOverrideAlgorithm_ValidOverridePersists
RCCL_OVERRIDE_ALGO=Ring ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverrideAlgorithm*

// Run all 6 tests together
RCCL_OVERRIDE_PROTO=Simple RCCL_OVERRIDE_ALGO=Ring ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverride*
```

The following command(s) can be used to run the tests below:
-`RcclOverrideProtocol_InvalidProtocol`
-`RcclOverrideAlgorithm_InvalidAlgorithm`
-`RcclOverrideProtocol_InvalidOverridePersists`
-`RcclOverrideAlgorithm_InvalidOverridePersists`
```
// Run RcclOverrideProtocol_InvalidProtocol and RcclOverrideProtocol_InvalidOverridePersists
RCCL_OVERRIDE_PROTO=XXXX ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverrideProtocol*

//Run RcclOverrideAlgorithm_InvalidAlgorithm and RcclOverrideAlgorithm_InvalidOverridePersists
RCCL_OVERRIDE_ALGO=XXXX ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverrideAlgorithm*

// Run all 4 tests together
RCCL_OVERRIDE_PROTO=XXXX RCCL_OVERRIDE_ALGO=XXXX ./rccl-UnitTestsFixtures --gtest_filter=*Rcclwrap.RcclOverride*
```
## Approval Checklist

### DIFF
--- a/test/RcclWrapTests.cpp
+++ b/test/RcclWrapTests.cpp
@@ -177,6 +177,30 @@ static bool ShouldSkipRcclSetPipeliningTests() {
   return false;
 }
 
+// Helper function to validate protocol string against known valid protocols
+static bool isProtoStrValid(const char *envStr) {
+  if (!envStr)
+    return false;
+  for (int i = 0; i < NCCL_NUM_PROTOCOLS; ++i) {
+    if (strcasecmp(envStr, ncclProtoStr[i]) == 0) {
+      return true; // Match found
+    }
+  }
+  return false; // No match found
+}
+
+// Helper function to validate algorithm string against known valid algorithms
+static bool isAlgoStrValid(const char *envStr) {
+  if (!envStr)
+    return false;
+  for (int i = 0; i < NCCL_NUM_ALGORITHMS; ++i) {
+    if (strcasecmp(envStr, ncclAlgoStr[i]) == 0) {
+      return true; // Match found
+    }
+  }
+  return false; // No match found
+}
+
 TEST(Rcclwrap, RcclFuncMaxSendRecvCount) {
   ncclResult_t staticCheckResult = testStaticExposeCheck();
 #ifdef RCCL_EXPOSE_STATIC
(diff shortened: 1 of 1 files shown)


atulkulk (approved):

venksubramd:
We have some good initial tests for rcclOverrideProtocol and rcclOverrideAlgorithm. Nice work on those. We may want to look a bit closer to fill in a couple of more tests.

Some tests for rcclOverrideProtocol:
1.
if protoVal == NCCL_PROTO_UNDEF and rcclGetAlgoProtoIndex fails assert that it returns ncclInvalidUsage. In addition, as a way to check that the function set the static validInput variable, we would want to make a subsequent call to rcclOverrideProtocol and that it returns ncclInvalidUsage.

2.
Make call to rcclGetAlgoProtoIndex pass and assert that rcclOverrideProtocol returns ncclSuccess, and in addition, check that a subsequent call to rcclOverrideProtocol returns ncclSuccess if another call to rcclGetAlgoProtoIndex from with in the second call to rcclOverrideProtocol passes.
  
3.
When protoVal > NCCL_PROTO_UNDEF and table[info->algorithm][protoVal] == NCCL_ALGO_PROTO_IGNORE, check that the call to rcclOverrideProtocol returns ncclInvalidUsage and also that a subsequent call to rcclOverrideProtocol with  protoOverrideEnv being false returns ncclSuccess.

The general idea here with these additional tests is to verify that the internal state of the function, that is setting the static variable, continues to behave in the future as it is now.

Similar tests for the rcclOverrideAlgorithm function.

In addition to the above new tests, I request that we revisit two existing tests for the following:

TEST(Rcclwrap, RcclOverrideProtocol_NoOverride) {
In this test we check if protoOverrideEnv is set and in that case we're not running the test. Is it possible to ensure that protoOverrideEnv is not set and then move forward to run this test? We generally do not want tests to be skipped based on conditions. It would be better to move the check and instead set the flag to the desired value.

TEST(Rcclwrap, RcclOverrideProtocol_UnsupportedOverride) {
Likewise in this test, instead of checking the value for protoOverrideEnv, it will be great if we can set it to the desired value so this test can run instead of ever skipping.

Thank you.

Kapil-Shyam-Pawar:
@venksubramd Thank you for the comments.
I'll start working on the new tests as suggested.

Regarding the review points on the existing tests:
We are avoiding setting environment variables like `RCCL_OVERRIDE_PROTO` and `RCCL_OVERRIDE_ALGO` within the tests because they are static and cannot be updated again in a different test in the same execution. This can lead to test failures if other tests depend on different values. To prevent this, we have added a skip condition and set these environment variables via command-line arguments when running the tests.

Let me know what you think!

Kapil-Shyam-Pawar:
@venksubramd I’ve added the first two tests as suggested.

The issue with the 3rd test is that it would require unsetting the environment variable within the test before the second call, in order for `protoOverrideEnv` to be false. This could cause side effects on other tests that depend on that variable, and for the reasons mentioned in my comment above, we avoid modifying environment variables at runtime inside individual tests.

venksubramd:
Thank you @Kapil-Shyam-Pawar these newly added tests and your comments address the items I had previously mentioned.

corey-derochie-amd:
> @venksubramd Thank you for the comments. I'll start working on the new tests as suggested.
> 
> Regarding the review points on the existing tests: We are avoiding setting environment variables like `RCCL_OVERRIDE_PROTO` and `RCCL_OVERRIDE_ALGO` within the tests because they are static and cannot be updated again in a different test in the same execution. This can lead to test failures if other tests depend on different values. To prevent this, we have added a skip condition and set these environment variables via command-line arguments when running the tests.
> 
> Let me know what you think!

I agree with @venksubramd and I think we should develop a better strategy around these one-time-settable environment.

Perhaps we should build a test harness into rccl to allow the rccl env to be overridden by the unit test. We could add a way to override the values in NCCL_PARAM and RCCL_PARAM through a hidden API, or if necessary we could execute tests in new processes so that we can set NCCL_ and RCCL_ env vars.

corey-derochie-amd (approved):

Kapil-Shyam-Pawar:
Moving ahead with this PR. We’ll explore a new approach, as suggested, for handling environment variable overrides in parallel.

PR #1759, opened 2025-06-24
[DEVICE] Switch to amd-smi from rocm-smi

nileshnegi:
## Details

**Work item:**
Internal
AICOMRCCL-344 ~SWDEV-509916~

**What were the changes?**  
Added `amd-smi` calls to code sections that currently use `rocm-smi` API.

**Why were the changes made?**  
`rocm-smi` is deprecated and will be removed in ~ROCm 7.2.0~ ROCm 7.11.0. `amd-smi` is the way to go now.

**How was the outcome achieved?**  
Created alternative files for `amd-smi` calls, while preserving `rocm-smi` files for now to maintain backward compatibility with older ROCm versions.

**Additional Documentation:**
- This PR is more a dependency fix, to remove RCCL dependency on rocm-smi.
- By default, RCCL still relies on sysfs entries and alt_rsmi for detecting GPU topology.
- `RCCL_USE_AMD_SMI_LIB=1` can be used to enable `amd-smi`-based topology detection in RCCL. However, this needs extensive testing and expert feedback from @gabrpham and @marifamd to ensure correctness in the AMD SMI API usage.

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #1920, opened 2025-09-16
[INIT] Use rocm-smi API instead of CLI for querying FW version

nileshnegi:
## Details

**Work item:**
Internal

**What were the changes?**  
Switched to rocm-smi API calls for querying GPU FW version instead of parsing `amd-smi firmware` output

RCCL UnitTests fail with TheRock builds as amd-smi is not yet available.
Also, relying on CLI output instead of API is a bad practice.

Used `rsmi_dev_firmware_version_get(0, RSMI_FW_BLOCK_MEC, &fw_version)` instead of parsing output of `amd-smi firmware`

**Additional Documentation:**  
Currently using rocm-smi API. There is an active PR for switching to amd-smi https://github.com/ROCm/rccl/pull/1759.

## Approval Checklist

### DIFF
--- a/src/rccl_wrap.cc
+++ b/src/rccl_wrap.cc
@@ -24,6 +24,7 @@ THE SOFTWARE.
 #include "comm.h"
 #include "graph/topo.h"
 #include "enqueue.h"
+#include "rocm_smi/rocm_smi.h"
 
 // Use this param to experiment pipelining new data types besides bfloat16
 // Make sure you generate the device code with the new data type (i.e. in generate.py)
--- a/src/rccl_wrap.cc
+++ b/src/rccl_wrap.cc
@@ -342,37 +342,26 @@ std::vector<std::string> splitString(const std::string& s, char delimiter) {
   return tokens;
 }
 
-int parseFirmwareVersionImpl(FILE* file) {
-  constexpr std::size_t MAX_LINE_SZ = 1024;
-  char line[MAX_LINE_SZ];
-  bool found_pattern = false;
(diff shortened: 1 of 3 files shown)


src/rccl_wrap.cc
```diff
@@ -342,37 +342,26 @@ std::vector<std::string> splitString(const std::string& s, char delimiter) {
...
-      }
-      continue;
-    }
-
-    if (found_pattern && (parts[0] == "FW_VERSION")) {
-      return stoi(parts[1]) & 0x7ff;
-    }
-  }
-  return -1;
+int parseFirmwareVersionImpl() {
```
weilewei, line 346: nit: maybe we can have a better name for this function since we no longer parse texts and simply return firmware version. For example, getFirmwareVersion(), query*, etc.

weilewei (approved):
LGTM. Thanks.

src/rccl_wrap.cc
corey-derochie-amd: There is also no longer a need for a separate `impl` function.

corey-derochie-amd (approved):
lgtm with minor suggestion above

src/rccl_wrap.cc
nileshnegi: will fix this in PR https://github.com/ROCm/rccl/pull/1759

nileshnegi:
will incorporate suggestions in PR https://github.com/ROCm/rccl/pull/1759

PR #1924, opened 2025-09-17
Remove noinline by preventing compiler to move instructions around in…

carlobertolli:
…line assembly.

## Details

Internal.

**What were the changes?**  
Remove noinline on LLGenericOp and add builtin for waitcnt to prevent compiler reordering of instructions.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/prims_ll.h
+++ b/src/device/prims_ll.h
@@ -269,6 +269,7 @@ private:
     i4[2] = (val >> 32);
     i4[3] = flag;
     asm volatile ("flat_store_dwordx4 %0, %1 sc0 sc1 nt" :: "v"(dst), "v"(i4));
+    __builtin_amdgcn_s_waitcnt(0); // prevents the compiler to move instructions over ASM.
 #else
     union ncclLLFifoLine i4;
     i4.data1 = val & 0xffffffff;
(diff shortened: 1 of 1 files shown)


src/device/prims_ll.h
```diff
@@ -269,6 +269,7 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
     i4[2] = (val >> 32);
     i4[3] = flag;
     asm volatile ("flat_store_dwordx4 %0, %1 sc0 sc1 nt" :: "v"(dst), "v"(i4));
+    __builtin_amdgcn_s_waitcnt(0); // prevents the compiler to move instructions over ASM.
```
alex-breslow-amd, line 272: 1) Have you tried sandwiching the ASM between two __atomic_signal_fence(__ATOMIC_SEQ_CST) calls to prevent raising and lowering?  That may be enough.

2) Also, this isn't minimal for the waitcnt.  We only need to check that vmcnt=0 lgkmcnt(0).  I think you want __builtin_amdgcn_s_waitcnt(7<<4).  What you currently have does vmcnt(0) lgkmcnt(0) expmcnt(0).  The 7 shifted by 4 comes from here: https://llvm.org/docs/AMDGPU/gfx9_waitcnt.html

I verified in godbolt too for gfx942: https://godbolt.org/z/eehEr8qE6

3) I'm concerned about the performance impact that this will have.  Basically what this means is that for every single store, we wait for that store to complete by checking the lgkmcnt, expmcnt, and vmcnt registers to be zero.  Ideally, we would issue a single s_waitcnt for a group of stores to reduce the overhead by having greater memory level parallelism per wavefront.  I'm not sure if this is feasible/correct though for this protocol.

nileshnegi:
@carlobertolli can we close this PR?

alex-breslow-amd:
We can close this.

PR #1930, opened 2025-09-22
Enable inlining of LLGenericOp and disable 128-bit store

lfmeadow:
Enable inlining of LLGenericOp. Revert to 2x64-bit stores.

## Details
This improves performance vs. the non-inlined version.

**What were the changes?**  
Removed noinline attribute and reverted store.

Improve performance

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/prims_ll.h
+++ b/src/device/prims_ll.h
@@ -261,7 +261,7 @@ private:
 
   __device__ void storeLL(union ncclLLFifoLine* dst, uint64_t val, uint32_t flag) {
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
-#if (defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
+#if 0 //(defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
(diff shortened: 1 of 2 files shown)


src/device/prims_ll.h
```diff
@@ -261,14 +261,15 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
 
   __device__ void storeLL(union ncclLLFifoLine* dst, uint64_t val, uint32_t flag) {
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
-#if (defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
+#if 0 //(defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
```
alex-breslow-amd, line 264: Since this will require an update to ROCr to work, can you put in the appropriate version guards?

I think you want something like the following:
```
#if ROCM_VERSION >= 70200
```

Same with whether to enable the inlining.  I need to check what's in HIP_HOST_UNCACHED_MEMORY.  The guards I suppose are already there, but we may have to update the ROCM_VERSION checks.

PR #1932, opened 2025-09-23
Fix hang on MI350 for both cached and uncached memory

lfmeadow:
## Details
- add missing fences, this fixes the hangs, they aren't needed if pages are UC
- remove asm store_dwordx4, not needed since compiler fuses adjacent stores

**Work item:** 
Internal

**What were the changes?**  
Described above

Fixed the hang on MI350

There was a persistent hanging problem when inlining LLGenericOp in prims_ll.h. This was thought to be a compiler problem. In fact, it turned out that there were missing memory fences so stores were not globally visible.

Two calls to __builtin_amdgcn_fence(__ATOMIC_RELEASE, "") cause the GPU cache to be flushed to system scope and makes the stores visible to the external reader. The inline asm store was not needed since the compiler fuses two adjacent calls to __buitin_nonterminal_store into a single flat_store_dwordx4 ..

**Additional Documentation:**  
The performance vs. the old non-inlined code seems to have improved by about 30%, but extensive benchmarking on MI350 still needs to be done, and the code has not been tested on MI300. When UC memory is properly deployed in ROCr for MI350, the two fences can be removed, for a modest performance improvement.

## Approval Checklist
 No such changes were made.

### DIFF
--- a/src/device/prims_ll.h
+++ b/src/device/prims_ll.h
@@ -261,15 +261,6 @@ private:
 
   __device__ void storeLL(union ncclLLFifoLine* dst, uint64_t val, uint32_t flag) {
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
-#if (defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
-    using Vec = uint32_t __attribute__((ext_vector_type(4)));
-    Vec i4;
-    i4[0] = val & 0xffffffff;
-    i4[1] = flag;
-    i4[2] = (val >> 32);
-    i4[3] = flag;
-    asm volatile ("flat_store_dwordx4 %0, %1 sc0 sc1 nt" :: "v"(dst), "v"(i4));
-#else
     union ncclLLFifoLine i4;
     i4.data1 = val & 0xffffffff;
     i4.flag1 = flag;
@@ -277,6 +268,8 @@ private:
     i4.flag2 = flag;
     __builtin_nontemporal_store(i4.v[0], dst->v);
     __builtin_nontemporal_store(i4.v[1], dst->v+1);
+#if defined(__gfx950__)
+    __builtin_amdgcn_fence(__ATOMIC_RELEASE, ""); // flush cache
 #endif
 #else
     asm volatile("st.volatile.global.v4.u32 [%0], {%1,%2,%3,%4};" :: "l"(&dst->i4), "r"((uint32_t)val), "r"(flag), "r"((uint32_t)(val >> 32)), "r"(flag) : "memory");
@@ -351,6 +344,9 @@ private:
       __builtin_nontemporal_store(u4, (uint32_t*)dst);
     else
       __builtin_nontemporal_store(u8, (uint64_t*)dst);
+#if defined(__gfx950__)
+    __builtin_amdgcn_fence(__ATOMIC_RELEASE, ""); // flush cache
+#endif
 #else
     if(sizeof(U) == 1)
       asm volatile("st.volatile.global.b8 [%0],%1;" :: "l"(dst), "r"(u4) : "memory");
@@ -432,11 +428,7 @@ private:
   }
 
   template <int RECV, int SEND, int SrcBuf, int DstBuf>
-#if defined(__gfx950__)
-  __device__ __attribute__((noinline)) void LLGenericOp(intptr_t srcIx, intptr_t dstIx, int nelem, bool postOp) {
-#else
   __device__ void LLGenericOp(intptr_t srcIx, intptr_t dstIx, int nelem, bool postOp) {
-#endif
     constexpr int SRC = SrcBuf != -1 ? 1 : 0;
     constexpr int DST = DstBuf != -1 ? 1 : 0;
     T *srcElts = SrcBuf == -1 ? nullptr : userBufs[SrcBuf] + srcIx;


src/device/prims_ll.h
```diff
@@ -261,23 +262,16 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
...
     i4.data1 = val & 0xffffffff;
     i4.flag1 = flag;
     i4.data2 = (val >> 32);
     i4.flag2 = flag;
+    // The compiler merges the two stores into a store_dwordx4
     __builtin_nontemporal_store(i4.v[0], dst->v);
     __builtin_nontemporal_store(i4.v[1], dst->v+1);
-#endif
+    // This fence can be removed for uncached memory
+    __builtin_amdgcn_fence(__ATOMIC_RELEASE, "");
```
ssahasra: This system-scope fence flushes the entire cache anyway, then what's the point of the nontemporal stores? Is there any measurable difference if you use ordinary stores? Also, doesn't this flush cause a measureable performance hit on the memory accesses being performed by all other waves on the same device?

alex-breslow-amd: @ssahasra, we have to perform benchmarking to see.  Without the sc0 sc1 bits set, it's not a system scope store, so we need this fence if the host memory is not mapped as uncached on the GPU.  Note nontemporal makes no guarantee about cache bypassing.  The sc0 sc1 bits are necessary.  __builtin_nontemporal_store and __builtin_nontemporal_load have sc0=0 and sc1=0.

alex-breslow-amd: This needs to be gated by #ifdef(__gfx950__).  Otherwise, it will get invoked on MI300X.

src/device/prims_ll.h
```diff
@@ -71,11 +71,8 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
   inline __device__ void barrier() {
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
     if (nthreads != WARP_SIZE)
-      #if defined(__gfx942__) || (defined(__gfx950__) && defined(HIP_HOST_UNCACHED_MEMORY))
-        barrier_generic(__threadfence_block(), nthreads, barrier_next, barriers);
-      #else
-        barrier_generic(__threadfence(), nthreads, barrier_next, barriers);
-      #endif
+      barrier_generic(__builtin_amdgcn_fence(__ATOMIC_ACQUIRE, "workgroup"),
```
alex-breslow-amd: @lfmeadow, I'm curious if this acquire is necessary.  Maybe it's better to stick with the more hip idiomatic __threadfence_block here if it works.  On pre-gfx942, I don't think the __threadfence is that expensive, so it can probably stay as is.  Otherwise, we'll need to test for MI250 and prior generations.

src/device/prims_ll.h
```diff
@@ -89,10 +86,10 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
 
   __device__ inline int checkAbort(int &abortCache, const int abortValue, int &spins) {
     if (abortCache == 0 && ++spins == NCCL_SPINS_BEFORE_CHECK_ABORT) {
-      int abort = __atomic_load_n((ncclShmem.comm.abortFlag), __ATOMIC_SEQ_CST);
+      int abort = __scoped_atomic_load_n(ncclShmem.comm.abortFlag, __ATOMIC_SEQ_CST, __MEMORY_SCOPE_SYSTEM);
```
alex-breslow-amd: I don't hate this, but it does increase the likelihood of introducing a bug.

src/device/prims_ll.h
```diff
@@ -295,29 +289,24 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
...
 #else
-      u8 = __builtin_nontemporal_load((uint64_t*)src);
+    if(sizeof(U) == 1)
+      u1 = __builtin_nontemporal_load((uint8_t*) src);
+    else if(sizeof(U) == 2)
+      u2 = __builtin_nontemporal_load((uint16_t*) src);
+    else if(sizeof(U) == 4)
+      u4 = __builtin_nontemporal_load((uint32_t*) src);
+    else
+      u8 = __builtin_nontemporal_load((uint64_t*) src);
```
alex-breslow-amd: These changes too?  Are they necessary?  I'm not super familiar with GFX11, but I'm assuming this code might break there without the special treatment using atomics.

src/device/prims_ll.h
lfmeadow: I had __scoped_atomic_store_n(... __ATOMIC_RELEASE, __MEMORY_SCOPE_SYSTEM) and it was far slower. I used the nontemporal stores because the compiler optimizes them into a single dwordx4 store. In the uncached case, the fence does indeed slow things down by 3 to 5%.

lfmeadow: Or it can just be deleted all together once MI300 UC is in place.

src/device/prims_ll.h
lfmeadow: This generates identical code to __threadfence_block, as we discussed. I'm fine changing it back.

src/device/prims_ll.h
lfmeadow: Fine, I'll remove it.

src/device/prims_ll.h
alex-breslow-amd: We can't control which ROCm version is on a system, so I'm thinking we won't be able to delete it.  It will have to stay there in perpetuity for backward compatibility.  But maybe I'm missing something.

src/device/prims_ll.h
lfmeadow: The code is identical to the previous code, but no, the changes aren't necessary.

src/device/prims_ll.h
alex-breslow-amd: It's worth noting that this is in part a byproduct of the protocol, which is LL here.  The delta is bigger with the simple protocol.

src/device/prims_ll.h
```diff
@@ -351,6 +340,8 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
       __builtin_nontemporal_store(u4, (uint32_t*)dst);
     else
       __builtin_nontemporal_store(u8, (uint64_t*)dst);
+    // This fence can be removed for uncached memory
+    __builtin_amdgcn_fence(__ATOMIC_RELEASE, "");
```
alex-breslow-amd: I wasn't sure what the keyword was here for system scope, so we settled with "" since it's system scope by default.

lfmeadow: I will look it up.

src/device/prims_ll.h
lfmeadow: Per request, I've redone the fix to have minimal changes to the baseline prims_ll.h . Therefore the nontemporal stores will remain.

lfmeadow: #ifdef is added, new PR landed as soon as I can test it .

src/device/prims_ll.h
```diff
@@ -295,29 +289,24 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
...
-#ifdef __GFX11__
       u4 = __atomic_load_n((uint32_t*)src, __ATOMIC_RELAXED);
-#else
-      u4 = __builtin_nontemporal_load((uint32_t*)src);
-#endif
     else
-#ifdef __GFX11__
       u8 = __atomic_load_n((uint64_t*)src, __ATOMIC_RELAXED);
 #else
-      u8 = __builtin_nontemporal_load((uint64_t*)src);
```
alex-breslow-amd, line 320: My concerns remain here about deleting this code that GFX11 specific.

lfmeadow: There was no code deleted, but anyway, it's back the way it was.

alex-breslow-amd (approved):
LGTM.  I think we can merge once we have completed a full performance sweep and CI passes.  Thanks for all of your hard work on this @lfmeadow.

mustafabar (approved):

src/device/prims_ll.h
ssahasra: > @ssahasra, we have to perform benchmarking to see. Without the sc0 sc1 bits set, it's not a system scope store, so we need this fence if the host memory is not mapped as uncached on the GPU. Note nontemporal makes no guarantee about cache bypassing. The sc0 sc1 bits are necessary. __builtin_nontemporal_store and __builtin_nontemporal_load have sc0=0 and sc1=0.

We are actually talking in the opposite directions. If the system-scope all-address-space release fence is present, then the loads and stores don't need anything special. They can be just ordinary loads and stores. Setting any of the extra bits by any means has no additional effect.

ssahasra: > Per request, I've redone the fix to have minimal changes to the baseline prims_ll.h . Therefore the nontemporal stores will remain.

Please put a comment next to the fence saying that its presence precludes the need for nontemporal stores, but they are present purely for historical reasons.

src/device/prims_ll.h
ssahasra: > I don't hate this, but it does increase the likelihood of introducing a bug.

How? Aren't they identical?

src/device/prims_ll.h
alex-breslow-amd: > > @ssahasra, we have to perform benchmarking to see. Without the sc0 sc1 bits set, it's not a system scope store, so we need this fence if the host memory is not mapped as uncached on the GPU. Note nontemporal makes no guarantee about cache bypassing. The sc0 sc1 bits are necessary. __builtin_nontemporal_store and __builtin_nontemporal_load have sc0=0 and sc1=0.
> 
> We are actually talking in the opposite directions. If the system-scope all-address-space release fence is present, then the loads and stores don't need anything special. They can be just ordinary loads and stores. Setting any of the extra bits by any means has no additional effect.

This is correct, yes.

src/device/prims_ll.h
```diff
@@ -351,6 +344,9 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, useAcc>:
       __builtin_nontemporal_store(u4, (uint32_t*)dst);
     else
       __builtin_nontemporal_store(u8, (uint64_t*)dst);
+#if defined(__gfx950__)
```
alex-breslow-amd, line 347: @lfmeadow, is this fence necessary?  I don't think this code path gets exercised on the sender side.  It seems pretty innocuous though.

nileshnegi:
precheckin CI passed on all systems except for 1, due to infra issue. safe to merge.

src/device/prims_ll.h
alex-breslow-amd: Resolving, will remove in post.

PR #1940, opened 2025-09-25
gfx950 channel tuning for ReduceScatter and AllGather

isaki001:
## Details

**What were the changes?**  
Changed the number of channels selected for reduceScatter on multi-node gfx950.

During enqueue, the number of channels are adjusted based on the value of threadThreshold. The resulting performance was suboptimal for many message sizes.

- Added entry to tuning table for specifying the number of channels along with min/max thresholds to trigger their selection.
- Thresholds are based on size-per-rank. 
- After the channels are adjusted based on the threadThreshold, a new routine rcclOverrideChannels is invoked to override the number of channels if the number of nodes is greater than two and if a valid threshold is specified for the corresponding message size and particular collective. 
- Another factor that can change the number of channels is the MinTrafficPerChannel which is hardcoded to 16KB. I have chosen not to change this for now based even though it may slightly decrease the number of channels. This is because on small message sizes, RCCL's internal channel overriding helps. This may require adjustment in the future. 

**Additional Documentation:**  

## Approval Checklist


mustafabar (approved):

src/enqueue.cc
```diff
@@ -2075,8 +2077,10 @@ static ncclResult_t topoGetAlgoInfo(
     rcclUpdateThreadThreshold(comm, nBytes, info, threadThreshold);
     INFO(NCCL_INIT, "pre-adjustment threadThreshold:%i nBytes:%lu nc:%i", threadThreshold, nBytes, nc);
 
+    int minNChannels = ncclParamMinNchannels();
```
nileshnegi, line 2077: is this affected if the user uses `NCCL_MIN_CTAS` instead of `NCCL_MIN_NCHANNELS`?

isaki001: No, that would not be picked up. Adding a commit to address this.

nileshnegi (approved):

PR #1785, opened 2025-07-02
Add collective latency profiler

ycui1984:
## Details
This feature profiles the collective information like collective type, latency, variance, message size, etc. The purpose of this profiler is too parts.
1. it tells you what are the most expensive collectives after running an end to end workloads, which helps guiding the optimization direction, like optimize allReduce or allGather, the message that we should pay attention to, etc. 
2. It profiles the variance across different ranks, and provide an aggregation view. Sometimes, we wanted to confirm that a collective is running well across all ranks, etc. 

**What were the changes?**  
1. Added
src/include/latency_profiler/CollTrace.h
src/include/latency_profiler/CollTraceEvent.h
src/include/latency_profiler/CollTraceFunc.h
src/include/latency_profiler/CollTraceUtils.h
src/include/latency_profiler/EventQueue.h
src/misc/latency_profiler/CollTrace.cc
src/misc/latency_profiler/CollTraceEvent.cc
src/misc/latency_profiler/CollTraceFunc.cc
src/misc/latency_profiler/CollTraceUtils.cc
test/latency_profiler/LatencyProfilerUnitTest.cpp
CMakeLists.txt

2. Modified
src/enqueue.cc
src/include/comm.h
src/init.cc
test/CMakeLists.txt

Implementation steps
1) measure collective latency by adding cuda/hip event before and after a kernel launch in RCCL.
2) Started a separate worker thread to collect latency data into its local ring buffer.
3) Perform CPU level all gather to exchange latency data between threads (200 to 300us per 100 latency data exchange)
4) Report when results buffer is full or every few minutes

**Why were the changes made?**  
The work is previously done via manually analyzing traces, which is hard to capture the whole end to end behavior and analyzing multiple gpu traces for a single collective execution stats is not scalable. 

**How was the outcome achieved?**  
1. unit tests
./rccl-UnitTests --gtest_filter=CollTraceUtilsTest*
================================================================================
 Environment variables:
 - UT_DEBUG_PAUSE       Pause for debugger attach                  (  0) <unset>
 - UT_SHOW_NAMES        Show test case names                       (  1) <unset>
 - UT_MIN_GPUS          Minimum number of GPUs to use              (  1) <unset>
 - UT_MAX_GPUS          Maximum number of GPUs to use              (  8) <unset>
 - UT_POW2_GPUS         Only allow power-of-2 # of GPUs            (  0) <unset>
 - UT_PROCESS_MASK      Whether to run single/multi process        (  3) <unset>
[...]

PR #1941, opened 2025-09-26
Adding a barrier feature to highlight delays from preceding kernels in RCCL calls

mberenjk:
## Details

**Work item:**LWPCOMMLIBS-519

**What were the changes?**  
This commit implements a barrier to expose delays introduced by the preceding kernel. Currently supported only on single-node setups.
Barrier can be enabled with `RCCL_INSERT_BARRIER=1`

Captures delays caused by preceding kernels, which manifest in NCCL calls.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/barrier.h
+++ b/src/device/barrier.h
@@ -1,0 +1,168 @@
+/*************************************************************************
+ * Copyright (c) 2015-2022, NVIDIA CORPORATION. All rights reserved.
+ * Modifications Copyright (c) 2019-2022 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include "device.h"
+#include "collectives.h"
+#include "primitives.h"
+
+__device__ __attribute__((noinline)) void  copyShmemData(struct ncclDevComm* comm, struct channelMasks channelMask, struct ncclDevKernelArgs const* args){
+  const int tid = threadIdx.x;
+  int tn = blockDim.x;
+  int x = tid;
+  int total = 0, y;
+  int num = MAXCHANNELS/64 > 0 ? MAXCHANNELS/64 : 1;
+
+  // Copy kernel args to shmem and then only read those. Otherwise the compiler
+  // will end up putting the args into thread local stack which is very wasteful.
+  if (tid < sizeof(ncclDevKernelArgs)/sizeof(uint32_t)) {
+    ((uint32_t*)&ncclShmem.args)[tid] = ((uint32_t*)args)[tid];
+  }
+
+  // To map blockId to channelId, we need the n'th set bit of channelMask which
+  // is the inverse of counting the number of set bits among the the first n.
+  // PTX has the fns instruction which does this but is extremely slow. We can
+  // do better when we know all threads are querying the same bitmask.
+  switch (tid/WARP_SIZE) {
+  case 0:
+  //ncclShmem.channelId = blockIdx.x;
+    for (int i = 0; i < num; i++) {
+      if (channelMask.masks[i] & (1ull<<x)) {
+        y = __popcll(channelMask.masks[i] & ((1ull<<x)-1));
+        y = total + y;
+        if (blockIdx.x == y) {
+          ncclShmem.channelId = x + total;
+          break;
+        }
+      }
+      if (WARP_SIZE < 64) {
+        x = WARP_SIZE + tid;
+        if (channelMask.masks[i] & (1ull<<x)) {
+          y = __popcll(channelMask.masks[i] & ((1ull<<x)-1));
+          y = y + total;
+          if (blockIdx.x == y) {
+            ncclShmem.channelId = x + total;
+            break;
+          }
+        }
+      }
+      total = total + __popcll(channelMask.masks[i]);
+    }
+    break;
+  case 1:
+    if (tid < WARP_SIZE + NCCL_MAX_GROUPS) {
+      if (tid == WARP_SIZE) ncclShmem.barrier_pat = 0;
+      ncclShmem.groups[tid-WARP_SIZE].barrier = 0;
+    }
+    break;
+  case 2:
+#ifdef ENABLE_FAULT_INJECTION
+    /* load faults injection before first sync threads */
+    if (tid == 2*WARP_SIZE) ncclShmem.faults = comm->faults;
+#endif
+    break;
+  case 3:
+    /* set abort flag to 0 */
+    if (tid == 3*WARP_SIZE) ncclShmem.aborted = 0;
+    break;
+  default:
+    break;
+  }
+
+  __syncthreads(); // publish ncclShmem.{args, channelId}
+
+  /* set abort flag to 0 */
+  if (tid == 0) {
+    ncclShmem.aborted = 0;
+    ncclShmem.channel.workCounter = comm->channels[ncclShmem.channelId].workCounter;
+  }
+
+  // Use first 2 warps to load comm and channel, and remaining load work batch.
+  switch (tid/WARP_SIZE) {
+  case 0:
+    { void* dst = &ncclShmem.comm;
+      void* src = comm;
+      int bytes = sizeof(ncclDevComm);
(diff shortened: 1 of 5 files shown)


src/init.cc
```diff
@@ -666,6 +668,19 @@ static ncclResult_t commAlloc(struct ncclComm* comm, struct ncclComm* parent, in
...
+    NCCLCHECK(ncclCudaHostCalloc(&comm->barrierWork, sizeof(ncclDevWorkColl)));
+    comm->barrierWork->sendbuff = (void*)comm->barrierSendBuffer;
+    comm->barrierWork->recvbuff = (void*)comm->barrierRecvBuffer;
+    comm->barrierWork->channelLo = 1;
+    comm->barrierWork->channelHi = 1;
+    comm->barrierWork->cbd.countLo = 1;
+    comm->barrierWork->cbd.countHi = 1;
+    comm->barrierWork->root = 0;
+  }
+
```
alex-breslow-amd, line 683: Any opposition to having this barrier code gated at compile time to prevent it from impacting production builds?  I'm wondering in light of https://github.com/ROCm/rccl/pull/1949 and SWDEV-556961.  That being said, I see the value to this being out there as part librccl.so in ROCm releases.  Do you have a sense of whether this balloons our binary size?  I don't think so, but I wanted to check.  Probably worth keeping it as is if we can.

corey-derochie-amd: This feature is specifically for debugging in production settings, so if it was off by default in production builds, we wouldn't be able to use it without building from source. We are hoping this will cut down on support calls, but if they have to build from source to run this test, we will have to get involved anyway. Opinion on this @gilbertlee-amd ?

Nevertheless, let's check that binary size @mberenjk . If it's a big problem I agree we should gate this.

wenkaidu:
The title itself needs clarification: this will not mitigate any delays caused by preceding kernel. It only exposes the delays to more visible. This may deflect issues from RCCL, but end of day, user need root cause and a fix of this issue. Only saying this is not caused by RCCL is not very helpful. Better method is to capture rocprof traces and understand which preceding kernels causing the delay.

src/device/barrier.h
```diff
@@ -0,0 +1,168 @@
...
+if (prevRank == root) {
+  //printf("1111 prevRank = %d, rank = %d, root = %d channelCount = %zu, chunkCount = %zu\n", prevRank, rank, root, channelCount, chunkCount);
+  for (size_t elemOffset = 0; elemOffset < channelCount; elemOffset += chunkCount) {
+    offset = gridOffset + elemOffset;
+    nelem = min(chunkCount, channelCount - elemOffset);
+    prims.send(offset, nelem);
+  }
+}
+else if (rank == root) {
+  //printf("2222 prevRank = %d, rank = %d, root = %d channelCount = %zu, chunkCount = %zu\n", prevRank, rank, root, channelCount, chunkCount);
```
alex-breslow-amd, line 145: Remove commented code

src/device/barrier.h
```diff
@@ -0,0 +1,168 @@
...
+const size_t chunkCount = 4096;
+const size_t channelCount = 1;
+const size_t gridOffset = 0;
+size_t offset;
+int nelem;
+Primitives<uint8_t, FuncSum<uint8_t>, FanSymmetric<1>, 0, ProtoLL, 0>
+  prims(tid, nthreads, &ring->prev, &ring->next, work->sendbuff, work->recvbuff, 0, 0, 0, 0);
+
+if (prevRank == root) {
+  //printf("1111 prevRank = %d, rank = %d, root = %d channelCount = %zu, chunkCount = %zu\n", prevRank, rank, root, channelCount, chunkCount);
```
alex-breslow-amd, line 137: Remove commented code

src/device/barrier.h
```diff
@@ -0,0 +1,168 @@
...
+    { int subtid = tid - 2*WARP_SIZE;
+      int subtn = tn - 2*WARP_SIZE;
+      // Coverity reports a possible thread divergence due to not all threads participating in the collective.
+      // However, the code ensures that the participation is on a per-warp basis.
+      // coverity[device_thread_diverged:FALSE]
+      //loadWorkBatchToShmem(subtid, subtn, args, /*batchIx=*/blockIdx.x);
+    } break;
+  }
+  __syncthreads();
+}
```
alex-breslow-amd, line 110: I see why you're copying all of this code like this because there might be stuff in NCCL kernel main that you don't want, but from what I can tell most if not all of it has #ifdef guards on it.  Why not just call NCCL kernel main directly?  What was the issue?  I bring this up because this code is more prone to breaking and is harder to maintain if the code here diverges from the main RCCL.  If you use the RCCL functions directly, then this might be easier to maintain.

src/device/barrier.h
```diff
@@ -0,0 +1,168 @@
...
+else {
+  for (size_t elemOffset = 0; elemOffset < channelCount; elemOffset += chunkCount) {
+    offset = gridOffset + elemOffset;
+    nelem = min(chunkCount, channelCount - elemOffset);
+    prims.recvReduceSend(offset, nelem);
+  }
+}
+}
+}
```
alex-breslow-amd, line 160: This also seems to be a lot of code duplication.  What was the blocker to calling the LL implementation of allreduce directly?  Let me know if I'm missing something.

src/device/barrier.h
mberenjk: The ncclKernelMain call is reached only after a long sequence of initializations and a deep call stack that launches the correct kernel inside ncclKernelMain. I didn’t need to replicate that entire path. The reason I added this portion of the ncclKernelMain code is to perform the necessary initialization for ncclShmem, which is required by the primitive calls. Beyond that, my understanding is that the intent is not to mimic the full path used to launch the barrier kernel. Instead, this is an independent kernel designed to introduce almost zero overhead.

alex-breslow-amd:
> The title itself needs clarification: this will not mitigate any delays caused by preceding kernel. It only exposes the delays to more visible. This may deflect issues from RCCL, but end of day, user need root cause and a fix of this issue. Only saying this is not caused by RCCL is not very helpful. Better method is to capture rocprof traces and understand which preceding kernels causing the delay.

Yeah, I think the intent here is that this feature reduces the number of tickets that get filed against RCCL and provides a quick way for calculating what percentage speedup one can actually get from optimizing RCCL (faster ticket resolution).  But I generally agree that analysis of torch profiler traces can get you similar information if not more.

wenkaidu:
> > The title itself needs clarification: this will not mitigate any delays caused by preceding kernel. It only exposes the delays to more visible. This may deflect issues from RCCL, but end of day, user need root cause and a fix of this issue. Only saying this is not caused by RCCL is not very helpful. Better method is to capture rocprof traces and understand which preceding kernels causing the delay.
> 
> Yeah, I think the intent here is that this feature reduces the number of tickets that get filed against RCCL and provides a quick way for calculating what percentage speedup one can actually get from optimizing RCCL (faster ticket resolution). But I generally agree that analysis of torch profiler traces can get you similar information if not more.

If we only need to look at RCCL starting variation, such feature has been merged on develop from https://github.com/ROCm/rccl/pull/1785. This implementation doesn't make any kernel side change. It also logs latency from every kernel launch, so outliers can be seen easily.

src/init.cc
thananon: I agree with @corey-derochie-amd. We should not compile this out. Lets @mberenjk build with debug build. If that passes,we are ok.

src/device/barrier.h
thananon: We want different kernel name to show up on trace.

alex-breslow-amd: That's fair.  I wasn't advocating for calling the LL kernel but rather the function that implements most of the functionality, but this is fine.  Just thought I'd make some constructive suggestions. This is more prone to breaking in certain situations with updates to the codebase.  Now you have to maintain near duplicate code and keep it in sync.  :-}

ammallya:
Cannot import to ROCm/rocm-systems due to merge conflicts please reopen this PR on rocm/rocm-systems.

PR #1949, opened 2025-09-30
Change to use -O0 instead of -O1 in debug build.

thananon:
## Details

**Work item:** SWDEV-556961

**What were the changes?**  
Using -O0 for debug build.

-O1 still does compiler optimization.

Change cmake config.

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -886,6 +886,15 @@ foreach(file ${GENERATED_FILES})
   list(APPEND HIP_SOURCES ${file})
 endforeach()
 
+# Apply -O1 optimization for generated files in debug builds
+# This is needed for device object to not go over the stacksize limit at link time while allowing
+# -O0 for host code in debug build.
+if(CMAKE_BUILD_TYPE MATCHES "Debug" AND NOT DEFINED ENV{CXXFLAGS})
+  message(STATUS "Setting -O1 optimization for device code in ${GEN_DIR} (Debug build)")
+  set_source_files_properties(${GENERATED_FILES} PROPERTIES COMPILE_OPTIONS "-O1")
+endif()
+
+
 # Create an initial git_version.cpp file (that will be updated with latest git version)
 #==================================================================================================
 file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/git_version.cpp "")
--- a/toolchain-linux.cmake
+++ b/toolchain-linux.cmake
@@ -13,7 +13,7 @@ else()
 endif()
 
 if (NOT DEFINED ENV{CXXFLAGS})
-  set(CMAKE_CXX_FLAGS_DEBUG "-g -O1")
+  set(CMAKE_CXX_FLAGS_DEBUG "-g -O0")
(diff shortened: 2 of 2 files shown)


atulkulk (approved):

rahulvaidya20 (approved):

nileshnegi:
should be noted that `-O0` might not work on older ROCm or gfx targets.

thananon:
> should be noted that `-O0` might not work on older ROCm or gfx targets.

What?!

thananon:
requesting a re-review @alex-breslow-amd @nileshnegi @atulkulk

CMakeLists.txt
```diff
@@ -886,6 +886,15 @@ foreach(file ${GENERATED_FILES})
   list(APPEND HIP_SOURCES ${file})
 endforeach()
 
+# Apply -O1 optimization for generated files in debug builds
+# This is needed for device object to not go over the stacksize limit at link time while allowing
+# -O0 for host code in debug build.
+if(CMAKE_BUILD_TYPE MATCHES "Debug" AND NOT DEFINED ENV{CXXFLAGS})
+  message(STATUS "Setting -O1 optimization for device code in ${GEN_DIR} (Debug build)")
+  set_source_files_properties(${GENERATED_FILES} PROPERTIES COMPILE_OPTIONS "-O1")
```
alex-breslow-amd, line 894: I may have missed it, but why do we specify the flags for debug builds in two different places with different flags?  Why isn't there just one source of truth?

thananon: both install script and CMAKE will use CMakeList. I made a change here to affect both cases. I do not know why we use another file on top of CMakeList for install.sh.

alex-breslow-amd: In that case, my vote would be to simply delete the toolchain file to avoid confusion.  However, it doesn't have to be this PR.

nileshnegi: toolchain file was added as a ROCm-wide directive last year.

alex-breslow-amd: So then maybe toolchain should be the source of truth for all builds

nileshnegi: compilers folks suggested we use `-Xarch_host` and `-Xarch_device` to pass host vs device specific flags... however it fails for me... if this works, we can just add these to the toolchain file.

wenkaidu:
I got linker error with this commit with debug build. Previous commit was working fine.
```
lld: error: linking module flags 'Debug Info Version': IDs have conflicting override values in '/tmp/all_reduce_sum_bf16-gfx942-1d9b09.o' and 'ld-temp.o'
lld: error: linking module flags 'Debug Info Version': IDs have conflicting override values in '/tmp/device_table-gfx942-7f5df3.o' and 'ld-temp.o'
lld: error: linking module flags 'Debug Info Version': IDs have conflicting override values in '/tmp/host_table-gfx942-c72f03.o' and 'ld-temp.o'
lld: error: linking module flags 'Debug Info Version': IDs have conflicting override values in '/tmp/sendrecv_sum_i8-gfx942-53c195.o' and 'ld-temp.o'
```

PR #1959, opened 2025-10-07
MSCCL++ fix split path null deref

speriaswamy-amd:
## Details

SWDEV-556961

**What were the changes?**  

- Removed an unsafe, unused dereference of job->commId in ncclCommInitRankFunc that could null-deref in split path under -O0.
- In the split MSCCL++ path, replaced uses of *job->commId with the parent communicator’s ncclUniqueId via ncclCommToUniqueIdMap[job->parent].

- Split path does not allocate job->commId; dereferencing it is undefined behavior and can crash in true debug builds.
- MSCCL++ bookkeeping in split must be keyed by the parent’s ncclUniqueId, not by job->commId (which is null in split).

- Reproduced using a small program that creates a communicator and immediately calls ncclCommSplit, with MSCCL++ compiled in.
- Verified the unfixed code could be made to crash under -O0 when the dereference is actually used.
- After applying fixes, rebuilt and re-ran the reproducer with MSCCL++ enabled and -O0 host; no crash, child communicator initializes and is usable.

**Additional Documentation:**  
- No API or ABI changes; behavior is unchanged for users.
- The fix only touches internal init flow and MSCCL++ bookkeeping for split communicators.
- Tested with:
   - Debug build: host -O0, device -O1 (to avoid device stack overflows), MSCCL++ enabled at build time.
   - Reproducer: ncclCommInitRank then ncclCommSplit with RCCL_MSCCLPP_ENABLE=1.
   - Risk: low; removes UB and corrects map key for split branch.

## Approval Checklist


thananon (approved):

src/init.cc
```diff
@@ -2042,10 +2042,11 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
       comm->mscclppCompatible = true;
       comm->mscclpp_threshold = job->parent->mscclpp_threshold;
       comm->mscclpp_comm = job->parent->mscclpp_comm;
-      auto& mscclppUniqueId = mscclpp_uniqueIdMap[*job->commId];
-      mscclpp_uniqueIdMap[*job->commId] = mscclppUniqueId;
-      mscclpp_uniqueIdReverseMap[mscclppUniqueId].insert(*job->commId);
-      ncclCommToUniqueIdMap[comm] = *job->commId;
+      const ncclUniqueId& parentUniqueId = ncclCommToUniqueIdMap[job->parent];
+      auto& mscclppUniqueId = mscclpp_uniqueIdMap[parentUniqueId];
+      mscclpp_uniqueIdMap[parentUniqueId] = mscclppUniqueId;
```
PJAvinash: Is this line redundant ? value = map[key] ; map[key] = value;

speriaswamy-amd: Thanks @PJAvinash, good catch, the preceding line `auto& mscclppUniqueId = mscclpp_uniqueIdMap[parentUniqueId];` already ensures the key exists in the map and returns a reference to the stored value, the line `mscclpp_uniqueIdMap[parentUniqueId] = mscclppUniqueId;` is a no-op.

PJAvinash (approved):

PR #1972, opened 2025-10-10
Memory leak fixes in Init , group.cc 

PJAvinash:
## Details

**Work item:** _"Internal"

**What were the changes?**  

- Changed heap allocations for structs to thread local static structs 
- Context tracking add/drop matching in commAlloc, commFree

 To address memory leaks

N/A

**Additional Documentation:**  
Waiting for initial review and CI 

## Approval Checklist


src/init.cc
```diff
@@ -520,6 +519,10 @@ static ncclResult_t commFree(ncclComm_t comm) {
   free(comm->topParentRanks);
   free(comm->topParentLocalRanks);
   free(comm->gproxyConn);
+  if (rcclNeedEnableContextTrack(comm->cudaDev)) {
+    ncclCudaContextDrop(comm->context);
```
thananon: This is already done in `commFree()`. Is this change really needed or something you added just in case?

src/init.cc
```diff
@@ -533,8 +536,8 @@ static ncclResult_t commFree(ncclComm_t comm) {
   NCCLCHECK(ncclProfilerPluginFinalize(comm));
   NCCLCHECK(ncclNetFinalize(comm));
   // Disable until we validate NCCL_LAUNCH_IMPLICIT_ORDER support.
-  // but can be enabled via environment variable
-  if (rcclParamEnableContextTracking() == 1) {
+  // but enable for Radeon due to big impact on performance
+  if (rcclNeedEnableContextTrack(comm->cudaDev)) {
     ncclCudaContextDrop(comm->context);
```
thananon, line 538: like here?

thananon (approved):

src/group.cc
corey-derochie-amd, line 700: Does this allow the `groupJob` pointer to escape the function scope? If so, this may not be functionally equivalent. Original function `calloc`s a new `groupJob` on each call and emits it here or below. Your current solution gives the same `groupJob` pointer to all `job`s.

ammallya:
Imported to ROCm/rocm-systems

PR #1978, opened 2025-10-15
[Device] Adjust threadblock size for gfx950 to increase LL64/Simple performance for AR, RS and AG

mustafabar:
## Details

**What were the changes?**  

- This PR increases the threadblock size to 512 
- It also refactors how the number of threads per block is managed and propagated down to the barrier_generic, mainly to solve regression caused by falling back to software barrier for partial blocks, that uses atomics and spinning for synchronization. The issue happened due to hardcoding the threadblock size to NCCL_MAX_NTHREADS. So when a block is assigned lower number of threads, for example in the LL region, slower software barrier in unnecessarily used. 
- It adds several architecture-specific optimizations and safeguards for thread count selection in collective algorithms.

- Initial TransferBench data suggests that the default 256 threads does not achieve the peak BW for MI350
- Lay the foundation for optimizing block sizes based on recommendation from HW teams and empirical findings
- Up to 60% uplifts observed for ReduceScatter - bfloat16, and 48% for AllReduce - bfloat16  on MI350 (tested with xgmi speed 36 Gb/s) 

* Increased `NCCL_MAX_NTHREADS` and `NCCL_LL128_MAX_NTHREADS` to 512. Also, set `NCCL_LL_MAX_NTHREADS` to 256. (`src/include/device.h`) [[1]](diffhunk://#diff-d576984cb7701fcebb633b7c01fd980c6d9a39978143c0bb7907c992aca115a3R116-R124) [[2]](diffhunk://#diff-d576984cb7701fcebb633b7c01fd980c6d9a39978143c0bb7907c992aca115a3R141-R146)
* Updated kernel planning logic to use different thread block sizes depending on architecture, including clamping or adjusting `plan->threadPerBlock` for `__gfx950__` . (`src/enqueue.cc`) [[1]](diffhunk://#diff-adabec4bcb917d3813fc4640158f06f56ac041a8c3277efe7dd72ddaf2ea2e49L243-R247) [[2]](diffhunk://#diff-adabec4bcb917d3813fc4640158f06f56ac041a8c3277efe7dd72ddaf2ea2e49R871-R875) [[3]](diffhunk://#diff-adabec4bcb917d3813fc4640158f06f56ac041a8c3277efe7dd72ddaf2ea2e49L1203-R1213)
* Added logic to halve the number of threads for single-node collectives, and to override thread counts for reduce_scatter <= 512K per-rank size and LL protocols. (`src/enqueue.cc`)
* Added a `threadsPerBlock` member to all device primitive classes and initialized it from `blockDim.x`. This is to ensure the actual number of threads per block is used to trigger the hardware builtin for barrier. (`src/device/prims_ll.h`, `src/device/prims_ll128.h`, `src/device/prims_simple.h`) [[1]](diffhunk://#diff-44bc8b0bd04f9f42758fd8d5ed82dc60856b2227b7102352918eed676a00b323R27) [[2]](diffhunk://#diff-44bc8b0bd04f9f42758fd8d5ed82dc60856b2227b7102352918eed676a00b323L648-R649) [[3]](diffhunk://#diff-5eaba0c8c616dbbf3ab042bb05140aa0056e8534758b432f6f7e3837cd8e5723R38) [[4]](diffhunk://#diff-5eaba0c8c616dbbf3ab042bb05140aa0056e8534758b432f6f7e3837cd8e5723L580-R581) [[5]](diffhunk://#diff-d8e18ba4d99cf5bcfea19a9540ccb207ed517a0a27ca0afeabc7a92ba30bda85R54) [[6]](diffhunk://#diff-d8e18ba4d99cf5bcfea19a9540ccb207ed517a0a27ca0afeabc7a92ba30bda85L760-R761)
* Changed the barrier macro in `primitives.h` to use the new `threadsPerBlock` variable instead of the fixed `NCCL_MAX_NTHREADS`. (`src/device/primitives.h`)
* Adjusted tuning logic so that the maximum number of threads for LL protocol is now `NCCL_LL_MAX_NTHREADS`, which is set to 256, instead of always matching `NCCL_MAX_NTHREADS`. (`src/graph/tuning.cc`)

**Additional Documentation:**  

## Approval Checklist


src/device/prims_simple.h
```diff
@@ -51,6 +51,7 @@ class Primitives<
   int index; // Peer index I'm responsible for
   int flags;
   const int group;
+  const int threadsPerBlock;
```
alex-breslow-amd, line 54: I may have missed it, but where is this actually getting used?  Same for LL and LL128.  If we're not using it, my vote would be to delete it.  Update:  I see, in the barrier.  Hmmm.

src/device/primitives.h
```diff
@@ -15,9 +15,8 @@
 #include "common.h"
 
 #define NCCL_SPINS_BEFORE_CHECK_ABORT 10000
-
 #define barrier_generic(__THREAD_FENCE, NWORKERS, BARRIER_NEXT, BARRIERS_PTR) do { \
-  if (nthreads == NCCL_MAX_NTHREADS) { \
+  if (nthreads == threadsPerBlock) { \
```
alex-breslow-amd, line 20: blockDim.x?  Maybe just use it directly rather than adding an additional field to every version of the primitives.

src/enqueue.cc
```diff
@@ -2134,7 +2143,14 @@ static ncclResult_t topoGetAlgoInfo(
     info->nMaxChannels = nc;
   }
   if (info->algorithm == NCCL_ALGO_TREE) nt = NCCL_MAX_NTHREADS; // Tree now uses all threads always.
-  if (info->algorithm == NCCL_ALGO_PAT) nt = NCCL_MAX_NTHREADS;
+  if (info->algorithm == NCCL_ALGO_PAT)  nt = NCCL_MAX_NTHREADS;
+#if defined(__gfx950__)
+  if (comm->nNodes == 1) nt = NCCL_MAX_NTHREADS / 2; // For single node, we use half the number of threads for perf reasons.
```
alex-breslow-amd: Maybe put a #define in device.h, which is NCCL_MAX_NTHREADS_ONE_NODE and then reference it here, so 
if (comm ->nodes == 1) nt = NCCL_MAX_NTHREADS_ONE_NODE.  Or have a function even though it's a one liner.

I guess a general comment I have is that this logic is spread over multiple files, and there's no unifying function or logic gluing it together.  My concern is that this will make the code more difficult to maintain.  My preference would be to put as much of this logic as possible in one file and then import it where necessary.

alex-breslow-amd (approved):
Going to stamp.  Please consider some of my comments.

src/device/primitives.h
mustafabar: Using blockDim.x directly adds ~5us latency - my assumption is that const is triggering a compiler optimization that is not present when blockDim.x is read multiple times. I am interested in getting to the roots of this.

src/enqueue.cc
mustafabar: I had the same valid concern while working on it. I could spot at least 5 functions that adjust threads `topoGetAlgoInfo` `finishPlan` `scheduleCollTasksToPlan` `scheduleP2PTasksToPlan` and `ncclTopoTuneModel`. All of these control the final number of threads. I think we will have to do a refactor PR so that this logic is not spread out especially if we embrace this as an optimization variable. 
I added a macro `NCCL_SINGLE_NODE_MAX_NTHREADS`. Thanks for your feedback.

.gitignore
```diff
@@ -3,6 +3,7 @@
 /coverage/
 build/
 ext/
-
+logs
+sweep_rccl.sh
```
nileshnegi: do we need these?

src/enqueue.cc
```diff
@@ -241,8 +242,10 @@ static void finishPlan(struct ncclComm* comm, struct ncclKernelPlan* plan) {
   size_t workBytes = plan->workBytes;
   size_t batchBytes = plan->nWorkBatches*sizeof(struct ncclDevWorkBatch);
 
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
```
alex-breslow-amd, line 245: Why not just negate this to avoid the empty #if? Not a big deal.

alex-breslow-amd (approved):
LGTM

PedramAlizadeh (approved):

AbandiGa (approved):

src/enqueue.cc
mustafabar: We could do that, but I feel the or is easier to understand
Otherwise, we could have negation bugs like this one :) https://github.com/ROCm/rccl/blob/b703ffdfa4f565e85c4c11308c53b8963e203055/src/enqueue.cc#L1566

nileshnegi:
Merging as this PR has passed previous CI runs.

PR #1910, opened 2025-09-10
Fix issue where staging/mainline build commit hash doesn't match actual

thananon:
## Details

**What were the changes?**  
CMake git hash fetching function call should be made from RCCL source directory.

**Why were the changes made?**  
Our staging and release build versioning when set NCCL_DEBUG=version shows different git hash and we cannot trace back to our commit. This is due to build pipeline of our packaging team. 

**How was the outcome achieved?**  
Edit our CMake function.

**Additional Documentation:**  

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #1981, opened 2025-10-16
Fix git version fetching logic.

thananon:
## Details
Fix a bug where if the build directory is not RCCL source directory, the version info printout will be the git information of current build directory instead, as seen in mainline builds. We are currently taking git hash of CQE build script, resulting in mismatch and difficulties in debugging issues.

This has been attempted in #1910  but that is not sufficient.

**What were the changes?**  
Change CMake script to pass actual source directory and use git info from that directory.

Wrong version printout in staging/mainline/release build.

CMake passing directory as parameter.

Tested with custom cmake build directory and it prints correct version info:
```
RCCL version : 2.27.7-bugfix/git_version:7e287c5
HIP version  : 7.0.51831-a3e329ad8
ROCm version : 7.0.1.0-42-9428210
Hostname     : banff-pla-r27-14
Librccl path : /nfs/users/apatinya/build/librccl.so.1
```

## Approval Checklist

### DIFF
--- a/cmake/scripts/git_version.cmake
+++ b/cmake/scripts/git_version.cmake
@@ -19,8 +19,16 @@
 # SOFTWARE.
 
 # Attempt to collect the latest git hash
-set(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+# Use RCCL_SOURCE_DIR if passed, otherwise fallback to CMAKE_CURRENT_SOURCE_DIR
+if(NOT DEFINED RCCL_SOURCE_DIR)
+  set(RCCL_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+endif()
+if(NOT DEFINED RCCL_BINARY_DIR)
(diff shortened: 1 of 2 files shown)


nileshnegi (approved):

corey-derochie-amd (approved):

cmake/scripts/git_version.cmake
```diff
@@ -31,10 +39,12 @@ else()
   # Check for changes (denote with a '+') after hash
   execute_process(
     COMMAND bash -c "git diff --quiet --exit-code || echo +"
+    WORKING_DIRECTORY ${RCCL_SOURCE_DIR}
     OUTPUT_VARIABLE GIT_DIFF)
   # Collect branch information
   execute_process(
     COMMAND git rev-parse --abbrev-ref HEAD
+    WORKING_DIRECTORY ${RCCL_SOURCE_DIR}
```
alex-breslow-amd, line 47: Ah ok, I see this is the secret sauce.

alex-breslow-amd (approved):
LGTM

PR #1985, opened 2025-10-20
Fix protocol and channel override when tuner is used

nawrinsu:
## Details

**Work item:** _"Internal"._

**What were the changes?**  
_Skip updating protocol after tuner plugin finds a match in config file._
_Set number of channels to 0 in tuner plugin for proper fallback to default when channel is not set or no match found is config._

_To honor configuration provided by tuner plugin._

**Additional Documentation:**  

## Approval Checklist


ext-tuner/example/plugin.c
```diff
@@ -331,7 +331,7 @@ __hidden ncclResult_t pluginGetCollInfo(void* context, ncclFunc_t collType, size
   if (!ctx) return ncclInternalError;
 
   // Default channels
-  *nChannels = 1;
+  *nChannels = 0;
```
nusislam, line 334: Why is setting `nChannels=0` needed?

nawrinsu: In `getAlgoInfo` function channel count only falls back to default value when number of channels is 0. Setting `nChannels` to 1 bypass this fallback behavior and uses 1 channel when no channel config exists. See the code in https://github.com/ROCm/rccl/blob/develop/src/enqueue.cc#L2203. We have two options - either set `nChannels` to 0 or change the fallback logic.

nusislam:
@atulkulk, @Kapil-Shyam-Pawar - can you also review this? @Kapil-Shyam-Pawar - I saw that you added some tests for the plugin.

nusislam:
@nawrinsu - please pull latest develop.

ext-tuner/example/plugin.c
Kapil-Shyam-Pawar: This does make sense when the plugin finds no matching configuration.
Without this change, the plugin sets the channels to 1 by default, which may lead to a performance drop

Kapil-Shyam-Pawar: This also helps when using wildcard (-1) for channels in the csv conf. 
It will use the appropriate channel count based on topology-tuner, rather than forcing 1 channel.

Can we add a comment here? It would be really helpful.

nusislam (approved):

Kapil-Shyam-Pawar (approved):
LGTM. Please check if the README for the basic imp needs to be updated, as it mentions using 1 channel. Otherwise, good to go.

nileshnegi:
Failing CI `rccl_build_gfx90a` not related to PR. Safe to merge once the remaining CI pipelines pass.

PR #1990, opened 2025-10-21
disabling AR_Bias for gfx90a

mberenjk:
## Details

**Work item:** LWPCLPAT-599

**What were the changes?**  
disabling AR_Bias for gfx90a

The build time for AR_Bias was really long on gfx90a.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/generate.py
+++ b/src/device/generate.py
@@ -349,13 +349,20 @@ with open(os.path.join(gensrc, "device_table.h"), "w") as f:
   for fn in primary_funcs:
     sym = paste("_", "ncclDevFunc", *fn)
     if fn[2] == "LL128":
-      out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
+      if fn[5] == "0":
+        out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
+      else:
+        out("#if (defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
       out("%s %s();\n#else\n" % (func_declaration, sym))
       fn_ll = fn[:2] + ("LL",) + fn[3:]
       sym_ll = paste("_", "ncclDevFunc", *fn_ll)
       out("%s %s();\n#endif\n" % (func_declaration, sym_ll))
     else:
-      out("%s %s();\n" % (func_declaration, sym))
+      if fn[5] == "1":
+        out("#if (defined(__gfx942__) || defined(__gfx950__))\n")
+        out("%s %s();\n#endif\n" % (func_declaration, sym))
+      else:
+        out("%s %s();\n" % (func_declaration, sym))
   out("\n")
 
   out("typedef void(*ncclDevFuncPtr_t)();\n\n")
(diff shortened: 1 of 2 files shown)


src/device/generate.py
```diff
@@ -366,13 +373,20 @@ def custom_sort_key(fn):
     if unroll != "1": continue
     sym = paste("_", "ncclDevFunc", *fn)
     if fn[2] == "LL128":
-      out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
+      if fn[5] == "0":
```
alex-breslow-amd, line 376: Would you be to making this allReduceWithBiasEnabled(fn)?  I'm afraid that long term we're going to lose institutional memory of why this is here.  This looks very opaque.

allReduceWithBiasEnabled = (lambda fn: fn[5] == 0)  <--- Something like this.  It doesn't have to be a lambda function, arguably better if it isn't.

alex-breslow-amd: In general, these subscripts probably should have variable names instead of just numbers.  If we used a named tuple or DataClass, we could just to fn.datatype instead of fn[2].  That seems a lot cleaner to me.  That also would simplify the code above in my previous comment and prevent certain classes of bugs when we add more fields.

src/device/generate.py
```diff
@@ -400,13 +421,20 @@ def custom_sort_key(fn):
     if unroll != "4": continue
     sym = paste("_", "ncclDevFunc", *fn)
     if fn[2] == "LL128":
-      out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
+      if fn[5] == "0":
+        out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__)) && defined(ENABLE_LL128)\n")
```
alex-breslow-amd, line 425: These also feel like they could be candidates for using some "||".join operation.  This is fine though.  I can create another task to clean up generate.py.

src/device/generate.py
```diff
@@ -612,14 +640,22 @@ def partition_by_name(fns):
...
+          out("#if (defined(__gfx90a__) || defined(__gfx942__) || defined(__gfx950__))\n")
       out(
         "DEFINE_ncclDevFunc({sym}, ncclFunc{coll}, {redop_cxx}, {ty_cxx}, NCCL_ALGO_{algo}, NCCL_PROTO_{proto}, {acc}, {pipeline}, {unroll})\n"
         .format(sym=sym, coll=coll, redop_cxx=redop_to_cxx[redop], ty_cxx=ty_to_cxx[ty],
                 algo=(algo or "RING"), proto=(proto or "SIMPLE"), acc=acc, pipeline=pipeline, unroll=unroll)
       )
       if proto == "LL128":
         out("#endif\n")
+      if acc == "1" and proto != "LL128":
+        out("#endif\n")
```
corey-derochie-amd, line 658: This should work, no?
```suggestion
      if proto == "LL128" or acc == "1":
        out("#endif\n")
```

corey-derochie-amd:
I don't *think* this is problem because we are now using a keyed map instead of an index-based array, but:
Is there any chance that conditionally including these kernels could somehow mess up the lookup to cause a mismatch, which could explain why some unit tests were failing?

nileshnegi:
Closing this PR, as https://github.com/ROCm/rccl/pull/2030 resolves this issue.

PR #1993, opened 2025-10-21
Added a Process Isolated Test Runner

atulkulk:
## Details

**Work item:** Internal

**What were the changes?**  
A lightweight C++ testing framework that executes Google Test cases in separate processes, ensuring complete isolation between tests.

RCCL code extensively uses static variables for configuration caching (e.g., rcclSetP2pNetChunkSize). In traditional testing, static variables persist across test cases, causing tests to pollute each other's state and produce unreliable results. This is particularly problematic when testing:
- Environment variable behavior
- Architecture-specific logic
- One-time initialization patterns
- Configuration changes

- Uses fork() for process isolation
- Hooked into rccl-UnitTestsFixtures binary

**Additional Documentation:**  
Read test/README.md for more information.

## Approval Checklist

### DIFF
--- a/test/common/ProcessIsolatedTestRunner.hpp
+++ b/test/common/ProcessIsolatedTestRunner.hpp
@@ -131,7 +131,7 @@ private:
   static void applyEnvironmentVariables(const TestConfig &config);
 
   /**
-   * @brief Execute a single test in the current process
+   * @brief Execute a single test in the child process
    * @param config Test configuration
    * @return Exit code (0 for success, non-zero for failure)
    */
@@ -172,6 +172,8 @@ public:
    * @brief Execute all registered tests sequentially
    * @param options Execution options (defaults to continue on failure)
    * @return True if all tests passed, false otherwise
+   * @note This method automatically clears all test registrations and results
+   *       after execution, ensuring a clean state for the next test suite.
    */
   static bool
   executeAllTests(const ExecutionOptions &options = ExecutionOptions());
@@ -191,6 +193,10 @@ public:
 
   /**
    * @brief Clear test registry and results (thread-safe)
+   * @note Calling this method manually is typically not necessary, as
+   *       executeAllTests() automatically clears registrations after execution.
+   *       This method is primarily useful for advanced use cases or when tests
+   *       are registered but not executed.
    */
   static void clear();
 
--- a/test/common/ProcessIsolatedTestRunner.hpp
+++ b/test/common/ProcessIsolatedTestRunner.hpp
@@ -121,6 +121,15 @@ public:
     };
 
 private:
+    /**
+     * @brief Structure to hold captured process output
+     */
+    struct CapturedOutput
+    {
+        std::string stdoutContent; ///< Captured stdout content
+        std::string stderrContent; ///< Captured stderr content
+    };
+
     // Thread-safe static members for test management
     static std::mutex              testConfigsMutex_;
     static std::vector<TestConfig> testConfigs_;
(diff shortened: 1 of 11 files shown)


venksubramd:
Nice work on this one @atulkulk 

I really like the builder pattern to configure the process, for example where a call to registerTest(...) is followed with .withEnvironment(), .withTimeout(), and so on.

Also, really nice documentation in ProcessIsolatedTestRunner.md. It was very useful and easy to follow. Thanks for this.

A few of questions:

1. In ProcessIsolatedTestRunner.md, in the "Why use Process Isolation?" section, this may be me misunderstanding, but, 

setenv("NCCL_P2P_NET_CHUNKSIZE", "12345", 1);
  // Static variable gets set to 12345

says the static variable is being set, but I do not see how the p2pNetChunkSize variable is being set by that call. My lack of understanding comes from the fact that p2pNetChunkSize is being set to value returned by a calculateValue() call. I may be missing the link between createValue() and the setting of the above environment variable. Some clarification for the example in this section will be helpful.

2. Looking at the file RcclWrapTests.cpp:

In TEST(Rcclwrap, RcclUpdateThreadThreshold_UserEnvSet), the test is being registered with a call to registerTest, but I do not see a call to executeAllTests(). Is that required or does the test run without an explicit call to it. If it is the latter, it would be useful to clarify that in the documentation.

3. I really like the emphasis in the "Best Practices" section to call clear before any registration. When clear is called, if there are tests that were registered but not executed or registered by not completed, it would be useful to know that. May be an error code? Or may be a log message? It would be useful to consider the situation where tests were registered by were not executed.

Alternatively, if there is an approach to check if not all registered tests were executed will be helpful.

atulkulk:
> Nice work on this one @atulkulk
> 
> I really like the builder pattern to configure the process, for example where a call to registerTest(...) is followed with .withEnvironment(), .withTimeout(), and so on.
> 
> Also, really nice documentation in ProcessIsolatedTestRunner.md. It was very useful and easy to follow. Thanks for this.
> 
> A few of questions:
> 
> 1. In ProcessIsolatedTestRunner.md, in the "Why use Process Isolation?" section, this may be me misunderstanding, but,
> 
> setenv("NCCL_P2P_NET_CHUNKSIZE", "12345", 1); // Static variable gets set to 12345
> 
> says the static variable is being set, but I do not see how the p2pNetChunkSize variable is being set by that call. My lack of understanding comes from the fact that p2pNetChunkSize is being set to value returned by a calculateValue() call. I may be missing the link between createValue() and the setting of the above environment variable. Some clarification for the example in this section will be helpful.
-  done (elaborated examples in README.md) 
> 2. Looking at the file RcclWrapTests.cpp:
> 
> In TEST(Rcclwrap, RcclUpdateThreadThreshold_UserEnvSet), the test is being registered with a call to registerTest, but I do not see a call to executeAllTests(). Is that required or does the test run without an explicit call to it. If it is the latter, it would be useful to clarify that in the documentation.
- done
> 3. I really like the emphasis in the "Best Practices" section to call clear before any registration. When clear is called, if there are tests that were registered but not executed or registered by not completed, it would be useful to know that. May be an error code? Or may be a log message? It would be useful to consider the situation where tests were registered by were not executed.
> 
> Alternatively, if there is an approach to check if not all registered tests were executed will be helpful.
- done

@venksubramd I have addressed all your comments.

venksubramd:
Nice changes. I especially like the warning for unexecuted tests when clear is called.

test/common/ProcessIsolatedTestRunner.hpp
```diff
@@ -0,0 +1,204 @@
...
+  static std::vector<TestResult> testResults_;
+
+  /**
+   * @brief Apply environment variables to current process
+   * @param config Test configuration containing environment settings
+   */
+  static void applyEnvironmentVariables(const TestConfig &config);
+
+  /**
+   * @brief Execute a single test in the current process
```
corey-derochie-amd: This doc disagrees with the header comment in the cpp. Is the test run in the current process or in a separate process?

atulkulk: This method is called in the child process [(ref)](https://github.com/ROCm/rccl/pull/1993/files/02317630e8bfb4fac38597afc5622a7bd555b878#diff-498d12a8be44ff6689f057ceed7635627a30b32a73418ff7455e213056a0d8b1R274) after fork, hence the text as `in the current process`. But for brevity, I will add more clarity in the comment.

atulkulk: Addressed.

corey-derochie-amd (approved):
lgtm!

Kapil-Shyam-Pawar (approved):

PR #1996, opened 2025-10-21
Enable MPI support to execute MPI specific unit/functional tests

atulkulk:
## Details

**Work item:** Internal

**What were the changes?**  
A simple C++ testing framework for multi-process RCCL testing using MPI and Google Test integration.

Testing RCCL at unit and functional levels requires multiple processes with GPU-to-GPU data transfer to properly validate collective operations (AllReduce, Broadcast), point-to-point communication, and even specific features such as transport mechanism, etc.
Features
- Multi-process distributed testing: Each MPI rank runs in its own process with dedicated GPU
- Automatic RCCL communicator management: Test-specific communicators for complete isolation between tests
- Multi-node support: Works seamlessly on single-node (shared memory) or multi-node (network) configurations
- Google Test integration: Familiar test patterns with TEST_F and assertion macro
- Added provision to include non-GTest (standalone) tests

- MPITestCore: Base class for GTest and non-GTest test cases
- MPITestBase: Base class providing test infrastructure (located in test/common/MPITestBase.hpp/cpp)
- MPIEnvironment: Global MPI environment for initialization and cleanup

**Additional Documentation:**  
Read test/common/MPITestRunner.md for more information.

## Approval Checklist


venksubramd:
I think this one is pretty solid. I really like the documentation and the descriptions as well.

One unrelated observation, going over the tests in P2pMPITests.cpp that use the MPI test running:
There are several places where we are cleaning up memory (as we should) but that's happening after ASSERT_EQ. If the assert were to fail, these clean up steps would not run. Do we need to revisit these and ensure that we have proper clean up irrespective of the termination of a test due to assertion failures?

atulkulk:
> I think this one is pretty solid. I really like the documentation and the descriptions as well.
> 
> One unrelated observation, going over the tests in P2pMPITests.cpp that use the MPI test running: There are several places where we are cleaning up memory (as we should) but that's happening after ASSERT_EQ. If the assert were to fail, these clean up steps would not run. Do we need to revisit these and ensure that we have proper clean up irrespective of the termination of a test due to assertion failures?

@venksubramd I have addressed your comment. Added resource guards to handle cleaning up of the memory.

corey-derochie-amd:
@eidenyoshida this PR only adds unit tests, no effect on performance. Regression testing not required.

nileshnegi:
> @eidenyoshida this PR only adds unit tests, no effect on performance. Regression testing not required.

@corey-derochie-amd this PR was randomly selected just for testing the regression detector pipeline.

thananon (approved):

venksubramd:
Nice work, addresses my previous comments. Thanks.

Kapil-Shyam-Pawar (approved):

test/common/DeviceBufferHelpers.hpp
```diff
@@ -0,0 +1,456 @@
...
+    static constexpr const char*    name  = "uint64_t";
+};
+
+/** @brief Specialization for int8_t */
+template<>
+struct NcclTypeTraits<int8_t>
+{
+    static constexpr ncclDataType_t value = ncclInt8;
+    static constexpr const char*    name  = "int8_t";
+};
```
alex-breslow-amd: This is a lot of boilerplate.  I'm wondering there's a way to reduce it so that there's less to maintain.  You might be able to call ```typeid(T).name()``` to get the const char* name.  Similarly, it feels to me that there's probably some preprocessor magic that could likely be used to map something like uint64_t to ncclUint64.  Just a thought here.  Not sure how easy it is.

test/common/DeviceBufferHelpers.hpp
```diff
@@ -0,0 +1,456 @@
...
+ *
+ * @tparam T Element type (float, int, etc.)
+ * @param device_buffer Device memory pointer (from hipMalloc)
+ * @param num_elements Number of elements
+ * @param rank MPI rank (used in pattern generation)
+ * @param multiplier Pattern multiplier (default: 1000)
+ * @return hipError_t from hipMemcpy, or hipSuccess
+ */
+template<typename T>
+hipError_t initializeBufferWithPattern(void*  device_buffer,
```
alex-breslow-amd: Why not implement this by calling initializeBufferWithCustomPattern with the appropriate lambda?  This is probably fine, just seems like a lot of duplicate code.

test/common/DeviceBufferHelpers.hpp
```diff
@@ -0,0 +1,456 @@
...
+                               hipMemcpyDeviceToHost);
+    if(err != hipSuccess)
+    {
+        return false;
+    }
+
+    // Verify samples
+    for(size_t i = 0; i < num_samples; i++)
+    {
+        T expected = static_cast<T>(expected_rank * multiplier + i);
```
alex-breslow-amd: I'm wondering if this should be a pluggable function so that you can patch in different behavior.  Fine for now.

test/common/DeviceBufferHelpers.hpp
```diff
@@ -0,0 +1,456 @@
...
+    if(err != hipSuccess)
+    {
+        return {err, {}};
+    }
+
+    // Create host data with pattern
+    std::vector<T> host_data(num_elements);
+    for(size_t i = 0; i < num_elements; i++)
+    {
+        host_data[i] = static_cast<T>(rank * multiplier + i);
```
alex-breslow-amd: It feels like you should be calling initializeBufferWithPatternhere to improve code reuse or the custom version of it.

alex-breslow-amd (approved):
Going to stamp to unblock.  In the future, if there's a way that you could make smaller PRs, it would help me to review the code.  This is thousands of line of code in a single PR, which makes it very difficult to review.

atulkulk:
> Going to stamp to unblock. In the future, if there's a way that you could make smaller PRs, it would help me to review the code. This is thousands of line of code in a single PR, which makes it very difficult to review.

Yes, I agree, and will plan it better to have smaller PRs going forward. I do not like to have bigger PRs myself for reviews.

test/common/DeviceBufferHelpers.hpp
atulkulk: Addressed with a macro.

test/common/DeviceBufferHelpers.hpp
atulkulk: Addressed.

test/common/DeviceBufferHelpers.hpp
atulkulk: Addressed.

test/common/DeviceBufferHelpers.hpp
atulkulk: Addressed.

PR #1998, opened 2025-10-23
[Net plugin] Enable capping the number of QPs created for send/recv colls

AbandiGa:
Work item:
Internal, LWPCLPAT-594

What were the changes?
This work introduces selective QP-per-connection tuning for P2P collectives versus non-P2P collectives. This approach allows using a higher or customized number of QPs for non-P2P collectives and making more effective use of NCCL_IB_QPS_PER_CONNECTION.

Why were the changes made?
Transport layer 

How was the outcome achieved?
More information about the code design -> LWPCLPAT-594

Additional Information
RCCL_IB_QPS_PER_P2P enables separate QP control: When set (≥1), P2P operations (Send/Recv) use RCCL_IB_QPS_PER_P2P while collectives (AllReduce/AllGather/etc.) use NCCL_IB_QPS_PER_CONNECTION. When not set: NCCL_IB_QPS_PER_CONNECTION controls all operations QPs (P2P and collectives), maintaining NCCL original behavior where one parameter applies to everything including P2P operations. 
## Approval Checklist

### DIFF
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -774,6 +774,11 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   commConfig.trafficClass = req->trafficClass == NCCL_CONFIG_UNDEF_INT ? NCCL_NET_TRAFFIC_CLASS_UNDEF : req->trafficClass;
   NCCLCHECK(ncclNetGetDeviceHandle(resources->netDeviceType, resources->netDeviceVersion, false /*isRecv*/, &resources->netDeviceHandle));
   bool rccl_anp = !(strcmp(proxyState->ncclNet->name, RCCL_ANP_PLUGIN_STR));
+  // Call plugin-specific hook to encode P2P policy into handle
+  if (ncclIbNetEncodeP2pPolicy) {
+    // IB plugin provides strong symbol implementation
+    ncclIbNetEncodeP2pPolicy(req->handle, resources->isP2p);
+  }
   if (resources->shared) {
     // Shared buffers
     struct ncclProxyProgressState* progressState = &proxyState->progressState;
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -30,6 +30,11 @@
 static_assert(sizeof(ncclNetHandle_t) <= CONNECT_SIZE, "NET Connect info is too large");
 
 #define RCCL_ANP_PLUGIN_STR  "RCCL-ANP"
+// Default weak implementation - no-op if no plugin provides strong symbol
+__attribute__((weak))
+int ncclNetEncodeP2pPolicy(void* handle, int isP2p) {
+  return 0;  // Not handled
+}
 
 #define NCCL_NET_MAP_HOSTMEM 0
 #define NCCL_NET_MAP_DEVMEM 1
@@ -775,9 +780,9 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   NCCLCHECK(ncclNetGetDeviceHandle(resources->netDeviceType, resources->netDeviceVersion, false /*isRecv*/, &resources->netDeviceHandle));
   bool rccl_anp = !(strcmp(proxyState->ncclNet->name, RCCL_ANP_PLUGIN_STR));
   // Call plugin-specific hook to encode P2P policy into handle
-  if (ncclIbNetEncodeP2pPolicy) {
+  if (ncclNetEncodeP2pPolicy) {
     // IB plugin provides strong symbol implementation
-    ncclIbNetEncodeP2pPolicy(req->handle, resources->isP2p);
+    ncclNetEncodeP2pPolicy(req->handle, resources->isP2p);
   }
   if (resources->shared) {
     // Shared buffers
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -224,6 +224,7 @@ static ncclResult_t sendSetup(struct ncclComm* comm, struct ncclTopoGraph* graph
   req.connIndex = connIndex;
   req.curr_hdp_reg = 0;
   req.netDev = -1;
+  // Determine if this is a P2P connection or not based on the graph pointer
   if(graph == NULL) {
     req.isP2p = 1;
   } else {
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -776,7 +776,7 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   bool rccl_anp = !(strcmp(proxyState->ncclNet->name, RCCL_ANP_PLUGIN_STR));
   
   if (rcclNetP2pPolicy) {
-    rcclNetP2pPolicy(req->handle, resources->isP2p);
+    NCCLCHECK(rcclNetP2pPolicy(req->handle, resources->isP2p));
   }
   
   if (resources->shared) {
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -122,6 +122,7 @@ struct sendNetResources {
   ncclNetDeviceHandle_t* netDeviceHandle;
   size_t maxP2pBytes;
   volatile uint32_t* curr_hdp_reg;  // Curr GPU in ring (for rdma transport use only)
+  int isP2p;
 };
 
 struct recvNetResources {
@@ -194,6 +195,7 @@ struct setupReq {
   int channelId;
   int connIndex;
   uint32_t* curr_hdp_reg;
+  int isP2p;
 };
 
 NCCL_PARAM(NetOptionalRecvCompletion, "NET_OPTIONAL_RECV_COMPLETION", 1);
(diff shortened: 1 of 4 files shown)


src/transport/net_ib.cc
```diff
@@ -969,6 +969,7 @@ ncclResult_t ncclIbGetProperties(int dev, ncclNetProperties_t* props) {
 static_assert(MAX_REQUESTS <= 256, "request id are encoded in wr_id and we need up to 8 requests ids per completion");
 
 #define NCCL_IB_MAX_QPS 128
+#define P2P_MAX_QPS 1
```
mustafabar: Please make this environment controlled `RCCL_PARAM` with default value set to 1. `RCCL_IB_QPS_PER_P2P`

src/transport/net_ib.cc
```diff
@@ -1422,10 +1426,20 @@ ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHan
...
-  localNqps  = ncclParamIbQpsPerConn() * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
-  remoteNqps = ncclParamIbQpsPerConn() * remoteVProps.ndevs;
-  comm->base.nqps = remoteNqps > localNqps ? remoteNqps : localNqps; // Select max nqps (local or remote)
 
+  // Read isP2p from handle
+  isP2p = handle->isP2p;
+  INFO(NCCL_NET, "NET/IB: ncclIbConnect isP2p=%d", isP2p);
+  
+  if(isP2p) {
+    localNqps  = P2P_MAX_QPS * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
```
mustafabar: this should use the rcclParam instead

mustafabar (changes requested):

mustafabar (approved):

mustafabar:
pls update CHANGELOG.md

src/transport/net_ib.cc
```diff
@@ -1422,9 +1426,24 @@ ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHan
   mergedDev = ncclIbMergedDevs + dev;
   comm->base.vProps = mergedDev->vProps;
   int localNqps, remoteNqps;
-  localNqps  = ncclParamIbQpsPerConn() * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
-  remoteNqps = ncclParamIbQpsPerConn() * remoteVProps.ndevs;
+
+  // Read isP2p from handle
+  isP2p = handle->isP2p;
+  INFO(NCCL_NET, "NET/IB: ncclIbConnect isP2p=%d", isP2p);
+  if (rcclParamIbQpsPerP2p() > 0) {
```
mustafabar: simplify condition 
```suggestion
  if (rcclParamIbQpsPerP2p() > 0 && isP2P) {
    //use IB_QPS_PER_P2P
  } else {
    // use IB_QPS_PER_CONNECTION
  }
```

src/transport/net_ib.cc
```diff
@@ -1744,9 +1758,24 @@ ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle
   struct ncclIbDevInfo* remDevInfo;
   struct ncclIbQp* qp;
   bool useDmaBuf; 
+  int localNqps, remoteNqps;
 
   mergedDev = ncclIbMergedDevs + lComm->dev;
   rComm->base.nRemDevs = remMeta.ndevs;
+  if (rcclParamIbQpsPerP2p() > 0) {
```
mustafabar: check above

mustafabar (changes requested):

src/transport/net_ib.cc
```diff
@@ -2717,6 +2747,12 @@ ncclResult_t ncclIbCloseListen(void* listenComm) {
   return ncclSuccess;
 }
 
+int rcclNetP2pPolicy(void* handle, int isP2p) {
+  struct ncclIbHandle* ibHandle = (struct ncclIbHandle*)handle;
+  ibHandle->isP2p = isP2p;
+  return 1;  // Handled
```
mustafabar: there's no meaning for this return, can you check if `ncclIbHandle` is not null before dereferencing it? If null return ncclInvalidArg..
use `ncclResult_t` for return

also you might want to use the `magic` field to see if it is a valid address.

mustafabar (approved):

src/transport/net_ib.cc
```diff
@@ -1423,9 +1427,19 @@ ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHan
...
+  // Read isP2p from handle
+  isP2p = handle->isP2p;
+  INFO(NCCL_NET, "NET/IB: ncclIbConnect isP2p=%d", isP2p);
+  if (rcclParamIbQpsPerP2p() > 0 && isP2p) {
+    localNqps  = rcclParamIbQpsPerP2p() * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
+    remoteNqps = rcclParamIbQpsPerP2p() * remoteVProps.ndevs;
+  } else {
+    localNqps  = ncclParamIbQpsPerConn() * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
+    remoteNqps = ncclParamIbQpsPerConn() * remoteVProps.ndevs;
+  }
```
alex-breslow-amd: This identical to the code below from what I can tell.  Can you encapsulate this in a function?

src/transport/net_ib.cc
```diff
@@ -1745,9 +1754,19 @@ ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle
...
 
   mergedDev = ncclIbMergedDevs + lComm->dev;
   rComm->base.nRemDevs = remMeta.ndevs;
+  if (rcclParamIbQpsPerP2p() > 0 && remMeta.isP2p) {
+    localNqps  = rcclParamIbQpsPerP2p() * rComm->base.vProps.ndevs;
+    remoteNqps = rcclParamIbQpsPerP2p() * remMeta.ndevs;
+  } else {
+    localNqps  = ncclParamIbQpsPerConn() * rComm->base.vProps.ndevs;
+    remoteNqps = ncclParamIbQpsPerConn() * remMeta.ndevs;
+  }
```
alex-breslow-amd: Duplicate code from above.  If these are the same, they should call a function.

src/transport/net.cc
```diff
@@ -222,6 +224,11 @@ static ncclResult_t sendSetup(struct ncclComm* comm, struct ncclTopoGraph* graph
   req.connIndex = connIndex;
   req.curr_hdp_reg = 0;
   req.netDev = -1;
+  if(graph == NULL) {
+    req.isP2p = 1;
+  } else {
+    req.isP2p = 0;
+  }
```
nileshnegi, line 232: could you pls add a comment for this condition?

src/transport/net_ib.cc
```diff
@@ -2718,6 +2738,14 @@ ncclResult_t ncclIbCloseListen(void* listenComm) {
   return ncclSuccess;
 }
 
+ncclResult_t rcclNetP2pPolicy(void* handle, int isP2p) {
```
alex-breslow-amd, line 2736: It's gross that this is a void*, but I see it comes from the code we've inherited, so ok.

src/transport/net_ib.cc
```diff
@@ -1423,9 +1427,19 @@ ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHan
   mergedDev = ncclIbMergedDevs + dev;
   comm->base.vProps = mergedDev->vProps;
   int localNqps, remoteNqps;
-  localNqps  = ncclParamIbQpsPerConn() * comm->base.vProps.ndevs; // We must have at least 1 qp per-device
-  remoteNqps = ncclParamIbQpsPerConn() * remoteVProps.ndevs;
+
+  // Read isP2p from handle
+  isP2p = handle->isP2p;
+  INFO(NCCL_NET, "NET/IB: ncclIbConnect isP2p=%d", isP2p);
+  if (rcclParamIbQpsPerP2p() > 0 && isP2p) {
```
alex-breslow-amd: I think you can simplify this logic a bit.  Basically, just do this:
```
auto qp_multiplier = (rcclParamIbQpsPerP2p() > 0 && isP2p) ? rcclParamIbQpsPerP2p() : ncclParamIbQpsPerConn();
localNqps = qp_multiplier * comm->base.vProps.ndevs;
remoteNqps = qp_multiplier * remoteVProps.ndevs;
```

I'll let you decide if you think this is cleaner or not.

alex-breslow-amd (approved):
Thanks for the updates

src/transport/net_ib.cc
```diff
@@ -151,6 +153,18 @@ static void ncclIbQpFatalError(struct ibv_qp* qp) {
 static void ncclIbCqFatalError(struct ibv_cq* cq) {
   ncclIbStatsFatalError((struct ncclIbStats*)cq->cq_context);
 }
+// Calculate number of QPs based on P2P flag and device counts
+static int ncclIbCalculateNqps(int isP2p, int localNdevs, int remoteNdevs, const char* funcName) {
+  auto qp_multiplier = (rcclParamIbQpsPerP2p() > 0 && isP2p) ?
```
alex-breslow-amd, line 158: I had a question here.  Will this have spillover into our direct AllGather since it is P2P?

AbandiGa: Yes, direct AllGather will use RCCL_IB_QPS_PER_P2P if it's set to a value > 0. Note that the default is 0, so by default it will fall back to using NCCL_IB_QPS_PER_CONNECTION.

PR #1999, opened 2025-10-24
Empty kernel test enhancements [tools]

PJAvinash:
## Details
Added option to collect  per iteration CPU, GPU and outer loop timing information
**Work item:** "Internal"

**What were the changes?**  
Added additional args with default values and logic to collect timing information

To improve user experience collecting EmptyKernelTest data, and isolating CPU and GPU duration collecting paths to minimize the over head from each other.

Minimal code changes

**Additional Documentation:**  

- CPU timing can get affected slightly by first loop iteration because branch predictors might need to warm up. But it is negligible as per experiment data so far.
- In cpuTime:1 gpuTime:0 outerLoop:1 case, the first CPU iteration duration seems to longer than avg. The reason for this is not very clear ( still the data generated is valid)
- Isolating GPU timing code path and CPU timing code path seems to give significantly different results. 

## Approval Checklist

### DIFF
--- a/tools/EmptyKernelTest/EmptyKernelTest.cpp
+++ b/tools/EmptyKernelTest/EmptyKernelTest.cpp
@@ -85,13 +85,18 @@ int main(int argc, char **argv)
   int numIterations = (argc > 1 ? atoi(argv[1]) : 10);
   int gridSize      = (argc > 2 ? atoi(argv[2]) : 1);
   int blockSize     = (argc > 3 ? atoi(argv[3]) : 1);
+  bool cpuTime      = (argc > 4 ? atoi(argv[4]) != 0 : true);
+  bool gpuTime      = (argc > 5 ? atoi(argv[5]) != 0 : true);
+  bool outerLoop    = (argc > 6 ? atoi(argv[6]) != 0 : false); // only outer loop
   int numWarmups    = 3;
-  printf("Running %d iterations <<<%d,%d>>>\n", numIterations, gridSize, blockSize);
+  printf("Running %d iterations <<<%d,%d>>> cpuTime:%d gpuTime:%d outerLoop:%d\n", numIterations, gridSize, blockSize, cpuTime , gpuTime, outerLoop);
 
   // Create events and stream
-  hipEvent_t startEvent, stopEvent;
+  hipEvent_t startEvent, stopEvent, startEventOuterLoop, stopEventOuterLoop;
   HIP_CALL(hipEventCreate(&startEvent));
   HIP_CALL(hipEventCreate(&stopEvent));
+  HIP_CALL(hipEventCreate(&startEventOuterLoop));
+  HIP_CALL(hipEventCreate(&stopEventOuterLoop));
   hipStream_t stream;
   HIP_CALL(hipStreamCreate(&stream));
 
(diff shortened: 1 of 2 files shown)


thananon (approved):

tools/EmptyKernelTest/EmptyKernelTest.cpp
```diff
@@ -108,29 +113,51 @@ int main(int argc, char **argv)
...
+  double cpuDeltaMsecOuterLoop;
+  float gpuDeltaMsecOuterLoop;
+  if(outerLoop) { HIP_CALL(hipEventRecord(startEventOuterLoop, stream));}
   for (int iteration = 0; iteration < numIterations; iteration++)
   {
     // Start timing
-    auto cpuStart = std::chrono::high_resolution_clock::now();
-    HIP_CALL(hipEventRecord(startEvent, stream));
+    if(cpuTime){cpuStart = std::chrono::high_resolution_clock::now();}
+    if(gpuTime){HIP_CALL(hipEventRecord(startEvent, stream));}
```
nusislam: outerloop and cputime/gpuTime should not be enabled at the same time. Similarly, cputime and gputime should not be enabled at the same time.

PJAvinash: ./EmptyKernelTest 10 1 1 **0 0 1**
Running 10 iterations <<<1,1>>> cpuTime:0 gpuTime:0 outerLoop:1 

Each of the 8 combinations  [0,1]x[0,1]x[0,1] can be called as needed.

nusislam: Please add a warning/message saying that different timing types should not be enabled at the same time.

tools/EmptyKernelTest/EmptyKernelTest.cpp
```diff
@@ -108,29 +113,51 @@ int main(int argc, char **argv)
...
+      allCpuDeltaMsec[iteration] = cpuDeltaMsec * 1000.0;
+      cpuSum += cpuDeltaMsec * 1000.0;
+    }
+  }
+  if(outerLoop) {
+    cpuDeltaOuterloop = std::chrono::high_resolution_clock::now() - cpuStartOuterloop;
+    HIP_CALL(hipEventRecord(stopEventOuterLoop, stream));
+    HIP_CALL(hipStreamSynchronize(stream));
+    HIP_CALL(hipEventElapsedTime(&gpuDeltaMsecOuterLoop, startEventOuterLoop, stopEventOuterLoop));
+    cpuDeltaMsecOuterLoop = std::chrono::duration_cast<std::chrono::duration<double>>(cpuDeltaOuterloop).count() * 1000.0;
```
nusislam, line 168: Here you should report the average time per iteration.

PJAvinash: Addressed in the below code.

tools/EmptyKernelTest/EmptyKernelTest.cpp
PJAvinash: Added warning messages, and run.sh script to generate data only for meaningful combinations. Decoupled inner and out loops, It's on user to force certain combinations. Warning message will be printed if the combination falls outside the recommended ones.

tools/EmptyKernelTest/EmptyKernelTest.cpp
```diff
@@ -108,50 +116,83 @@ int main(int argc, char **argv)
   // NOTE: Timing is done per-iteration, instead of batching multiple iterations
   double cpuSum = 0.0;
   double gpuSum = 0.0;
+  auto cpuStart = std::chrono::high_resolution_clock::now();
+  auto cpuDelta = std::chrono::high_resolution_clock::now() - cpuStart;
```
alex-breslow-amd, line 121: You likely want to put some fencing around this to prevent code being reordered and sandwiched between these two statements.  See https://theunixzoo.co.uk/blog/2021-10-14-preventing-optimisations.html.  Let me know if you have any questions.  Another option is sandwiching cpuStart and cpuDelta between three std::atomic_signal_fence calls.

tools/EmptyKernelTest/EmptyKernelTest.cpp
```diff
@@ -108,50 +116,83 @@ int main(int argc, char **argv)
...
+  double cpuDeltaMsec;
+  float gpuDeltaMsec;
+
+  auto cpuStartOuterloop = std::chrono::high_resolution_clock::now();
+  auto cpuDeltaOuterloop = std::chrono::high_resolution_clock::now() - cpuStartOuterloop;
```
alex-breslow-amd, line 126: Comment above applies here too.

alex-breslow-amd:
My main comments pertain to timing.  When you insert these timing calls, make sure that the compiler can't reorder code past them.  Details are in an inline code comment.

tools/EmptyKernelTest/EmptyKernelTest.cpp
nusislam: This is host-side code. Still need additional instructions to prevent reordering?

PJAvinash: The run to run variance seems to be larger than effects of instruction reordering & out of order loads/stores from CPU side. I'll  still add a new commit to address these.

alex-breslow-amd: Yes, what I reference applies to host-side code.  The compiler is free to reorder instructions between std::chrono calls.  They are not implicit instruction barriers.

nusislam (approved):

tools/EmptyKernelTest/EmptyKernelTest.cpp
PJAvinash: This is just variable initialization doesn't affect correctness of timing.

tools/EmptyKernelTest/EmptyKernelTest.cpp
PJAvinash: placed atomic_signal_fence(memory_order_seq_cst); wherever necessary

PR #1911, opened 2025-09-10
Print out the hipRuntimeVersion message from WARN to always show up

amd-jiali:
## Details
___Do not mention proprietary info or link to internal work items in this PR.___

**Work item:** _"Internal", or link to GitHub issue (if applicable)._

**What were the changes?**  
_Change the format of this warning message from WARN only to always printing out.

**Why were the changes made?**  
_Display hipruntime version directly, instead of only when NCCL_DEBUG=WARN.

**How was the outcome achieved?**  
_Technical details behind the work. Explain any publicly-available hardware peculiarities._

**Additional Documentation:**  
_What else should the reviewer know?_

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #2002, opened 2025-10-24
Added ERROR message class to handle fatal error messages.

thananon:
New ERROR message class will print the message in all debug level, including none.

Change some of the fatal error message to be in ERROR instead of WARN.

Added new error handler function to print out more meaningful error message in the future.

## Details

**What were the changes?**  
New debug message class ERROR to always print out the message. This should be used to communicate something really important.

Some of our error messages are not straightforward and caused user to miss it or misunderstand the error that is caused by underlying libraries to be RCCL's fault.

Removed NCCL_LOG_NONE and change it to NCCL_LOG_ERROR as the top message priority.

## Approval Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ### Changed
 
-* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages by default. User can suppress fatal error messages by setting NCCL_DEBUG=NONE.
+* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages by default. Fatal error messages can be suppressed by setting `NCCL_DEBUG=NONE`.
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.2.0
 
+### Changed
+
+* RCCL error message has been made more verbose in several cases. RCCL will now start printing out fatal error message at all debug level.
+
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
 ### Resolved Issues
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ### Changed
 
-* RCCL error message has been made more verbose in several cases. RCCL will now start printing out fatal error message at all debug level.
+* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages at all debug levels.
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ### Changed
 
-* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages at all debug levels.
+* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages by default. User can suppress fatal error messages by setting NCCL_DEBUG=NONE.
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
--- a/src/debug.cc
+++ b/src/debug.cc
@@ -47,7 +47,7 @@ static void ncclDebugInit() {
     ncclDebugFile = stdout;
   }
   if (nccl_debug == NULL) {
-    tempNcclDebugLevel = NCCL_LOG_NONE;
+    tempNcclDebugLevel = NCCL_LOG_ERROR;
   } else if (strcasecmp(nccl_debug, "VERSION") == 0) {
     tempNcclDebugLevel = NCCL_LOG_VERSION;
   } else if (strcasecmp(nccl_debug, "WARN") == 0) {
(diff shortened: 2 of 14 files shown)


corey-derochie-amd (dismissed):
Your PR contains a modification to mscclpp submodule, please revert that part. Otherwise, lgtm.

src/include/nccl_common.h
```diff
@@ -11,7 +11,7 @@
 #include "nccl.h"
 
 typedef enum {
-  NCCL_LOG_NONE = 0,
```
alex-breslow-amd, line 14: Should we keep NCCL_LOG_NONE and just make it -1 or bump the other values?

corey-derochie-amd: Looks like ERROR is intended to only print when something fatal happened, and if that occurs it should ALWAYS print. Since NONE wasn't being used otherwise, I think this is fine imo

nileshnegi:
~can you move the HSA_NO_SCRATCH_RECLAIM one to this category? https://github.com/ROCm/rccl/pull/1911~
scratch that... already done... thanks!

ext-src/mscclpp
nileshnegi, line 1: what is happening here?

src/include/debug.h
```diff
@@ -40,4 +41,22 @@ void ncclSetThreadName(pthread_t thread, const char *fmt, ...);
...
+    ERROR("Cuda failure '%s'", cudaGetErrorString(err));
+
+    // Special error message here:
+    switch (err) {
+    case cudaErrorStreamCaptureInvalidated:
+	    ERROR("Application is trying to use an invalidated stream to launch RCCL kernel. "
+	          "This operation is invalid. RCCL is exiting.");
+	    break;
+    default:
+	    break;
```
alex-breslow-amd, line 57: I'd consider indenting ``case`` and ```default``` since they are part of the switch

ext-src/mscclpp
alex-breslow-amd: Is this unexpected with mscclpp updates?  I guess it is strange that it would get triggered on this PR in particular.

nileshnegi: our submodule is tied to a particular commit, so this should not be happening.

nileshnegi: Corey already brought it up earlier... closing this thread.

src/debug.cc
```diff
@@ -372,6 +372,8 @@ void ncclDebugLog(ncclDebugLogLevel level, unsigned long flags, const char *file
     auto delta = std::chrono::steady_clock::now() - ncclEpoch;
     double timestamp = std::chrono::duration_cast<std::chrono::duration<double>>(delta).count()*1000;
     len += snprintf(buffer+len, sizeof(buffer)-len, "[%d] %f %s:%d NCCL TRACE ", cudaDev, timestamp, filefunc, line);
+  } else if (level == NCCL_LOG_ERROR) {
+    len += snprintf(buffer+len, sizeof(buffer)-len, "[%d] [FATAL ERROR]: ", cudaDev);
```
alex-breslow-amd, line 378: This snprintf makes me sad, but it matches the existing code.

alex-breslow-amd (approved):
LGTM

nileshnegi (approved):

src/include/debug.h
thananon: I just keep the style consistent with other files.

src/include/debug.h
```diff
@@ -40,4 +41,22 @@ void ncclSetThreadName(pthread_t thread, const char *fmt, ...);
 
 void ncclResetDebugInit();
 
+// RCCL custom error message handling.
+static inline ncclResult_t rcclCudaErrorHandler(cudaError_t err) {
+
+    // Print the cuda error
+    ERROR("Cuda failure '%s'", cudaGetErrorString(err));
```
thananon: Are we caring about Cuda compilation or we can just HIP it here?

thananon:
> Your PR contains a modification to mscclpp submodule

Nahhhhh

src/include/debug.h
thananon: I think HIP is more meaningful for our debugging. Changing to HIP.

corey-derochie-amd (approved):

gilbertlee-amd (approved):

gilbertlee-amd (changes requested):
This is a potentially big change, that needs to be communicated within CHANGELOG.

nileshnegi:
@thananon i think occurences like this need to change as well: https://github.com/ROCm/rccl/blob/develop/ext-tuner/example/nccl/common.h#L10

thananon:
- Addressed @nileshnegi comment.
- I am waiting for #2008 to go in before I add changelog to this PR.

thananon:
Changelog added. Please re-review @gilbertlee-amd

CHANGELOG.md
gilbertlee-amd, line 1: This feature needs more clarity.  RCCL (and other libraries) generally don't have access to stderr/stdout, so it's important to make that distinction know as to not surprise end-users.  
Also, the CHANGELOG should also provide information about how to "opt-out" of these new change if they really don't want error messages printed off to stderr/stdout.

gilbertlee-amd (changes requested):

CHANGELOG.md
thananon: I cant imagine a case where the user does not want to see a fatal error. This class of error is only going to be used in fatal case. We can see this example in MPI/libfabric/UCX which always prints fatal error message.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.2.0
 
+### Changed
+
+* RCCL error message has been made more verbose in several cases. RCCL will now start printing out fatal error message at all debug level.
```
```suggestion
* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages at all debug levels.
```

CHANGELOG.md
thananon: Added opt-out.

gilbertlee-amd (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.2.0
 
+### Changed
+
+* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages by default. User can suppress fatal error messages by setting NCCL_DEBUG=NONE.
```
```suggestion
* RCCL error messages have been made more verbose in several cases. RCCL now prints out fatal error messages by default. Fatal error messages can be suppressed by setting `NCCL_DEBUG=NONE`.
```

thananon:
The last modification only for CHANGELOG.md. CI had passed earlier. I will wait until EOD today before merging this.

PR #2011, opened 2025-10-28
P2p batching hang-fix

isaki001:
## Details

**What were the changes?**  
- reverted use of bit-reversal for channel-to-part mapping
- prevent p2p-batching when recv/send bytes in a batch are different (this applies for a2av)
- p2p-batching is supported only up to 32 nodes

- RCCL_P2P_BATCH_ENABLE=1 provides perf. boost for alltoall and should be enabled by default
- linear mapping use was reverted in favor of bit-reversal to address regressions on e2e workloads, originally linear mapping was introduced to improve peak-bw, however this improvement is less impactful than the e2e workload regression and needs to be reverted
- the requirement for recv/send bytes to be the same addresses a hang on a2av. The utilization of the RCCL_P2P_BATCH_THRESHOLD could lead to cases where the send bytes would require batching but the recv bytes would not, and this inconsistency would lead to the hang. This hang could also be addressed by not using a RCCL_P2P_BATCH_THRESHOLD, or forcing/disabling batching on all message sizes. 
- p2p-batching is disabled beyond 32 nodes to address hang for a2a as a temporary solution

- bit-reversal enablement only required changes in ncclP2pChannelForPart and ncclP2pChannelToPart
- p2p-batching adjustments only required changes in enqueue.cc

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
 ### Changed
-* `RCCL_P2P_BATCH_ENABLE` is now set to 1, enabling p2p batching by default
+* Enabling P2P batching with `RCCL_P2P_BATCH_ENABLE=1` is only applicable up to 32 nodes.
 
 ### Resolved Issues
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
+### Changed
+* `RCCL_P2P_BATCH_ENABLE` is now set to 1, enabling p2p batching by default
+
 ### Resolved Issues
 
 * Fixed crash when using the librccl-profiler plugin with the all-to-all collective after the 2.27 update.
@@ -17,7 +20,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 * Added `RCCL_P2P_BATCH_THRESHOLD` to set the message size limit for batching P2P operations. This mainly affects small message performance for alltoall at a large scale but also applies to alltoallv.
 * Added `RCCL_P2P_BATCH_ENABLE` to enable batching P2P operations to receive performance gains for smaller messages up to 4MB for alltoall when the workload requires it. This is to avoid performance dips for larger messages.
 * added `RCCL_CHANNEL_TUNING_ENABLE` to enable channel tuning that overrides RCCL's internal adjustments based on threadThreshold.
-* `RCCL_P2P_BATCH_ENABLE` is now set to 1, enabling p2p batching by default
+
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 * Added `RCCL_FORCE_ENABLE_DMABUF` as a debugging feature if the user wants to explicitly enable DMABUF and forego system/kernel checks.
 * Added `RCCL_P2P_BATCH_THRESHOLD` to set the message size limit for batching P2P operations. This mainly affects small message performance for alltoall at a large scale but also applies to alltoallv.
 * Added `RCCL_P2P_BATCH_ENABLE` to enable batching P2P operations to receive performance gains for smaller messages up to 4MB for alltoall when the workload requires it. This is to avoid performance dips for larger messages.
-* added `RCCL_CHANNEL_TUNING_ENABLE` to enable channel tuning that overrides RCCL's internal adjustments based on threadThreshold.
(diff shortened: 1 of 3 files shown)


src/include/device.h
```diff
@@ -317,10 +317,24 @@ inline __host__ uint8_t ncclP2pChannelBaseForRound(struct ncclComm* comm, int p2
 // ncclP2pChannelToPart and ncclP2pChannelForPart are inverses. The device code
 // uses ncclP2pChannelToPart to determine which part "this" channel is responsible for.
 inline __host__ int ncclP2pChannelForPart(int nP2pChannels, int base, int part, int nParts, int nNodes) {
+  if (nNodes > 2) {
+    // Only works because nP2pChannels is pow2
+    int nChannelsLog2 = countOneBits(nP2pChannels-1);
```
mustafabar, line 322: can we add opt-in option for linear mapping (as discussed, it helped Pollara NICs)

mustafabar (dismissed):

src/include/device.h
```diff
@@ -317,10 +317,24 @@ inline __host__ uint8_t ncclP2pChannelBaseForRound(struct ncclComm* comm, int p2
...
 }
 inline __device__ int ncclP2pChannelToPart(int nP2pChannels, int base, int channel, int nParts, int nNodes) {
-    return (channel - base * nParts) & (nP2pChannels-1);
+  if (nNodes > 2) {
+    // Only works because nP2pChannels is pow2
+    int nChannelsLog2 = countOneBits(nP2pChannels-1);
+    int delta = (channel-base) & (nP2pChannels-1);
+    return reverseBits(delta, nChannelsLog2);
+  } else {
+    return (channel - base * nParts) & (nParts-1);
```
nileshnegi: @mustafabar was there a reason to use `return (channel - base * nParts) & (nP2pChannels-1);` with your previous linear-mapping-only changes instead of `return (channel - base * nParts) & (nParts-1);`?

mustafabar: The reason is that one function should be exactly the reverse of the other. The previous logic did not make sense to me and I didn't know how it worked. It used `(nParts-1)` on one direction and nP2pChannels in the other direction.

nileshnegi: @isaki001 pls correct this line.

src/include/device.h
isaki001: As discussed, applying this requires propagating an env. variable value to the kernel and adding an extra if-statement in a ncclP2pChannelToPart, which could have negative impact. Initial runs didn't show a regression, but we need more extended tests to safely add this.

nileshnegi (approved):

nileshnegi:
@isaki001 please add a changelog entry, as batched P2P is now enabled by default

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 * Added `RCCL_FORCE_ENABLE_DMABUF` as a debugging feature if the user wants to explicitly enable DMABUF and forego system/kernel checks.
 * Added `RCCL_P2P_BATCH_THRESHOLD` to set the message size limit for batching P2P operations. This mainly affects small message performance for alltoall at a large scale but also applies to alltoallv.
 * Added `RCCL_P2P_BATCH_ENABLE` to enable batching P2P operations to receive performance gains for smaller messages up to 4MB for alltoall when the workload requires it. This is to avoid performance dips for larger messages.
-* Added `RCCL_CHANNEL_TUNING_ENABLE` to enable channel tuning that overrides RCCL's internal adjustments based on `threadThreshold`.
+* added `RCCL_CHANNEL_TUNING_ENABLE` to enable channel tuning that overrides RCCL's internal adjustments based on threadThreshold.
```
nileshnegi: any reason to change the earlier statement?
```suggestion
* Added `RCCL_CHANNEL_TUNING_ENABLE` to enable channel tuning that overrides RCCL's internal adjustments based on `threadThreshold`.
```

CHANGELOG.md
```diff
@@ -16,7 +16,8 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
...
+* `RCCL_P2P_BATCH_ENABLE` is now set to 1, enabling p2p batching by default
```
nileshnegi: this should be moved to the `Unreleased` section

CHANGELOG.md
isaki001: typo, reverted change.

AbandiGa (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Full documentation for RCCL is available at [https://rccl.readthedocs.io](https:
 
 ## Unreleased - RCCL 2.27.7 for ROCm 7.1.1
 
+### Changed
+* `RCCL_P2P_BATCH_ENABLE` is now set to 1, enabling p2p batching by default
```
nileshnegi: needs to be edited (or) removed.
also, maybe we need a statement here about supporting batched P2P only up to 32 nodes.

nileshnegi:
PR has passed CI checks in the past. Latest changes are doc/changelog/default parameter changes. Safe to merge.

PR #2006, opened 2025-10-27
[GEN/BUILD] Refactor generate.py and reduce build time for older archs

BertanDogancay:
## Details
___Do not mention proprietary info or link to internal work items in this PR.___

**Work item:** _"Internal", or link to GitHub issue (if applicable)._
LWPCLPAT-599

**What were the changes?**  
_One sentence describing the work done._
Cleaning and refactoring generate.py and reduce build time on older archs.
* Use DataClass for fn.
* New function `get_arch_guard()` for arch requirements and long term maintenance.
* Allow nullptr entries in device table.
* Get rid of duplicate code.

**Why were the changes made?**  
_Explain the motivation behind the work. Provide any publicly-available historical context._
Continue PR #1990 work and address the comments by @alex-breslow-amd

**How was the outcome achieved?**  
_Technical details behind the work. Explain any publicly-available hardware peculiarities._

**Additional Documentation:**  
_What else should the reviewer know?_
Will close PR #1990 once this is merged.

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #2021, opened 2025-10-30
Revert "[GEN/BUILD] Refactor generate.py and reduce build time for older archs (#2006)"

nileshnegi:
This reverts commit bed7cdf86341a5a9fb36953eebd00a25b83a1ea6.

## Details

**Work item:**
Internal

**What were the changes?**  
Reverts https://github.com/ROCm/rccl/commit/bed7cdf86341a5a9fb36953eebd00a25b83a1ea6

With the refactored generator in #2006, the `host_table.cpp` entries relied on GPU arch. guards (ifdefs) for deciding between valid/invalid entries. However, since these ifdefs are not used during host compilation, the guarded entries were always invalid in the host table. As a result, the lookup would fail to find an entry, and RCCL would segfault, even for valid device functions.

**Additional Documentation:**  
A proper fix will be implemented as soon as possible. We are reverting for now to unblock critical deliveries.

## Approval Checklist

### DIFF
--- a/src/device/generate.py
+++ b/src/device/generate.py
@@ -2,19 +2,22 @@
 import os
 import sys
 import subprocess
-from dataclasses import dataclass
 
 # Order of colls, redops, tys, protos, algos must match src/include/device.h
-all_colls     = ["Broadcast", "Reduce", "AllGather", "ReduceScatter", "AllReduce", "SendRecv", "", "", "AllToAllPivot"]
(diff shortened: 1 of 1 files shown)


alex-breslow-amd (approved):
LGTM, thanks for the fix.

mustafabar (approved):
This is a critical issue. Let's merge asap

PR #2023, opened 2025-11-01
Fail the job if flag HIP_HOST_UNCACHED_MEMORY is not set on MI350x

dmwu:
as $title. Place the check after initTransportsRank as the GPU arch info in comm->topo->nodes info is populated after that.

## Details

**What were the changes?**  

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -162,7 +162,7 @@ ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
     return ncclSuccess;
   #else
     if( IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx950") ){
-      WARN("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
+      ERROR("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
       return ncclSystemError;
     }
     else {
--- a/src/init.cc
+++ b/src/init.cc
@@ -155,6 +155,22 @@ ncclResult_t checkHsaEnvSetting() {
   }
   return ncclSuccess;
 }
+
+// Fail the job if build flag HIP_HOST_UNCACHED_MEMORY is not set on mi350x
+ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
+  #if defined(HIP_HOST_UNCACHED_MEMORY)
+    return ncclSuccess;
+  #else
+    if( IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx950") ){
+      WARN("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
+      return ncclSystemError;
+    }
+    else {
+      return ncclSuccess;
+    }
+  #endif   
+}
+
 static void initOnceFunc() {
   NCCLCHECKGOTO(checkHsaEnvSetting(), initResult, exit);
   initEnv();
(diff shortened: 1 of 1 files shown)


src/init.cc
```diff
@@ -155,6 +155,22 @@ ncclResult_t checkHsaEnvSetting() {
...
   return ncclSuccess;
 }
+
+// Fail the job if build flag HIP_HOST_UNCACHED_MEMORY is not set on mi350x
+ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
+  #if defined(HIP_HOST_UNCACHED_MEMORY)
+    return ncclSuccess;
+  #else
+    if( IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx950") ){
+      WARN("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
```
nileshnegi: @dmwu can you make this ERROR instead of WARN?

we recently added a new ERROR class, which is enabled by default, instead of relying on users setting `NCCL_DEBUG=WARN`

```suggestion
      ERROR("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
```

alex-breslow-amd (approved):
Looks good to me.  Thanks for the fix @dmwu!

@dmwu, does Meta have any interest in contributing back the buck2 targets files and build infra to RCCL?  I think this may reduce future issues for you, as AMD may be able to catch issues like this sooner.

src/init.cc
alex-breslow-amd: I concur.  This is cleaner.

dmwu:
> Looks good to me. Thanks for the fix @dmwu!
> 
> @dmwu, does Meta have any interest in contributing back the buck2 targets files and build infra to RCCL? I think this may reduce future issues for you, as AMD may be able to catch issues like this sooner.

@alex-breslow-amd This is a good idea. Let me discuss internally on this and followup

nileshnegi (approved):

src/init.cc
```diff
@@ -155,6 +155,22 @@ ncclResult_t checkHsaEnvSetting() {
...
   return ncclSuccess;
 }
+
+// Fail the job if build flag HIP_HOST_UNCACHED_MEMORY is not set on mi350x
+ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
+  #if defined(HIP_HOST_UNCACHED_MEMORY)
+    return ncclSuccess;
+  #else
+    if( IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx950") ){
+      ERROR("Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x");
```
dmwu, line 165: @nileshnegi  @alex-breslow-amd  Just realized that NCCL doesn't provide a ERROR log function. We return ncclSystemError anyway, so the job failed with explicit error msg:
"​[2025-10-31 17:12:40] devgpu048:379761:380299 [2] buck-out/v2/gen/fbsource/fb777038863bb0cd/third-party/rccl/__hipify__/hipify/develop/src/init.cc.hip:156 NCCL WARN Build flag HIP_HOST_UNCACHED_MEMORY must be set to avoid memory corruption on mi350x"

nileshnegi: true @dmwu, but a common theme is that if users do not set `NCCL_DEBUG=WARN` or `NCCL_DEBUG=INFO`, the error message is not visible, and a generic `unhandled system error (run with NCCL_DEBUG=INFO for details)` gets printed.

and if folks do not heed this generic warning to re-run with NCCL_DEBUG=INFO, things get flagged as RCCL failure, which is why we had to add this new ERROR class to print the exact error message by default.

dmwu:
@nileshnegi  The Rock CI multi-node tests keeps failing. Mind taking a look? thanks!

nileshnegi:
> @nileshnegi The Rock CI multi-node tests keeps failing. Mind taking a look? thanks!

there's a credentials issue with TheRock multi-node CI at the moment.
while folks resolve that, we can merge this PR.

dmwu:
@nileshnegi  I don't see the merge PR option available on this page. Is it because the failing CI?

PR #2029, opened 2025-11-03
proxy: handle progressOps return code properly.

thananon:
## Details

**Work item:** SWDEV-553150

**What were the changes?**  
Add proper error handing from `progressOps`. The patch is taken directly from NCCL.

This bug is introduced with NCCL 2.26 sync which later been fixed in NCCL 2.28. We still have not finished our 2.28 sync but we do need to fix this bug now. 

Port the fix from latest NCCL.

## Approval Checklist


PJAvinash (approved):

thomas-huber (approved):

alex-breslow-amd (approved):
Thanks

PR #2032, opened 2025-11-04
Reduce LL threshold for alltoall

mustafabar:
## Details

**What were the changes?**  
_Change default for P2P_LL_THRESHOLD._

_Improve Alltoall 1MB message performance. On MI350, there's a solid uplift from ~45 GB/s to ~55 GB/s for 1MB message size. On MI300x, it is not as significant_

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/enqueue.cc
+++ b/src/enqueue.cc
@@ -918,7 +918,7 @@ static ncclResult_t scheduleCollTasksToPlan(
   return ncclSuccess;
 }
 
-NCCL_PARAM(P2pLLThreshold, "P2P_LL_THRESHOLD", 16384);
+NCCL_PARAM(P2pLLThreshold, "P2P_LL_THRESHOLD", 8192);
 RCCL_PARAM(P2pNetThreshold, "P2P_NET_THRESHOLD", 131072);
 NCCL_PARAM(ChunkSize, "CHUNK_SIZE", 0);
 


alex-breslow-amd (approved):
LGTM

src/enqueue.cc
```diff
@@ -918,7 +918,7 @@ static ncclResult_t scheduleCollTasksToPlan(
   return ncclSuccess;
 }
 
-NCCL_PARAM(P2pLLThreshold, "P2P_LL_THRESHOLD", 16384);
+NCCL_PARAM(P2pLLThreshold, "P2P_LL_THRESHOLD", 8192);
```
nusislam, line 931: My understanding was that alltoall does not use LL protocol by default. Has this changed?

mustafabar: I don't think this has changed. For multi-node, we need to set `NCCL_ALLOC_P2P_NET_LL_BUFFERS`. When sending/receiving within the same node, [this condition](https://github.com/ROCm/rccl/blob/d6a53d20228fc7eeb567df6c96a560be4af0ac8f/src/enqueue.cc#L987 ) seems to check out since buff is not outside of the node, so it is not null. So it is a mixed bag when we have multi-node send/recv, which is quite weird

AbandiGa (approved):

PedramAlizadeh (approved):

PR #2036, opened 2025-11-06
Added new unit tests for AllReduce with Bias API

atulkulk:
## Details

**Work item:** LWPCLPAT-480

**What were the changes?**  
Added new unit tests for AllReduce with Bias API

To improve code quality.

New tests were added.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/test/AllReduceTests.cpp
+++ b/test/AllReduceTests.cpp
@@ -249,4 +249,448 @@ namespace RcclUnitTesting
     }
     callCollectiveForked(nranks, ncclCollAllReduce, sendBuff, recvBuff, expected, use_managed_mem);
   }
+
+#ifdef RCCL_ALLREDUCE_WITH_BIAS
+  // Note: All bias tests require:
+  // nRanks >= 2 (bias NOT supported for single rank)
+
+  // Named constants for bias test configuration
+  namespace BiasTestConstants
+  {
+  // Element counts for different operations
+  constexpr std::initializer_list<int> STANDARD_ELEM_COUNTS    = {2048, 384}; // For Sum/Max/Min
+  constexpr std::initializer_list<int> PROD_ELEM_COUNTS_MEDIUM = {32}; // For Int32/Uint32 Prod
+  constexpr std::initializer_list<int> PROD_ELEM_COUNTS_LARGE  = {64}; // For Int8/Uint8/Int64/Uint64/Float Prod
+
+  // Bias and input pattern constants
+  constexpr int BIAS_CONSTANT_ONE = 1; // Use constant bias value of 1 (prevents overflow)
+  constexpr int BIAS_INCREMENTAL_PATTERN
+      = -1; // Use incremental pattern: bias[i] = i (more thorough testing)
+  constexpr int INPUT_RANK_BASED_PATTERN
+      = -1; // Use rank-based pattern: input[rank][i] = (rank+i)%256
+  constexpr int INPUT_CONSTANT_ONE = 1; // Use constant input value of 1 (prevents overflow)
+  } // namespace BiasTestConstants
+
+  /*
+   * @brief Helper function for running bias tests with specific datatype and redOp
+   * @param dataType Data type
+   * @param redOp Reduction operation
+   * @param numElements Number of elements
+   * @param biasConstVal Bias constant value, -1 for incremental bias
+   * @param inputConstVal Input constant value, -1 for rank-based input
+   */
+  void RunBiasTest(ncclDataType_t   dataType,
+                   ncclRedOp_t      redOp,
+                   std::vector<int> numElements,
+                   int              biasConstVal  = BiasTestConstants::BIAS_INCREMENTAL_PATTERN,
+                   int              inputConstVal = BiasTestConstants::INPUT_RANK_BASED_PATTERN)
+  {
+      // Create TestBed first (doesn't create child processes yet)
+      TestBed testBed;
+
+      // Check if architecture is gfx94 (covers gfx942) or gfx95 (covers gfx950)
+      if (!testBed.ev.isGfx94 && !testBed.ev.isGfx95)
+      {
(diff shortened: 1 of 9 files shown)


test/AllReduceTests.cpp
```diff
@@ -249,4 +252,532 @@ namespace RcclUnitTesting
...
+   */
+  void RunBiasTest(TestBed&         testBed,
+                   ncclDataType_t   dataType,
+                   ncclRedOp_t      redOp,
+                   std::vector<int> numElements,
+                   int              biasConstVal  = BiasTestConstants::BIAS_INCREMENTAL_PATTERN,
+                   int              inputConstVal = BiasTestConstants::INPUT_RANK_BASED_PATTERN)
+  {
+      // Check if NCCL_ALGO is set to TREE (required for bias operations)
+      if (!IsTreeAlgorithmEnabled())
```
nusislam: Why is tree algo a requirement?

atulkulk: Yes, that was incorrect, I just removed that check. I tested against RING and it worked as expected. I was tracking `work->acc` and thought it worked for only TREE. I checked generate.py and it did create the kernel for other algos.

nusislam:
LGTM. Need to merge with develop. Also, check what the CI failures are about.

wenkaidu (approved):

test/AllReduceTests.cpp
```diff
@@ -249,4 +249,453 @@ namespace RcclUnitTesting
...
+                                                             inPlace,
+                                                             useManagedMem,
+                                                             useHipGraph);
+                  INFO("  %s (with bias, count=%d)\n", name.c_str(), numElem);
+              }
+
+              options.biasNumElements = numElem;
+
+              for(int rank = 0; rank < totalRanks; ++rank)
+              {
```
Kapil-Shyam-Pawar: Can you please check if this for loop is needed
`InitComms` is already getting passed with `totalRanks` on line 319
if you pass `rank=-1` to `SetCollectiveArgs`, it would set it for all the ranks as per the function logic
If this works for your use case, this for loop can be removed

test/AllReduceTests.cpp
```diff
@@ -249,4 +249,453 @@ namespace RcclUnitTesting
     }
     callCollectiveForked(nranks, ncclCollAllReduce, sendBuff, recvBuff, expected, use_managed_mem);
   }
+
+#ifdef RCCL_ALLREDUCE_WITH_BIAS
+  // Note: All bias tests require:
+  //   1. nRanks >= 2 (bias NOT supported for single rank)
+  //   2. Only NCCL_ALGO=TREE is supported
```
Kapil-Shyam-Pawar: Is this comment needed now since you removed the condition for tree?

test/common/CollectiveArgs.cpp
```diff
@@ -102,6 +107,23 @@ namespace RcclUnitTesting
...
+      this->numBiasBytesAllocated = this->numBiasElements * DataTypeToBytes(this->dataType);
+      CHECK_CALL(this->biasGpu.AllocateGpuMem(this->numBiasBytesAllocated, useManagedMem, userRegistered));
+      CHECK_CALL(this->biasCpu.AllocateCpuMem(this->numBiasBytesAllocated));
+      this->biasRegHandle = nullptr;
+    }
+    else
+    {
+      this->numBiasElements = 0;
+      this->numBiasBytesAllocated = 0;
+      this->biasRegHandle = nullptr;
```
alex-breslow-amd: This code is the same as the code above.  Seems like a good target for a function.  Incidentally, do you even need to call this if the initialization is handled by default above?

Perhaps we can make this automatic so that the user doesn't have to explicit call this.

test/common/CollectiveArgs.cpp
```diff
@@ -155,6 +177,15 @@ namespace RcclUnitTesting
...
     }
+
+    // Deallocate bias buffers if they were allocated
+    if (this->options.useBias && this->numBiasBytesAllocated > 0)
+    {
+      this->biasGpu.FreeGpuMem(this->userRegistered);
+      this->biasCpu.FreeCpuMem();
+      this->biasRegHandle = nullptr;
+    }
+
```
alex-breslow-amd, line 177: It would be nice if we didn't have to explicitly call this.  Ideally, this would just happen if the object falls out of scope.  Not sure if this is easy here.

test/common/CollectiveArgs.hpp
```diff
@@ -82,6 +82,15 @@ namespace RcclUnitTesting
     ScalarTransport scalarTransport;        // Used for custom reduction operators
     int             scalarMode = -1;        // -1 if scalar not used
 
+    // Bias support for fused AllReduce+Bias operations
+    bool            useBias = false;        // Enable bias addition
+    void*           biasPtr = nullptr;      // Pointer to bias buffer (GPU memory)
+    size_t          biasNumElements = 0;    // Number of elements in bias buffer
```
alex-breslow-amd, line 88: Since you already put this in a valid state here, why are you rezeroing it in other places?  Let me know if I'm missing something.

test/common/TestBed.cpp
```diff
@@ -636,12 +636,16 @@ namespace RcclUnitTesting
                                std::vector<bool>           const& inPlaceList,
                                std::vector<bool>           const& managedMemList,
                                std::vector<bool>           const& useHipGraphList,
-                               bool                        const& enableSweep)
```
alex-breslow-amd, line 639: I think this is fine, but for deep callstacks, it can be preferable to pass a struct into functions to reduce the number of changes to function signatures when new fields are added.  In this case, I don't think it matters.

alex-breslow-amd (approved):
LGTM

test/common/CollectiveArgs.cpp
atulkulk: There will be some effort required to achieve this and it will apply to other objects in this method too. For now, deferring on this.

PR #2038, opened 2025-11-07
BUCK and bazel files

ahmd-k:
As requested in https://github.com/ROCm/rccl/pull/2023#pullrequestreview-3412193960


nileshnegi:
thanks!
could you move this to a separate dir like `buck/` instead of the repo root?

ahmd-k:
> thanks! could you move this to a separate dir like `buck/` instead of the repo root?

Done in new commit

nileshnegi:
> Done in new commit

thanks!
sorry, another request, could you pull and rebase on top of latest develop. we hit a snafu in the develop branch and it had to be force updated.

src/include/net.h
dmwu, line 1: @ahmd-k  Can we exclude these irrelevant changes on the src code

PR #2037, opened 2025-11-06
Using hip_bf16.h instead of hip_bfloat16.h for the __bf16 intrinsic

PedramAlizadeh:
## Details
**Work item:** _"Internal", or link to GitHub issue (if applicable)._
LWPCLPAT-588
**What were the changes?**  
Switching to the hip_bf16.h implementation with the new __hip_bfloat16 datatype

**Why were the changes made?**  
To observe the performance gain for bf16 datatype through __bf16 intrinsic on MI350

**How was the outcome achieved?**  
By switching to the hip_bf16.h implementation with the new __hip_bfloat16 datatype

**Additional Documentation:**  

## Approval Checklist
___Do not approve until these items are satisfied.___
- [ ] Verify the CHANGELOG has been updated, if
  - there are any NCCL API version changes,
  - any changes impact library users, and/or
  - any changes impact any other ROCm library.

PR #2047, opened 2025-11-11
Disable Bfloatf16 pipelining for reduction collectives for gfx950

AbandiGa:
**What were the changes?**  
Disable Bfloatf16 pipelining for reduction collectives for gfx950 in generate.py

Based on current testing and the performance uplift observed from the new __bf16 intrinsic https://github.com/ROCm/rccl/pull/2037, we are no longer observing significant uplift from the pipelining in gfx950. 

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/device/generate.py
+++ b/src/device/generate.py
@@ -170,9 +170,9 @@ class Fn:
   def __iter__(self):
     return iter((self.coll, self.algo, self.proto, self.redop, self.ty, self.acc, self.pipeline, self.unroll))
 
-def calc_unroll_for_local_arch():
+def calc_unroll_and_pipeline_for_local_arch():
   if not is_local_arch_only:
-    return all_unrolls
+    return (all_unrolls, all_pipelines)
 
   rocminfo_path = os.environ.get('ROCM_PATH') + "/bin/rocminfo"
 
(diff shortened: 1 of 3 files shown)


src/device/generate.py
```diff
@@ -214,6 +214,41 @@ def calc_unroll_for_local_arch():
 # except for gfx950
 local_unroll = calc_unroll_for_local_arch()
 
+def calc_pipeline_for_local_arch():
+  # Default: pipelining enabled
+  default_pipeline = ["0", "1"]
```
nileshnegi: since this func runs only for local gfx950 builds, you can just override `all_pipelines`?

src/device/generate.py
```diff
@@ -214,6 +214,41 @@ def calc_unroll_for_local_arch():
...
+
+  # Disable pipelining for MI350 (gfx950)
+  if "gfx950" in gfx_targets:
+    return ["0"]  # Disable pipelining
+  
+  return default_pipeline
+
+# if building for gfx950, we need to disable pipelining
+local_pipeline = calc_pipeline_for_local_arch()
+
```
nileshnegi, line 217: also, maybe the pipeline and unroll stuff can be clubbed together to return a tuple, since the local_arch detection logic is common between both funcs.

src/device/generate.py
AbandiGa: you mean use all_pipelines here instead of default_pipeline ?

alex-breslow-amd: I agree with this.

alex-breslow-amd (approved):
You could consider typing calc_unroll_and_pipeline_for_arch so that it returns a dataclass, but a tuple of two lists is fine.  Accepting.

src/device/generate.py
```diff
@@ -226,6 +226,7 @@ def func_validate(coll, algo, proto, redop, ty, acc,  pipeline, unroll):
       ty not in tys_of_coll[coll] or
       acc not in acc_of_coll[coll] or
       pipeline not in pipelines_of_coll[coll] or (pipeline in ["1"] and ty not in pipelined_types) or
+      pipeline not in local_pipeline or
```
alex-breslow-amd, line 230: It feels to me like this pipeline logic may be able to be cleaned up.   That's a lot of or's and not's but not a blocker for me.

AbandiGa: yes, it's a lot of or's. I can break it into smaller conditions                                                                                          
```
if (algo not in algos_of_coll[coll] or
    proto not in protos_of_coll[coll] or
    redop not in redops_of_coll[coll] or
    ty not in tys_of_coll[coll] or
    acc not in acc_of_coll[coll] or
    unroll not in local_unroll):
  return False

if pipeline not in local_pipeline:
  return False
if pipeline not in pipelines_of_coll[coll]:
  return False
if pipeline == "1" and ty not in pipelined_types:
  return False

```

nileshnegi:
Won't this PR error out during multi-node runs, as for local gfx950 builds, we do not generate kernels with bf16 pipelining support, but https://github.com/ROCm/rccl/blob/develop/src/rccl_wrap.cc#L247-L261 still comes into play?

So isn't there a mismatch in expected vs supported device function lookup, like:
`NCCL WARN Fatal error: ncclDevFuncId: 17367572 not found for coll: 4, algo: 1, proto: 2, devRedOp: 0, type: 9, acc: 0, pipeline: 1`

mustafabar (approved):

src/device/generate.py
```diff
@@ -197,22 +195,28 @@ def calc_unroll_for_local_arch():
   # We want to remove duplicates but cannot use a dictionary since same gfx name can have different cu counts
   # Use (gfx_name, cu_count) as key for dictionary and convert it to list here
   gfx_targets = list(gfx_targets.keys())
+  
+  has_gfx950 = any(gfx_name == "gfx950" for gfx_name, _ in gfx_targets)
```
nileshnegi: this won't work as intended.
here `gfx_targets` is used for querying GPU targets available on the build system. it is not intended for user-supplied list of GPU targets,

AbandiGa: yes, I understand this will only work if we are building in gfx950 system but even if it's not local arch build.

PR #2049, opened 2025-11-11
[IFC] Remove noinline attributes and default to indirect function calls

doru1004:
Remove noinline attributes and default to indirect function calls.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ option(ENABLE_MSCCLPP_CLIP                     "Enable MSCCL++ CLIP"            
 option(ENABLE_MSCCLPP_EXECUTOR                 "Enable MSCCL++ Executor"                       OFF)
 option(ENABLE_MSCCLPP_FORMAT_CHECKS            "Enable formatting checks in MSCCL++"           OFF)
 option(ENABLE_NPKIT                            "Enable NPKit"                                  OFF)
-option(ENABLE_IFC                              "Enable indirect function call"                 OFF)
+option(ENABLE_IFC                              "Enable indirect function call"                 ON)
 option(GENERATE_SYM_KERNELS                    "Generate symmetric memory kernels"             OFF)
 option(INSTALL_DEPENDENCIES                    "Force install dependencies"                    OFF)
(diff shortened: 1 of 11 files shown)


alex-breslow-amd:
@doru1004, is this ready for review or WIP?

doru1004:
> @doru1004, is this ready for review or WIP?

Yes it is now.

CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ option(ENABLE_MSCCLPP_CLIP                     "Enable MSCCL++ CLIP"
 option(ENABLE_MSCCLPP_EXECUTOR                 "Enable MSCCL++ Executor"                       OFF)
 option(ENABLE_MSCCLPP_FORMAT_CHECKS            "Enable formatting checks in MSCCL++"           OFF)
 option(ENABLE_NPKIT                            "Enable NPKit"                                  OFF)
-option(ENABLE_IFC                              "Enable indirect function call"                 OFF)
+option(ENABLE_IFC                              "Enable indirect function call"                 ON)
```
alex-breslow-amd, line 35: This is just a no-op now, right?  Maybe we can delete it.

doru1004: Well not quite, the code to perform direct function calls is still there I haven't removed that yet. Once we clean up any trace of direct calls we can remove this flag completely.

nileshnegi:
Should this guard be modified?
https://github.com/ROCm/rccl/blob/develop/src/init.cc#L2018

alex-breslow-amd:
> Should this guard be modified? https://github.com/ROCm/rccl/blob/develop/src/init.cc#L2018
Nilesh, that's a very good point.  This is indeed something we should look at.  Early results suggest that the current optimizations may be a regression for AllToAll on gfx950, so maybe this helps.

doru1004:
> Should this guard be modified? https://github.com/ROCm/rccl/blob/develop/src/init.cc#L2018

Since it doesn't seem to have any effect on gfx950 ( I tried it ) I would say that any change we need to do to that code it can be independent of the changes in this PR.

ammallya:
Imported to ROCm/rocm-systems

PR #2054, opened 2025-11-13
Added install.sh flag to suppress warnings.

thananon:
## Details

Compiler warnings are usually a good thing but due to the nature of our codebase, we cannot get rid of some warnings that we inherited from NCCL and the warning messages have been overwhelming our build log. This PR added an option to suppress most of the warnings. This option is optional, **it is for developers that has to rebuild frequently. The developer should still look at warning messages** before submitting new code.

Usage:
```
./install.sh -q
./install.sh --quiet-warnings
```
```
cmake {dir} -DQUIET_WARNINGS=ON
```

**What were the changes?**  
Add an option to suppress warnings.

Keep our developer sane.

Compiler warning suppression.

## Approval Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ option(PROFILE                                 "Enable profiling"               
 option(TIMETRACE                               "Enable time-trace during compilation"          OFF)
 option(TRACE                                   "Enable additional tracing"                     OFF)
 option(FAULT_INJECTION                         "Enable fault injection"                        ON)
+option(QUIET_WARNINGS                          "Supress compiler warnings"                     OFF)
 
 # Default GPU architectures to build
 #==================================================================================================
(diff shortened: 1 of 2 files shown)


mustafabar (approved):
Thanks for this !

nileshnegi (approved):

corey-derochie-amd:
There are separate PR open to actually fix existing warnings, some of which I believe have real implications. Do we want to proceed with suppressing warnings, or should we continue on to fixing them? I will find the PR# to include here. @nileshnegi @mustafabar @thananon

thananon:
I think we can both add an option to suppress the warnings and actually fixing the warnings.

mustafabar:
> There are separate PR open to actually fix existing warnings, some of which I believe have real implications. Do we want to proceed with suppressing warnings, or should we continue on to fixing them? I will find the PR# to include here. @nileshnegi @mustafabar @thananon

@corey-derochie-amd - it is an opt in option, so won't suppress by default.

corey-derochie-amd (approved):

PR #2063, opened 2025-11-18
Use one side stream per process

wenkaidu:
## Details

Internal

**What were the changes?**  
Use one side stream per process

RCCL memory allocation needs a side stream so as not to interfere with graph capture. However, previous implementation of side stream caching doesn't completely solve this issue: it has been reported that thousands of streams being created and destroyed by RCCL.

Use only one side stream per process.

**Additional Documentation:**  

## Approval Checklist


src/include/alloc.h
```diff
@@ -31,6 +31,37 @@ constexpr size_t ncclSizeOfT() { return sizeof(T); }
...
+
+static inline ncclResult_t ncclDestroySideStream() {
+  pthread_mutex_lock(&sideStreamLock);
+  sideStreamRefCount--;
+  if (sideStreamRefCount== 0 && sideStream)
+    CUDACHECK(cudaStreamDestroy(sideStream));
+  pthread_mutex_unlock(&sideStreamLock);
+  return ncclSuccess;
+};
+
```
alex-breslow-amd: I'm thinking that we should probably have a macro for pthread_calls that calls perror if the operation fails like ```PTHREAD_OK()```.  I think we recently added some new logging for fatals that you may be able to leverage.

alex-breslow-amd: However, your code also works.  I see below that you check whether the stream is created successfully.

src/include/alloc.h
```diff
@@ -31,6 +31,37 @@ constexpr size_t ncclSizeOfT() { return sizeof(T); }
 template<>
 constexpr size_t ncclSizeOfT<void>() { return 1; }
 
+extern cudaStream_t sideStream;
+extern pthread_mutex_t sideStreamLock;
+extern uint64_t sideStreamRefCount;
+extern pthread_once_t sideStream_initonce;
+
+static inline void init_sideStreamLock(void) {
+  if (pthread_mutex_init(&sideStreamLock, NULL) != 0) {
```
alex-breslow-amd: If you use PTHREAD_MUTEX_INITIALIZER as you do below, this shouldn't be necessary.
https://pubs.opengroup.org/onlinepubs/7908799/xsh/[pthread_mutex_init](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_init.html).html

src/include/alloc.h
alex-breslow-amd: The same is probably also true for PTHREAD_ONCE_INIT, but you'd have to check on that.  Don't quote me on that.

wenkaidu:
Need rework to handle multiple GPUs per process cases.

src/include/alloc.h
wenkaidu: fixed

src/include/alloc.h
wenkaidu: fixed

src/init.cc
```diff
@@ -94,7 +94,13 @@ NCCL_PARAM(CtaPolicy, "CTA_POLICY", NCCL_CONFIG_UNDEF_INT);
 NCCL_PARAM(NvlsChannels, "NVLS_NCHANNELS", NCCL_CONFIG_UNDEF_INT);
 
 struct allocationTracker allocTracker[MAX_ALLOC_TRACK_NGPU] = {};
-static ncclResult_t commReclaim(ncclComm_t comm);
+ncclResult_t commReclaim(ncclComm_t comm);
+
+// Global data for side stream
+std::map<int64_t, ncclSideStream> sideStream;
```
alex-breslow-amd: std::unordered_map may be a bit faster since it uses a hashmap instead of a red-black tree.  I don't think it matters in this case.

src/init.cc
```diff
@@ -94,7 +94,13 @@ NCCL_PARAM(CtaPolicy, "CTA_POLICY", NCCL_CONFIG_UNDEF_INT);
...
+pthread_mutex_t sideStreamLock = PTHREAD_MUTEX_INITIALIZER;
+uint64_t sideStreamRefCount = 0;
+pthread_once_t sideStream_initonce = PTHREAD_ONCE_INIT;
```
alex-breslow-amd: What prevents moving this to alloc.h?

src/include/alloc.h
```diff
@@ -31,6 +32,78 @@ constexpr size_t ncclSizeOfT() { return sizeof(T); }
...
 constexpr size_t ncclSizeOfT<void>() { return 1; }
 
+struct ncclSideStream {
+  cudaStream_t stream;
+  uint64_t refCount;
+};
+
+extern std::map<int64_t, ncclSideStream> sideStream;
+extern pthread_mutex_t sideStreamLock;
+extern ncclResult_t getBusId(int cudaDev, int64_t *busId);
```
alex-breslow-amd: Do these need to be extern?  I think this is the only file where at least some of these are used.

alex-breslow-amd (approved):
Left some comments.  I think this would make things a bit cleaner.

PedramAlizadeh (approved):

src/include/alloc.h
```diff
@@ -31,6 +32,78 @@ constexpr size_t ncclSizeOfT() { return sizeof(T); }
...
+
+static inline ncclResult_t ncclCreateSideStream(int cudaDev) {
```
alex-breslow-amd, line 44: Question here.  If we only have a single GPU per process, how many of these sidestreams are there?  Is it worth optimizing for that case?

src/init.cc
wenkaidu: fixed

src/init.cc
wenkaidu: fixed

src/include/alloc.h
wenkaidu: fixed, except getBusId has include ordering issue.

src/include/alloc.h
wenkaidu: there will be only one side stream with single GPU per process

alex-breslow-amd: I guess this is fine.  If we detect it being a regression, we can always optimize it further.

wenkaidu:
The CI failures were all caused by infrastructure issues.

PR #2067, opened 2025-11-19
Allow for advanced tuning of P2P channels/part mapping via RCCL_P2P_SHIFT_SIZE

gilbertlee-amd:
## Details
Implement ability to adjust how P2P channels are mapped to parts

SWDEV-567810

**What were the changes?**  
Addition of new complementary ncclP2pChannelForPart / ncclP2pChannelToPart functions that take in a shiftSize variable that adjusts how work is assigned

This allows for further investigation to see if balancing channels across XCCs can yield improved performance

Modifications to the two key functions listed above, then additions to ncclComm / ncclDevComm to carry this new information.  
A new RCCL environment variable to adjust, as well as a function for properly setting the value / establishing default shiftSize.

**Additional Documentation:**  
image

## Approval Checklist

### DIFF
--- a/src/include/device.h
+++ b/src/include/device.h
@@ -330,22 +330,18 @@ inline __host__ uint8_t ncclP2pChannelBaseForRound(struct ncclComm* comm, int p2
 
 // ncclP2pChannelToPart and ncclP2pChannelForPart are inverses. The device code
 // uses ncclP2pChannelToPart to determine which part "this" channel is responsible for.
-inline __host__ int ncclP2pChannelForPart(int nP2pChannels, int base, int part, int nParts, int nNodes) {
+inline __host__ int ncclP2pChannelForPart(int nP2pChannels, int base, int part, int nParts, int nNodes, int shiftSize) {
   if (nNodes > 2) {
     // Only works because nP2pChannels is pow2
-    int nChannelsLog2 = countOneBits(nP2pChannels-1);
-    int delta = reverseBits(part, nChannelsLog2);
-    return (base + delta) & (nP2pChannels-1);
+    return (base + ((part + (base>>shiftSize))<<shiftSize)) & (nP2pChannels-1);
   } else {
     return (base * nParts + part) & (nP2pChannels-1);
   }
 }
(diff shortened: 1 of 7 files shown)


src/include/device.h
```diff
@@ -330,22 +330,18 @@ inline __host__ uint8_t ncclP2pChannelBaseForRound(struct ncclComm* comm, int p2
...
   }
 }
-inline __device__ int ncclP2pChannelToPart(int nP2pChannels, int base, int channel, int nParts, int nNodes) {
+inline __device__ int ncclP2pChannelToPart(int nP2pChannels, int base, int channel, int nParts, int nNodes, int shiftSize) {
   if (nNodes > 2) {
     // Only works because nP2pChannels is pow2
-    int nChannelsLog2 = countOneBits(nP2pChannels-1);
-    int delta = (channel-base) & (nP2pChannels-1);
-    return reverseBits(delta, nChannelsLog2);
+    return (((channel - base) - ((base>>shiftSize)<<shiftSize))>>shiftSize) & (nParts-1);
```
mustafabar: `nParts` should reflect `CHANNELS_PER_NET_PEER`. This is usually a smaller number << nChannels,  and 1 at large scale, which results in ANDing with 0. I have not debugged this to see what the final value in `ncclShmem.comm.p2pnChannelsPerPeer` is, but believe it is safer if we AND with `nP2PChannels-1` instead? I wonder how this worked when tested by AINIC team. 
Note that NCCL completely avoids nParts in its functions https://github.com/NVIDIA/nccl/blob/dbc86fd06e8b0c4517b95d8958a09ccacf9520c9/src/include/device.h#L270

mustafabar: Ok, thought about it again, I think this will hang if we force channels per net peer for > 2 nodes to a value > 1. It usually ends up auto tuned to 1 for > 2 nodes, so nParts is 1 and part id is 0 therefore it went unnoticed

ammallya:
Cannot import to ROCm/rocm-systems due to merge conflicts please reopen this PR on rocm/rocm-systems.

PR #2068, opened 2025-11-19
Add ncclCommDump API

ahmd-k:
# Context
PyTorch Distributed's ProcessGroupNCCL has a watchdog thread that detects collective hangs or errors. For Meta's NCCLX and RCCLX, this triggers a custom NCCL API we have defined called `ncclCommDump()`, which prints data from our ProxyTrace. This is useful because we can't rely on the ProxyTrace print in `commFree()`, since PyTorch Distributed does not always clean up all the communicators during NCCL errors.

We want to open-source this functionality so that PTD jobs on open-source RCCL can have the same functionality.

# Implementation Details
- In a new file, add barebones `ncclCommDump()` API that dumps the ProxyTrace contents, similar to our current behavior in `commFree()`
  - Add `ncclCommDump()` to nccl.h
  - Modify logs so we can distringuish between ncclCommDump and commFree output
- Introduce timestamps for post and lastSend, and add them to the ProxyTrace dump
  - Add a map from proxy op stage to timestamp
  - Add helper function for setting the timestamps
  - Add calls to set the timestamps in `transport/net.cc`
- Made the public APIs for ProxyTrace thread-safe
  - Instead of having the APIs as non-member functions that called ProxyTrace member functions underneath, moved them into the class. This addition layer of indirection was confusing and made it difficult to synchornize the APIs with a shared mutex.
     - `proxyTraceInit()` -> `ProxyTrace::ProxyTrace()`
     - `updateProxyOpCounter()`
     - `setProxyOpTimestamp()`
     - `addNewProxyOp()`
   - Added a lock acquire at the start of each public API call. 
   - Made internal implementation details like `addNewProxyTraceOpImpl()` private. These will not be called by other RCCL code directly and should not be public
   - Updated unit tests
     - Removed resetAll() and its corresponding unit test. This is not used anywhere.
     - Removed unnecessary testing of getOrCreateProxyOpId(). We should just test the public API calling it underneath (`addNewProxyOp()`)

# Testing
Testing these changes requires a multi-node setup. We did this internally with a 2-node PyTorch Distributed job, as well as a 2-node MPI test. We have a somewhat convoluted Python script for doing this, but the final run arguments look like this
```
" ".join(
    [
        f"{MPIRUN}",
        f"-x LD_LIBRARY_PATH={OMPI_LIB}:{RCCL_BUILD_DIR_LIB}",
        f"-np {total_gpus}",
        f"-host {hosts_with_slots}",
        "--allow-run-as-root",
        "-x RCCL_ENABLE_PROXY_TRACE=1",
        "-x NCCL_DEBUG=WARN",
        "-x HSA_NO_SCRATCH_RECLAIM=1",
        f"{rccl_test_binary_path} -b {args.minbytes} -e {args.maxbytes} -g 1 -f 2",
    ]
),
```
This is what the logs look like at the end when the communicator is dumped.
```
[0] ~/rccl/build/release/hipify/src/init.cc:445 NCCL WARN commDump for all active ops mapSizeMB:0.01
createT:1763627168308, lastT:1763627168328, postT:0, sendT:0, cntNm:7, <1502607184520987562:1367:0>, [fu,pr,pa,tb,ck]:4,2,1,0,2097152, 0->8(R), chan:53, status:DONE, ns:240, nb:1048576, po:240, ke:0, tail/h:39240, recvT:39240, connSz/h:39238, trans:240, flushed:240, recvd:240, done:240
createT:1763627168308, lastT:1763627168329, postT:1763627168328, sendT:1763627168328, cntNm:7, <1502607184520987562:1367:10>, [fu,pr,pa,tb,ck]:4,2,1,0,2097152, 0->8(S), chan:21, status:DONE, ns:240, nb:1048576, po:240, ke:240, tail/h:39238, recvT:39240, connSz/h:1048576, trans:240, flushed:0, recvd:0, done:240
createT:1763627168308, lastT:1763627168329, postT:1763627168328, sendT:1763627168329, cntNm:7, <1502607184520987562:1367:11>, [fu,pr,pa,tb,ck]:4,2,1,0,2097152, 0->8(S), chan:42, status:DONE, ns:240, nb:1048576, po:240, ke:240, tail/h:39238, recvT:39240
```
We also pass the unit tests
```
$ sudo ONLY_FUNCS="AllReduce * * Sum f32" ./install.sh --local_gpu_only --disable-msccl-kernel --prefix /tmp/rccl 
$ UT_DATATYPES=ncclfloat32 UT_REDOPS=prod /tmp/rccl/bin/rccl-UnitTests --gtest_filter="ProxyTraceTestFixture.*"
Note: Google Test filter = ProxyTraceTestFixture.*
[==========] Running 5 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 5 tests from ProxyTraceTestFixture
[ RUN      ] ProxyTraceTestFixture.nonEmptySingleton
[       OK ] ProxyTraceTestFixture.nonEmptySingleton (0 ms)
[ RUN      ] ProxyTraceTestFixture.addTraceOp
[       OK ] ProxyTraceTestFixture.addTraceOp (0 ms)
[ RUN      ] ProxyTraceTestFixture.getMapSizeMB
[       OK ] ProxyTraceTestFixture.getMapSizeMB (0 ms)
[ RUN      ] ProxyTraceTestFixture.updateTraceOp
[       OK ] ProxyTraceTestFixture.updateTraceOp (0 ms)
[ RUN      ] ProxyTraceTestFixture.updateTraceOp2
[       OK ] ProxyTraceTestFixture.updateTraceOp2 (0 ms)
[----------] 5 tests from ProxyTraceTestFixture (0 ms total)

[----------] Global test environment tear-down
[==========] 5 tests from 1 test suite ran. (0 ms total)
[  PASSED  ] 5 tests.
```
# References
ProcessGroupNCCL dump call: https://github.com/pytorch/pytorch/blob/fcc78410a8e51107a7f4a15431e57da137741aee/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L400-L412

### DIFF
--- a/src/commDump.cc
+++ b/src/commDump.cc
@@ -1,0 +1,26 @@
+// (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+
+#include "nccl.h"
+#include <cstring>
+#include "comm.h"
+#include "device.h"
+#include "archinfo.h"
+
+__attribute__ ((visibility("default")))
+ncclResult_t ncclCommDump(
+    const ncclComm_t comm,
+    std::unordered_map<std::string, std::string>& map) {
+  if (comm == nullptr) {
+    WARN("ncclCommDump comm is null");
+    return ncclSuccess;
+  }
+  if (comm->proxyState->proxyTrace == nullptr) {
+    WARN("ncclCommDump comm->proxyState->proxyTrace is null");
+    return ncclSuccess;
+  }
+
+  WARN("ncclCommDump() ProxyTrace:");
+  WARN("%s", comm->proxyState->proxyTrace->dump().c_str());
+
+  return ncclSuccess;
+}
--- a/src/init.cc
+++ b/src/init.cc
@@ -168,7 +168,7 @@ ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
     else {
       return ncclSuccess;
     }
-  #endif
+  #endif   
 }
 
 static void initOnceFunc() {
@@ -1454,7 +1454,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
     }
   }
   // For single node communicators that do not uses the full xgmi links per gpu, i.e., nranks < 8
-  // Inflate the nChannels a bit to achieve higher b/w.
+  // Inflate the nChannels a bit to achieve higher b/w. 
   if (IsArchMatch(comm->topo->nodes[GPU].nodes[idx].gpu.gcn, "gfx950")) {
     if (nranks == 2 && nNodes == 1){
       allGather3Data[rank].nc = 16;
@@ -1817,8 +1817,8 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   // Compute time models for algorithm and protocol combinations
   NCCLCHECKGOTO(ncclTopoTuneModel(comm, comm->minCompCap, comm->maxCompCap, graphs), ret, fail);
 
-  INFO(NCCL_INIT, "comm:%p, nRanks:%d, nNodes:%d, coll channels:%d collnet channels:%d, nvls channels:%d, p2p channels:%d, p2p channels per peer:%d", comm, comm->nRanks, comm->nNodes, comm->nChannels, comm->nChannels, comm->nvlsChannels, comm->p2pnChannels, comm->p2pnChannelsPerPeer);
-
+  INFO(NCCL_INIT, "comm:%p, nRanks:%d, nNodes:%d, coll channels:%d collnet channels:%d, nvls channels:%d, p2p channels:%d, p2p channels per peer:%d", comm, comm->nRanks, comm->nNodes, comm->nChannels, comm->nChannels, comm->nvlsChannels, comm->p2pnChannels, comm->p2pnChannelsPerPeer);  
+  
   if (comm->intraRank == 0) { // Load ncclParamLaunchMode
     const char* str = ncclGetEnv("NCCL_LAUNCH_MODE");
     enum ncclLaunchMode mode, modeOld;
@@ -2075,10 +2075,10 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   comm->cuCount = cuCount;
 
   NCCLCHECKGOTO(initTransportsRank(comm, job->parent, timers), res, fail);
-
+  
     // Check if using host uncached mem correctly
   NCCLCHECK(checkHostUncacheMemSetting(comm));
-
+  
   // RCCL: determine and set unroll factor for comm
   NCCLCHECK(commSetUnrollFactor(comm));
 
--- a/src/init.cc
+++ b/src/init.cc
@@ -168,7 +168,7 @@ ncclResult_t checkHostUncacheMemSetting(struct ncclComm* comm) {
     else {
       return ncclSuccess;
     }
-  #endif   
+  #endif
 }
 
 static void initOnceFunc() {
@@ -439,7 +439,7 @@ static ncclResult_t commFree(ncclComm_t comm) {
   free(comm->connectRecv);
 
   if (rcclParamEnableProxyTrace()) {
-    WARN("ProxyTrace:");
+    WARN("commFree() ProxyTrace:");
     if (comm->proxyState && comm->proxyState->proxyTrace){
       WARN("%s", comm->proxyState->proxyTrace->dump().c_str());
     }
@@ -1454,7 +1454,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
     }
   }
   // For single node communicators that do not uses the full xgmi links per gpu, i.e., nranks < 8
-  // Inflate the nChannels a bit to achieve higher b/w. 
+  // Inflate the nChannels a bit to achieve higher b/w.
   if (IsArchMatch(comm->topo->nodes[GPU].nodes[idx].gpu.gcn, "gfx950")) {
     if (nranks == 2 && nNodes == 1){
       allGather3Data[rank].nc = 16;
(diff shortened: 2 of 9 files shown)


src/init.cc
```diff
@@ -1454,7 +1454,7 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
     }
   }
   // For single node communicators that do not uses the full xgmi links per gpu, i.e., nranks < 8
-  // Inflate the nChannels a bit to achieve higher b/w. 
+  // Inflate the nChannels a bit to achieve higher b/w.
```
dmwu: @ahmd-k   avoiding these unnecessary edits?

src/init.cc
```diff
@@ -2075,10 +2075,10 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   comm->cuCount = cuCount;
 
   NCCLCHECKGOTO(initTransportsRank(comm, job->parent, timers), res, fail);
-  
+
```
dmwu: ditto

src/init.cc
ahmd-k: Resolved

src/init.cc
ahmd-k: Resolved

src/misc/proxy_trace/proxy_trace.cc
```diff
@@ -249,6 +255,22 @@ void facebook_rccl::updateProxyOpCounter(
...
+  if (!proxyTraceObj) {
+    return;
+  }
+
+  auto traceOpPtr = proxyTraceObj->getProxyTraceOpPtr(traceKey);
+  if (!traceOpPtr || traceOpPtr->timestamps.find(counter) == traceOpPtr->timestamps.end()) {
+    return;
+  }
+
+  traceOpPtr->timestamps[counter] = std::chrono::high_resolution_clock::now();
```
alex-breslow-amd, line 247: To be pedantic with timing, you may want to sandwich timing calls between atomic signal fences or employ the  DoNotOptimize builtin.  The compiler does not guarantee that timing calls aren't reordered relative to other blocks of code.

src/commDump.cc
```diff
@@ -0,0 +1,26 @@
...
+#include "nccl.h"
+#include <cstring>
+#include "comm.h"
+#include "device.h"
+#include "archinfo.h"
+
+__attribute__ ((visibility("default")))
+ncclResult_t ncclCommDump(
+    const ncclComm_t comm,
+    std::unordered_map<std::string, std::string>& map) {
```
alex-breslow-amd, line 12: Make **const** std::unordered_map<std::string, std::string>& map

alex-breslow-amd: I may have missed it, but why not pass in ncclComm_t as a const reference rather than passing it by value?

src/transport/net.cc
```diff
@@ -1314,6 +1314,7 @@ static ncclResult_t sendProxyProgress(struct ncclProxyState* proxyState, struct
         }
         ncclProfilerRecordProxyStepEventState(s, args, postedStepId, ncclProfilerProxyStepSendGPUWait);
         facebook_rccl::updateProxyOpCounter(proxyState->proxyTrace, sub->traceKey, facebook_rccl::ProxyCounterTypes::POSTED, sub->posted);
+        setProxyOpTimestamp(proxyState->proxyTrace, sub->traceKey, facebook_rccl::ProxyCounterTypes::POSTED);
```
alex-breslow-amd: In keeping with the style with other Meta-added features, I'm guessing it makes sense to put this in the facebook_rccl namespace.

src/commDump.cc
ahmd-k: > Make **const** std::unordered_map<std::string, std::string>& map

In NCCLX and RCCLX, this map is not const because it is where we store some structured trace data that callers like PyTorch can use. 

See the NCCLX API: https://github.com/meta-pytorch/torchcomms/blob/fe4e8116f2107b5aed0e38db10e072471ea95126/comms/ncclx/v2_27/meta/commDump.cc#L219

ahmd-k: > I may have missed it, but why not pass in ncclComm_t as a const reference rather than passing it by value?

Good point, I was just following the NCCLX implementation but I don't see why we copy the communicator here. @dmwu @YulunW any idea why the the communicator is passed by value in `ncclCommDump()`?

ahmd-k:
I've updated the PR to fix issues with thread safety.

src/commDump.cc
corey-derochie-amd: `ncclComm_t` is a typedef of a pointer, right? So pass by value is acceptable here, @alex-breslow-amd ? This is pretty standard in the API, no?

src/commDump.cc
```diff
@@ -0,0 +1,26 @@
+// (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+
+#include "nccl.h"
+#include <cstring>
+#include "comm.h"
+#include "device.h"
+#include "archinfo.h"
+
+__attribute__ ((visibility("default")))
```
corey-derochie-amd, line 9: Should this use the `NCCL_API` macro?

corey-derochie-amd (changes requested):
If we make additive changes to our public API, we should increment API version in api_trace.h.

src/commDump.cc
ahmd-k: NCCL_API sets `extern "C"`, so it won't work here since ncclCommDump uses CPP libraries.

Trying to use NCCL_API would return the following:
`error: declaration of 'ncclCommDump' has a different language linkage`

ahmd-k:
> If we make additive changes to our public API, we should increment API version in api_trace.h.

The functions listed in api_trace.h are all compatible with C and are setup to have C linkage with `RCCL_EXTERN_C_INIT`. `ncclCommDump`, on the other hand, is only compatible with C++. That's why in `nccl.h.in`, `ncclCommDump` is declared after ending the `extern C`.

Does it still make sense to increment the API version?

corey-derochie-amd:
> > If we make additive changes to our public API, we should increment API version in api_trace.h.
> 
> The functions listed in api_trace.h are all compatible with C and are setup to have C linkage with `RCCL_EXTERN_C_INIT`. `ncclCommDump`, on the other hand, is only compatible with C++. That's why in `nccl.h.in`, `ncclCommDump` is declared after ending the `extern C`.
> 
> Does it still make sense increment the API version?

This is a good question, I will have to sync with another team to answer this, they have build gates that will fail if the public API changes without a version bump. I'll report back how this is meant to work.

corey-derochie-amd:
> > > If we make additive changes to our public API, we should increment API version in api_trace.h.
> > 
> > 
> > The functions listed in api_trace.h are all compatible with C and are setup to have C linkage with `RCCL_EXTERN_C_INIT`. `ncclCommDump`, on the other hand, is only compatible with C++. That's why in `nccl.h.in`, `ncclCommDump` is declared after ending the `extern C`.
> > Does it still make sense increment the API version?
> 
> This is a good question, I will have to sync with another team to answer this, they have build gates that will fail if the public API changes without a version bump. I'll report back how this is meant to work.

I have spoken with rocprofiler team and concluded we can exclude `ncclCommDump` from our standard RCCL API, so no need for this here.

ahmd-k:
> > > > If we make additive changes to our public API, we should increment API version in api_trace.h.
> > > 
> > > 
> > > The functions listed in api_trace.h are all compatible with C and are setup to have C linkage with `RCCL_EXTERN_C_INIT`. `ncclCommDump`, on the other hand, is only compatible with C++. That's why in `nccl.h.in`, `ncclCommDump` is declared after ending the `extern C`.
> > > Does it still make sense increment the API version?
> > 
> > 
> > This is a good question, I will have to sync with another team to answer this, they have build gates that will fail if the public API changes without a version bump. I'll report back how this is meant to work.
> 
> I have spoken with rocprofiler team and concluded we can exclude `ncclCommDump` from our standard RCCL API, so no need for this here.

Thanks for getting back to me on this. Does this mean we can merge the PR?

alex-breslow-amd:
@ahmd-k, we are currently migrating RCCL to the rocm-systems super repo similar to what Meta has with fbcode.  This PR will get migrated there.  We can land it there.  After today, all RCCL PRs should be made against rocm-systems: https://github.com/ROCm/rocm-systems/tree/develop/projects/rccl

Update, we will continue to use this repository for the time being.  Migration has been delayed.

src/commDump.cc
alex-breslow-amd: Ah ok.

alex-breslow-amd (approved):

corey-derochie-amd (approved):

Issue #2044, opened 2025-11-11
rcclNetP2pPolicy returns ncclInvalidArgument when not using infiniband


ryanhankins:
### Problem Description

The following change in RCCL breaks aws-ofi-nccl.  rcclNetP2pPolicy returns ncclInvalidArgument.  My guess is there is no ncclIbHandle when not using IB.  The change prior to this one works.

https://github.com/ROCm/rccl/commit/45991fadadefe44d31ff8fd8278be981d57a3129

### Operating System

SUSE Linux Enterprise Server 15 SP6

### CPU

AMD EPYC 7A53 64-Core Processor

### GPU

MI250X

### ROCm Version

top of tree devel

### Steps to Reproduce

Run multi-node rccl-tests all_reduce_perf with this change on Slingshot with aws-ofi-nccl plugin.

ryanhankins:
@AbandiGa

PR #2072, opened 2025-11-21
Fix rcclNetP2pPolicy issue

AbandiGa:
## Details

**What were the changes?**  
add check for ncclNet before calling rcclNetP2pPolicy 

Fix this issue #[2044](https://github.com/ROCm/rccl/issues/2044), Since rcclNetP2pPolicy is declared as extern in net.h and implemented in the stock IB backend, the previous if (rcclNetP2pPolicy) check was incorrect. It originated from the earlier weak-symbol implementation, and when the implementation changed and the weak symbol was removed, the condition was always evaluated to true.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/transport/net.cc
+++ b/src/transport/net.cc
@@ -776,7 +776,8 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   NCCLCHECK(ncclNetGetDeviceHandle(resources->netDeviceType, resources->netDeviceVersion, false /*isRecv*/, &resources->netDeviceHandle));
   bool rccl_anp = !(strcmp(proxyState->ncclNet->name, RCCL_ANP_PLUGIN_STR));
   
-  if (rcclNetP2pPolicy) {
+  // Only call rcclNetP2pPolicy for InfiniBand
+  if (proxyState->ncclNet == &ncclNetIb) {
     NCCLCHECK(rcclNetP2pPolicy(req->handle, resources->isP2p));
   }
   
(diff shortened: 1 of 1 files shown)


src/transport/net.cc
```diff
@@ -776,7 +776,8 @@ static ncclResult_t sendProxyConnect(struct ncclProxyConnection* connection, str
   NCCLCHECK(ncclNetGetDeviceHandle(resources->netDeviceType, resources->netDeviceVersion, false /*isRecv*/, &resources->netDeviceHandle));
   bool rccl_anp = !(strcmp(proxyState->ncclNet->name, RCCL_ANP_PLUGIN_STR));
   
-  if (rcclNetP2pPolicy) {
+  // Only call rcclNetP2pPolicy for InfiniBand
+  if (proxyState->ncclNet == &ncclNetIb) {
```
mustafabar, line 780: What happens when the function is not implemented? Pls get time on the AINIC system to test this further.

mustafabar (changes requested):

src/transport/net.cc
AbandiGa: Can you clarify? The function is implemented for net_ib, and with the current check, it will only be called when using net_ib. If another network plugin is used, then it won’t be invoked. Previous condition was tied to the weak-symbol implementation. After removing the weak symbol, I should have updated that logic.

karthikarum: Tested with this PR on (2N) AINIC cluster and observed RCCL is running w/o issues.

mustafabar: Right, I see now — since `rcclNetP2pPolicy` is declared as `extern` in `net.h` and implemented in stock IB backend, the old `if (rcclNetP2pPolicy)` check was wrong and got inherited from the initial the weak-symbol implementation and was effectively always true.

mustafabar: Suggest to capture my comment above in the PR description since the check was always true. 
Also, in the comment you mentioned `Only call rcclNetP2pPolicy for InfiniBand`. The RCCL net_ib supports both IB and RoCE. 
The comment has to capture that the function is called when ncclNetIb is used as a network transport.

mustafabar (approved):

nileshnegi (approved):

PR #2073, opened 2025-11-22
[Device] WarpSpeed enablement and single node CU and perf opt for MI350

mustafabar:
**Work item:** _Internal_

**What were the changes?**  
This pull request introduces significant changes to enable per-warp channel assignment ("WarpSpeed") in RCCL device kernels. The main goal is to allow each GPU warp to operate on its own communication channel. Most device code is updated to reference the correct per-warp channel and channel ID, and kernel argument handling is generalized. 
Also added RCCL_WARP_SPEED_AUTO mode to auto select it for single node and reduce CU usage by 50% for >= 64 MB message sizes in AllReduce and AllGather, and for >=256 MB message sizes for ReduceScatter

image

It was found from TransferBench that higher/peak BW can be achieved on a single node with lower number of CUs when each warp within a CU transfers data to a remote GPU over XGMI on MI3xx platforms.

### WarpSpeed Enablement and Per-Warp Channel Assignment

* Added support for per-warp channel assignment in shared memory (`ncclShmem`), including new fields `warpChannelId` and `warpChannel`, and logic to assign channels to warps based on a mask and communication mode
* Updated kernel launch signatures and shared memory setup to use a generalized kernel argument storage type (`ncclDevKernelArgsDefaultStorage`), replacing the previous fixed-size type. 
### Device Collectives and Primitives Refactoring

* Update all device collective implementations (`all_gather.h`, `all_reduce.h`, `broadcast.h`, `reduce.h`, `reduce_scatter.h`) to use per-warp channel and channel ID for ring-based algorithms.
* Updated device primitives (`prims_simple.h`, `prims_ll.h`, `prims_ll128.h`) to reference per-warp channel data for peer notifications, barrier synchronization, and connection setup, supporting both warp and block-level communication. 

### Kernel Execution Logic

* Modified the main kernel execution logic to compute warp indices, assign channels per warp, and copy channel data to shared memory for each warp. Also added logic to handle both warp-level and block-level communication modes. 
* Updated collective work scheduling to make sure that warp groups are only used when warp communication is enabled.

### Miscellaneous

* Fixed a type issue in the enqueue logic, changing the type of `nChannels` to allow more than 128 channels. 

**Additional Documentation:**  
- This feature is only applicable to the ring algorithm. It is disabled for other algorithms such as Tree and Send/Recv collectives. 
- This feature has potential for e2e workloads: Notably higher relative peak BW when smaller grid is used (e.g. 16 x 256/512). 
- Higher multimode performance up to 32 x 256 grid dimensions
- Best microbenchmark performance requires medium size message algo/protocol/CU usage optimization for small to medium messages
- May need thresholding since it is not a latency-optimal approach (e.g. implementation uses more channel and synchronization)
- Currently provided as experimental feature for further testing with e2e requiring overlap with compute and lower CU usage from comms
- `RCCL_WARP_SPEED_AUTO` mode to auto selects it for single node and reduces CU usage by 50% for >= 64 MB message sizes in AllReduce and AllGather, and for >=256 MB message sizes for ReduceScatter
- To experiment peak performance impact on a single MI350 node use `RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256`

## Approval Checklist


src/rccl_wrap.cc
```diff
@@ -436,6 +514,15 @@ ncclResult_t rcclFuncMaxSendRecvCount(ncclFunc_t func, int nRanks, size_t count,
 }
 
 ncclResult_t commSetUnrollFactor(struct ncclComm* comm) {
+  if( rcclParamUnrollFactor() != -1 ) {
+    comm->unroll = rcclParamUnrollFactor(); //-1 to map to 0 based indexing
+    if(comm->unroll < NCCL_UNROLL_1 || comm->unroll >= NCCL_NUM_UNROLLS) {
+      WARN("Invalid RCCL_UNROLL_FACTOR %d specified. Valid values are 0 to 2 corresponding to unroll factors of 1, 2, and 4 respectively.", comm->unroll);
```
nileshnegi, line 555: this won't work for local gpu build targets, right, where we have a limited no. of unroll factors (depending on GPU targets)?

nileshnegi:
UnitTests failing due to MSCCL.

wenkaidu:
To reduce risk and make future NCCL sync easier, can we guard these code using #ifdef since this is a very significant code change?

mustafabar:
> UnitTests failing due to MSCCL.

I added a fix for MSCCL. I see there are still some failures in AllGather out-of-place. I am running those UTs locally to see

mustafabar:
> To reduce risk and make future NCCL sync easier, can we guard these code using #ifdef since this is a very significant code change?

@wenkaidu The code is currently a Macro chaos due to the NPKIT injections. I thought of this to make it less messy. Will this address your concern? any issues with this approach or ideas ?


```

​RCCL_INJECT_WS(
{
    // RCCL custom logic
},
{
    // NCCL original logic 
});

```




```
​#pragma once



#ifndef RCCL_WARP_SPEED
#define RCCL_WARP_SPEED 1
#endif



#if RCCL_WARP_SPEED



// run RCCL block, ignore NCCL block
#define RCCL_INJECT_WS(block_rccl, block_nccl) \
  do { block_rccl } while(0)



#else



// run NCCL block, ignore RCCL block
#define RCCL_INJECT_WS(block_rccl, block_nccl) \
  do { block_nccl } while(0)



#endif
```

src/rccl_wrap.cc
mustafabar: No it won't work unless all unrolls are built. It is more of a debugging feature and should not be recommended to users. We need to have a way to build all unrolls irrespective of the target for experimentation purposes

nileshnegi: it is possible to build with all unrolls, if you do not use -l and instead specify the target GPU like --amdgpu_targets=gfx950

wenkaidu:
> > To reduce risk and make future NCCL sync easier, can we guard these code using #ifdef since this is a very significant code change?
> 
> @wenkaidu The code is currently a Macro chaos due to the NPKIT injections. I thought of this to make it less messy. Will this address your concern? any issues with this approach or ideas ?
> 
> ```
> 
> ​RCCL_INJECT_WS(
> {
>     // RCCL custom logic
> },
> {
>     // NCCL original logic 
> });
> ```
> 
> ```
> ​#pragma once
> 
> 
> 
> #ifndef RCCL_WARP_SPEED
> #define RCCL_WARP_SPEED 1
> #endif
> 
> 
> 
> #if RCCL_WARP_SPEED
> 
> 
> 
> // run RCCL block, ignore NCCL block
> #define RCCL_INJECT_WS(block_rccl, block_nccl) \
>   do { block_rccl } while(0)
> 
> 
> 
> #else
> 
> 
> 
> // run NCCL block, ignore RCCL block
> #define RCCL_INJECT_WS(block_rccl, block_nccl) \
>   do { block_nccl } while(0)
> 
> 
> 
> #endif
> ```

Since this is a very specific enhancement for gfx950 only, it would be better to use cmake compilation option to have the capability to turn it off for majority of cases. I think ifdef is the best option as original code path is preserved, thus understanding/merging NCCL sync will be much easier.

mustafabar:
@wenkaidu I have added compiler based Macro `ENABLE_WARP_SPEED` that guards the changes and is only enabled for MI3xx by default (there's also opt out option with `--disable-warp-speed`)

src/device/common.h
```diff
@@ -583,9 +599,46 @@ __device__ __forceinline__ void ncclKernelMain(struct ncclDevKernelArgs const* a
...
+  if(tid == 0) {
+    ncclShmem.warpComm = args->comm->warpLevelComm;
+  }
 #endif
   __syncthreads(); // publish shmem
 
+#ifdef ENABLE_WARP_SPEED
+  // Determine per-warp channel assignment for WarpSpeed enablement
+  total = 0;
+  if(ncclShmem.warpComm == 1) {  // If warpComm is enabled, assing warps to channels that have the corresponding channel mask enabled
```
alex-breslow-amd: Fix comment

src/device/prims_ll.h
```diff
@@ -652,9 +652,13 @@ class Primitives<T, RedOp, Fan, Direct, ProtoLL, P2p, isNetOffload, Metadata, Pi
       bool ipcReg = false, bool netReg = false, int stepSize_ = 0
     ):
     redOp(redOpArg),
-    tid(tid), nthreads(nthreads), wid(tid%WARP_SIZE), group(group), threadsPerBlock(blockDim.x),
+    tid(tid), nthreads(nthreads), wid(threadIdx.x%WARP_SIZE), group(group), threadsPerBlock(blockDim.x),
```
alex-breslow-amd: Is the wid(threadIdx.x%WARP_SIZE) change safe in general?  I think this may need to be contingent on whether it's WarpSpeed.  Let me know if I am wrong.  If I recall correctly, there are cases where there is virtualization of the thread ID like in AllToAll, so the wid in that case is also virtual and not threadIdx.x%WARP_SIZE.

src/device/prims_simple.h
```diff
@@ -81,18 +81,14 @@ class Primitives<
 
   // Don't use barrier 0 as it's used by the final sync
   inline __device__ void barrier() {
-    if (nthreads == WARP_SIZE)
```
alex-breslow-amd, line 84: I think we may need something here to prevent code motion in the general case like an __atomic_signal_fence.  That's effectively what __syncwarp() is doing for us here I believe since our warps can't diverge at the hardware level.  Same with below.

src/enqueue.cc
```diff
@@ -2167,7 +2171,13 @@ static ncclResult_t topoGetAlgoInfo(
     }
   } else if (info->func == ncclFuncAllReduce && comm->topo->treeDefined == 1) {
     info->algorithm = NCCL_ALGO_TREE;
+#ifdef ENABLE_WARP_SPEED
+    nc = std::min(nc, 64); // Tree uses at most 64 channels as we don't support WarpSpeed Tree.
+  } else if (info->algorithm == NCCL_ALGO_TREE) {
+    nc = std::min(nc, 64); // Tree uses at most 64 channels as we don't support WarpSpeed Tree.
```
alex-breslow-amd, line 2183: I'd prefer making the 64 a variable that gets piped in from someone and will get updated in the event that the channels for Tree change.  This seems like a footgun.

alex-breslow-amd:
This PR is an impressive bit of engineering.  I think I share Wenkai's concerns about potential long-term maintenance challenges with these changes since there are many places where this could break with subsequent merges.  What is your plan regarding testing so that we can detect breakages on merges?

src/init.cc
```diff
@@ -1453,8 +1455,12 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
...
   }
+#ifdef ENABLE_WARP_SPEED
+  comm->topo->warpSpeedEnabled = comm->topo->warpSpeedEnabled && rcclParamWarpSpeedEnable(); // only use for MI3xx GPUs if enabled by user
+#endif
+
   // For single node communicators that do not uses the full xgmi links per gpu, i.e., nranks < 8
-  // Inflate the nChannels a bit to achieve higher b/w. 
+  // Inflate the nChannels a bit to achieve higher b/w.
   if (IsArchMatch(comm->topo->nodes[GPU].nodes[idx].gpu.gcn, "gfx950")) {
     if (nranks == 2 && nNodes == 1){
```
alex-breslow-amd, line 1466: Meta added this code to use more channels per GPU when there are fewer than 8 GPUs being used.  It appears that we override it below at lines 1473 when WarpSpeed is enabled.  I think you need to check that this change doesn't degrade performance for 2-GPU and 4-GPU configurations.  If you have, then ignore me.

src/rccl_wrap.cc
```diff
@@ -398,6 +407,66 @@ void rcclSetP2pNetChunkSize(struct ncclComm* comm,  int& rcclP2pNetChunkSize) {
   }
   rcclP2pNetChunkSize = p2pNetChunkSize;
 }
+#ifdef ENABLE_WARP_SPEED
```
alex-breslow-amd, line 417: You have ```#ifdef ENABLE_WARP_SPEED``` twice here, and they are nested.  I'm guessing you don't need both.

src/device/common.h
mustafabar: done

src/device/prims_ll.h
mustafabar: Makes sense. Having thought about it, threadId.x%warpsize is no different than (threadId.x%warpsize)%warpsize which I intended to address in this change. So it is safe to revert it

src/device/prims_simple.h
mustafabar: As discussed, I went with __syncwarp(), as this evaluation is out of scope for the PR

src/init.cc
mustafabar: Done

src/enqueue.cc
mustafabar: I'd leave it at that for now, since this is compile-time gated by MI3XX, so should not be a problem to use at most 64 channels for tree. I don't want to add yet another control param.

mustafabar:
> This PR is an impressive bit of engineering. I think I share Wenkai's concerns about potential long-term maintenance challenges with these changes since there are many places where this could break with subsequent merges. What is your plan regarding testing so that we can detect breakages on merges?

Thanks @alex-breslow-amd - based on the CI unit tests, I don't see any red flags and will check with Prod team before merging. 
I labeled this as "regression-detection" to try and catch regressions caused for the default (block-based) path. Otherwise, it has to go through testing and validation for upcoming release

src/device/common.h
```diff
@@ -583,9 +599,46 @@ __device__ __forceinline__ void ncclKernelMain(struct ncclDevKernelArgs const* a
...
+      if (args->channelMask.masks[i] & (1ull<<laneId)) {
+        y = __popcll(args->channelMask.masks[i] & ((1ull<<laneId)-1));
+        y = total + y;
+        if (globalWarpId == y) {
+          ncclShmem.warpChannelId[localWarpId] = laneId + total;
+          break;
+        }
+      }
+      total = total + __popcll(args->channelMask.masks[i]);
+    }
```
wenkaidu, line 626: I didn't get why warpChannelId[localWarpId] is needed and recalculated. Isn't it getting same value as original channelId calculation?

wenkaidu:
Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.

mustafabar:
> ### regression-detection run on commit [d808362](https://github.com/ROCm/rccl/commit/d8083624b269c53b5585d585b99178003d0ee09e)
> [Artifacts](http://math-ci.amd.com/job/main/job/batch/job/rccl/job/PR-2073/1/artifact/run1/regression-detector/results/) - [Results](http://math-ci.amd.com/job/main/job/batch/job/rccl/job/PR-2073/1/artifact/run1/regression-detector/results//127111/regressions/regressions_preview.txt)

commit#fe783766922d95e78872fe932921cd7846bb9a45 is pushed to address the gather (send/recv) regression

mustafabar:
> Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.

Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450

src/device/common.h
mustafabar: Since each channel is associated with a warp rather than a block, we need to check the bitmask for each globalWarpId, and get the channel ID accordingly. 

The vanilla logic uses blockId.x and needs 1 warp to search for the block's channel id in parallel
https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/device/common.h#L526

The warp-logic uses globalWarpId and needs all warps, so that each warp finds its own channel id
https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/device/common.h#L620

wenkaidu:
> > Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.
> 
> Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450

It would be better to add this information to documentation. Can you also clarify how to use the feature in real world workloads where there are mixed single and multi-node communicators? For example, RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256, are these applicable to both single and multi-node?

alex-breslow-amd (approved):
This looks reasonable to me.  It's gated, so I don't see an issue.  I'll let others weigh in.   I think @wenkaidu raises a good point though.

mustafabar:
> > > Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.
> > 
> > 
> > Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450
> 
> It would be better to add this information to documentation. Can you also clarify how to use the feature in real world workloads where there are mixed single and multi-node communicators? For example, RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256, are these applicable to both single and multi-node?

I am adding more information to "Additional documentation", and will add a line in change log.
The environment variables are just knobs for optimization. In terms of functionality, they should be functional for both single and multi-node. Eventually, values like (CUs, threads, etc) will be auto set on a per communicator basis.

wenkaidu:
> > > > Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.
> > > 
> > > 
> > > Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450
> > 
> > 
> > It would be better to add this information to documentation. Can you also clarify how to use the feature in real world workloads where there are mixed single and multi-node communicators? For example, RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256, are these applicable to both single and multi-node?
> 
> I am adding more information to "Additional documentation", and will add a line in change log. The environment variables are just knobs for optimization. In terms of functionality, they should be functional for both single and multi-node. Eventually, values like (CUs, threads, etc) will be auto set on a per communicator basis.

I am still unclear on effect of setting "RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256": after setting these env vars, will warp speed be enabled with these settings for all communicators in FSDP workloads, i.e. single/multi-node, all 8 GPUs vs 1/2/4 GPUs on a node?

mustafabar:
> > > > > Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.
> > > > 
> > > > 
> > > > Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450
> > > 
> > > 
> > > It would be better to add this information to documentation. Can you also clarify how to use the feature in real world workloads where there are mixed single and multi-node communicators? For example, RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256, are these applicable to both single and multi-node?
> > 
> > 
> > I am adding more information to "Additional documentation", and will add a line in change log. The environment variables are just knobs for optimization. In terms of functionality, they should be functional for both single and multi-node. Eventually, values like (CUs, threads, etc) will be auto set on a per communicator basis.
> 
> I am still unclear on effect of setting "RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256": after setting these env vars, will warp speed be enabled with these settings for all communicators in FSDP workloads, i.e. single/multi-node, all 8 GPUs vs 1/2/4 GPUs on a node?


 @wenkaidu I have added WARP_SPEED_AUTO_MODE so that it can be suggested for customers (currently enables it for single-node and from certain message size). In this mode, we can gradually enable WarpSpeed wherever it performs well automatically, as we incrementally tune our CU usage. The other environment variables are for future debugging and optimization work.

wenkaidu:
> > > > > > Can you comment on how tree is supposed to be handled? Double binary tree and PAT tree need more than one warp per channel.
> > > > > 
> > > > > 
> > > > > Tree is not supported. There's a fallback logic to original channel-per-block usage when for any usage of non-ring algorithm. https://github.com/mustafabar/rccl/blob/fe783766922d95e78872fe932921cd7846bb9a45/src/rccl_wrap.cc#L450
> > > > 
> > > > 
> > > > It would be better to add this information to documentation. Can you also clarify how to use the feature in real world workloads where there are mixed single and multi-node communicators? For example, RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256, are these applicable to both single and multi-node?
> > > 
> > > 
> > > I am adding more information to "Additional documentation", and will add a line in change log. The environment variables are just knobs for optimization. In terms of functionality, they should be functional for both single and multi-node. Eventually, values like (CUs, threads, etc) will be auto set on a per communicator basis.
> > 
> > 
> > I am still unclear on effect of setting "RCCL_WARP_SPEED_ENABLE=1 RCCL_UNROLL_FACTOR=1 RCCL_WARP_SPEED_CU_COUNT=56 RCCL_THREADS_PER_BLOCK=256": after setting these env vars, will warp speed be enabled with these settings for all communicators in FSDP workloads, i.e. single/multi-node, all 8 GPUs vs 1/2/4 GPUs on a node?
> 
> @wenkaidu I have added WARP_SPEED_AUTO_MODE so that it can be suggested for customers (currently enables it for single-node and from certain message size). In this mode, we can gradually enable WarpSpeed wherever it performs well automatically, as we incrementally tune our CU usage. The other environment variables are for future debugging and optimization work.

Thanks for the update. What will be number of CU used in case of 2 and 4 GPUs on single node? Juts want to confirm these use cases are also being taken care of.

src/rccl_wrap.cc
```diff
@@ -437,9 +438,10 @@ void rcclSetWarpSpeedCUs(struct ncclComm* comm, int algo, int threadsPerBlock, i
       INFO(NCCL_INIT, "RCCL Warp CU count set to user defined %d resulting in %d channels", rcclParamWarpSpeedCuCount(), rcclWarpSpeedChannels);
       return;
     }
+    // printf("RCCL Warp Speed Channels before adjustment: %d, warpsPerBlock: %d\n", rcclWarpSpeedChannels, warpsPerBlock);
     // reuse the existing channel tuning logic if possible
     if (comm->nNodes == 1) {
-      rcclWarpSpeedChannels = std::min(224, rcclWarpSpeedChannels * warpsPerBlock * 2);
+      rcclWarpSpeedChannels = std::min(224 * comm->localRanks / 8, rcclWarpSpeedChannels * warpsPerBlock * 2);
```
wenkaidu: I think 224 is for 8 GPUs and may not be valid with all subscriptions. Currently we use 32 channels for 2 GPUs which will be translated to 128. For 4 GPUs, default channel count is 48 which should be translated to 192.

mustafabar: This is a minimum value of 224 * (ratio of ranks being used). This results in 56 CUs for 8 ranks, 28 CUs for 4 ranks and 14 CUs for 2 ranks. 
For example, for 4 ranks on a single node, this yields `min(224 x 4 / 8 = 112, 32 * 4 * 2 = 256) = 112 channels` (112 channels results in 28 CUs for WarpSpeed since each CU consumes 4 channels due to the 256 thread block size for single node)

wenkaidu: When there are 4 GPUs, number of channels need to be multiple of 6 for even XGMI traffic distribution. 112 cannot be evenly divided by 6.
In general, using 224 is incorrect as it comes from multiple of 14 channels requirements with 8 GPUs.

mustafabar: Ah I see what you mean. It terms of performance, I couldn't spot a difference. There's issue if the number of channels does not divide the number of warps per CU. In this case, some warps will stay idle. So, if we choose 114 channels which is a multiple 6, it does not divide 4 and some warps will diverge. 
The best would be to go for a number that is a multiple of 6 x 4 for the 4 ranks case, and multiple of 2 x 4 for the 2 rank case. The exact number can be based on the lowest channel count that achieves perf, and we can just hard code it, as I don't see a point anymore in scaling 224 down. 
Thoughts?

wenkaidu: Existing RCCL can calculate correct number of channels. Would it be easier just use what RCCL calculated then multiply by a factor, i.e. rcclWarpSpeedChannels=original_channels*warpsPerBlock/2? Assume we always use half of CUs from original?

mustafabar: That works, but average perf is 1-2% lower. However, in the interest of better maintainability and clarity, we can go for a simpler math and less constants.

wenkaidu (approved):

PR #2077, opened 2025-11-26
Remove node-count and threshold restrictions from p2p-batching

isaki001:
## Details

**What were the changes?**  
- Enabled p2p-batching on more than 32 nodes
- removed restrictions on p2p-batching based on message size
- removed restrictions on p2p-batching that required the number of send and receive bytes to be the same

- hangs observed with p2p-batching enabled for more than 32 nodes

- used `__popcll`  instead of `__popc` in `loadWorkBatchToShmem`

**Additional Documentation:**  

## Approval Checklist


nileshnegi (approved):

mustafabar (approved):

src/enqueue.cc
mustafabar: do we need this?

src/enqueue.cc
```diff
@@ -1151,7 +1149,7 @@ static ncclResult_t addP2pToPlan(
     plan->channelMask.masks[channelId/64] |= uint64_t(1)<<(channelId%64);
     // Add batch first.
     int funcIdx = ncclDevFuncId_P2p();
-    addWorkBatchToPlan(comm, plan, channelId, ncclDevWorkTypeP2p, funcIdx, workOffset, p2pRound, batchP2P);
+    addWorkBatchToPlan(comm, plan, channelId, ncclDevWorkTypeP2p, funcIdx, workOffset, p2pRound, p2pBatchThreshold);
```
mustafabar: why are you passing threshold instead of flag?

mustafabar (changes requested):
pls address/resolve the comments

src/enqueue.cc
isaki001: Yes, I noticed that as well. It's a typo. I will fix this and add the by default enablement as well.

src/enqueue.cc
isaki001: No this will be removed as well.

wenkaidu (approved):

mustafabar (approved):

PR #2078, opened 2025-11-27
Add support for AMD AINIC within RCCL default internal network plugin.

karthikarum:
* provides out-of-the box experience for AMD AINIC
 * leverages the performance improvements offered by AINIC by default
 * enables channel pinning to pipeline for optimal performance
 * extends support for In-line CTS messages with 32B
 * Derive AINIC specific flags internally when RCCL AINIC param is set in env
 * Auto generate the AINIC net_ib with AINIC functionality from a patch file.
 *
Below is the miprun command line to run for in-box RCCL support for AINIC.
```
if [[ "$num_nodes" -eq "2" ]]; then                                                                                                                                                                            
    if [[ "$COLLECTIVE" == "alltoall" ]] || [[ "$COLLECTIVE" == "alltoallv" ]]; then                                                                                                                           
        PER_NET_PEER_ARG_OPT_STR="-x RCCL_DISABLE_RAIL_TREES=1"                                                                                                                                                
    fi                                                                                                                                                                                                         
fi                                                                                                                                                                                                             
                                                                                                                                                                                                               
if [[ "$COLLECTIVE" != "alltoall" ]]; then                                                                                                                                                                     
     EXT_STR="-x RCCL_LL128_FORCE_ENABLE=1"                                                                                                                                                                    
fi                                                                                                                                                                                                             
                                                                                                                                                                                                               
                                                                                                                                                                                                               
set -x                                                                                                                                                                                                                                   
${OMPI_INSTALL_DIR}/bin/mpirun --np ${NUM_PROCESS} \                                                                                                                                                                                     
-H ${HOST_LIST} \                                                                                                                                                                                                                        
--bind-to numa \                                                                                                                                                                                                                         
-x PATH=${PATH} \                                                                                                                                                                                                                        
-x LD_LIBRARY_PATH=${LD_LIBRARY_PATH} \                                                                                                                                                                                                  
--mca btl ^vader,openib \                                                                                                                                                                                                                
-mca btl_tcp_if_include ${OOB_IF} \                                                                                                                                                                                                      
-x NCCL_SOCKET_IFNAME=${OOB_IF} \                                                                                                                                                                                                        
-x NCCL_IB_QPS_PER_CONNECTION=${NUM_QP_PER_CHANNEL} \                                                                                                                                                                                    
-x NCCL_TOPO_DUMP_FILE=/tmp/system_run2.txt \                                                                                                                                                                                            
-x NCCL_IB_TC=${NCCL_IB_DATA_TOS} \                                                                                                                                                                                                      
-x NCCL_IB_FIFO_TC=${NCCL_IB_CTS_TOS} \                                                                                                                                                                                                  
-x NCCL_DEBUG=VERSION \                                                                                                                                                                                                                  
-x HSA_NO_SCRATCH_RECLAIM=1 \                                                                                                                                                                                                            
-x RCCL_GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING=0 \                                                                                                                                                                                        
-x NCCL_IGNORE_CPU_AFFINITY=1 \                                                                                                                                                                                                          
-x IONIC_LOCKFREE=all \                                                                                                                                                                                                                  
-x NCCL_PXN_DISABLE=0 \                                                                                                                                                                                                                  
-x NCCL_IB_PCI_RELAXED_ORDERING=1 \                                                                                                                                                                                                      
-x RCCL_AINIC_ROCE=1 \                                                                                                                                                                                                                   
${PER_NET_PEER_ARG_OPT_STR} \                                                                                                                                                                                                            
${EXT_STR} \                                                                                                                                                                                                                             
${RCCL_TESTS}/build/${COLLECTIVE}_perf -b ${START_MSG_SIZE} -e ${END_MSG_SIZE} -f 2 -g 1 -n ${NUM_ITER} -c 1 -w ${NUM_WARMUP}  
```

## Details

**What were the changes?**  
Add support for AMD AINIC RCCL default internal network plugin.

Provides out-of-the box experience for AMD AINIC and to enable optimal performance of the NIC w/o need for any additional plugin or configurations.

With this only RCCL library needs to be installed and no need for any additional packages/plugin.

**Additional Documentation:**  

## Approval Checklist


```
**RCCL test result with RCCL_AINIC_ROCE=1 where it picks up custom ROCM-IB plugin**
#### snip from log ######
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO RCCL_AINIC_ROCE set by environment to 1.                                                                                                                                                                                                                                                           
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO NET/Plugin: Could not find: libnccl-net.so.                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                       
[2025-12-16 03:25:23] smc300x-ccs-aus-gpufc55:452404:452471 [4] /apps/shared/karthik/repo/rccl_v2/build/release/hipify/src/misc/ionicdvsymbols.cc:48 NCCL WARN dlvsym loaded successfully for ionic_dv_qp_set_gda - version IONIC_1.0                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                       
[2025-12-16 03:25:23] smc300x-ccs-aus-gpufc55:452404:452471 [4] /apps/shared/karthik/repo/rccl_v2/build/release/hipify/src/misc/ionicdvsymbols.cc:49 NCCL WARN dlvsym loaded successfully for ionic_dv_pd_set_udma_mask - version IONIC_1.0                                                                                                            
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO Loaded dlvsym from libionic.so[.1]                                                                                                                                                                                                                                                                 
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO NCCL_SOCKET_IFNAME set by environment to ens50f0                                                                                                                                                                                                                                                   
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO NET/IB : Using [0]ionic_0:1/RoCE [1]ionic_1:1/RoCE [2]ionic_2:1/RoCE [3]ionic_3:1/RoCE [4]ionic_4:1/RoCE [5]ionic_5:1/RoCE [6]ionic_6:1/RoCE [7]ionic_7:1/RoCE [RO]; OOB ens50f0:10.235.200.207<0>                                                                                                 
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO NET/IB : AINIC RoCEv2 optimizations enabled: CTS Inline Data: Enabled; CTS Offload: Enabled; IB Use Inline: enabled; GDR Flush: disabled                                                                                                                                                           
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO Initialized NET plugin ROCM-IB                                                                                                                                                                                                                                                                     
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO Found /sys/kernel/mm/memory_peers/amdkfd/version                                                                                                                                                                                                                                                   
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO Assigned NET plugin ROCM-IB to comm                                                                                                                                                                                                                                                                
smc300x-ccs-aus-gpufc55:452404:452471 [4] NCCL INFO Using network ROCM-IB                                                        
############

+ /apps/shared/karthik/repo/ompi/install/bin/mpirun --np 16 -H DCCS-2205-NODE-49:8,DCCS-2205-NODE-51:8 --bind-to numa -x PATH=/apps/shared/karthik/repo/ompi/install/bin:/opt/rocm/bin:/home/karthik/.local/bin:/lib:/apps/shared:/apps/shared/karthik:/apps/shared/karthik/scripts:/apps/shared/karthik/ainic_scripts:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin -x LD_LIBRARY_PATH=/apps/shared/karthik/repo/ompi/install/lib:/opt/rocm/lib:/lib: --mca btl '^vader,openib' -mca btl_tcp_if_include ens50f0 -x NCCL_SOCKET_IFNAME=ens50f0 -x NCCL_IB_QPS_PER_CONNECTION=1 -x NCCL_TOPO_DUMP_FILE=/tmp/system_run2.txt -x NCCL_IB_TC=96 -x NCCL_IB_FIFO_TC=184 -x NCCL_DEBUG=VERSION -x RCCL_AINIC_ROCE=1 -x RCCL_GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING=0 -x NCCL_IGNORE_CPU_AFFINITY=1 -x IONIC_LOCKFREE=all -x RCCL_LL128_FORCE_ENABLE=1 /apps/shared/karthik/repo/rccl-tests/build/all_reduce_perf -b 1K -e 16G -f 2 -g 1 -n 20 -c 1 -w 5
# Collective test starting: all_reduce_perf
# nThread 1 nGpus 1 minBytes 1024 maxBytes 17179869184 step: 2(factor) warmup iters: 5 iters: 20 agg iters: 1 validation: 1 graph: 0
#
rccl-tests: Version develop:5272cd1
# Using devices
#  Rank  0 Group  0 Pid 864817 on smc300x-ccs-aus-gpu611e device  0 [0000:05:00] 
#  Rank  1 Group  0 Pid 864818 on smc300x-ccs-aus-gpu611e device  1 [0000:29:00] 
#  Rank  2 Group  0 Pid 864819 on smc300x-ccs-aus-gpu611e device  2 [0000:49:00] 
#  Rank  3 Group  0 Pid 864820 on smc300x-ccs-aus-gpu611e device  3 [0000:65:00] 
#  Rank  4 Group  0 Pid 864821 on smc300x-ccs-aus-gpu611e device  4 [0000:85:00] 
#  Rank  5 Group  0 Pid 864822 on smc300x-ccs-aus-gpu611e device  5 [0000:a9:00] 
#  Rank  6 Group  0 Pid 864823 on smc300x-ccs-aus-gpu611e device  6 [0000:c9:00] 
#  Rank  7 Group  0 Pid 864824 on smc300x-ccs-aus-gpu611e device  7 [0000:e5:00] 
#  Rank  8 Group  0 Pid 453835 on smc300x-ccs-aus-gpufc55 device  0 [0000:05:00] AMD Instinct MI300X
#  Rank  9 Group  0 Pid 453836 on smc300x-ccs-aus-gpufc55 device  1 [0000:29:00] AMD Instinct MI300X
#  Rank 10 Group  0 Pid 453837 on smc300x-ccs-aus-gpufc55 device  2 [0000:49:00] AMD Instinct MI300X
#  Rank 11 Group  0 Pid 453838 on smc300x-ccs-aus-gpufc55 device  3 [0000:65:00] AMD Instinct MI300X
#  Rank 12 Group  0 Pid 453839 on smc300x-ccs-aus-gpufc55 device  4 [0000:85:00] AMD Instinct MI300X
#  Rank 13 Group  0 Pid 453840 on smc300x-ccs-aus-gpufc55 device  5 [0000:a9:00] AMD Instinct MI300X
#  Rank 14 Group  0 Pid 453841 on smc300x-ccs-aus-gpufc55 device  6 [0000:c9:00] AMD Instinct MI300X
#  Rank 15 Group  0 Pid 453842 on smc300x-ccs-aus-gpufc55 device  7 [0000:e5:00] AMD Instinct MI300X
RCCL version : 2.27.7-ainic_support_in_rccl:915c6a3+
HIP version  : 7.2.25482-8728db54fa
ROCm version : 7.2.0.0-14-8728db54fa
Hostname     : smc300x-ccs-aus-gpu611e.prov.aus.ccs.cpe.ice.amd.com
Librccl path : /opt/rocm/lib/librccl.so.1
#
#                                                              out-of-place                       in-place          
#       size         count      type   redop    root     time   algbw   busbw #wrong     time   algbw   busbw #wrong                               
#        (B)    (elements)                               (us)  (GB/s)  (GB/s)            (us)  (GB/s)  (GB/s)                                      
        1024           256     float     sum      -1    42.70    0.02    0.04      0    35.89    0.03    0.05      0
        2048           512     float     sum      -1    36.53    0.06    0.11      0    38.41    0.05    0.10      0
        4096          1024     float     sum      -1    38.79    0.11    0.20      0    37.62    0.11    0.20      0
        8192          2048     float     sum      -1    43.35    0.19    0.35      0    38.06    0.22    0.40      0
       16384          4096     float     sum      -1    40.08    0.41    0.77      0    39.11    0.42    0.79      0
       32768          8192     float     sum      -1    41.47    0.79    1.48      0    42.35    0.77    1.45      0
       65536         16384     float     sum      -1    48.46    1.35    2.54      0    43.12    1.52    2.85      0
      131072         32768     float     sum      -1    47.04    2.79    5.22      0    45.23    2.90    5.43      0
      262144         65536     float     sum      -1    51.94    5.05    9.46      0    52.88    4.96    9.29      0
      524288        131072     float     sum      -1    63.93    8.20   15.38      0    68.54    7.65   14.34      0
     1048576        262144     float     sum      -1    62.21   16.85   31.60      0    64.17   16.34   30.64      0
     2097152        524288     float     sum      -1    71.58   29.30   54.94      0    72.23   29.04   54.44      0
     4194304       1048576     float     sum      -1    82.08   51.10   95.81      0    82.04   51.13   95.86      0
     8388608       2097152     float     sum      -1    186.2   45.04   84.46      0    189.2   44.33   83.12      0
    16777216       4194304     float     sum      -1    249.2   67.32  126.23      0    256.7   65.35  122.54      0
    33554432       8388608     float     sum      -1    390.5   85.94  161.13      0    395.5   84.85  159.09      0
    67108864      16777216     float     sum      -1    597.9  112.23  210.44      0    600.4  111.78  209.59      0
   134217728      33554432     float     sum      -1    918.9  146.06  273.87      0    904.1  148.45  278.35      0
   268435456      67108864     float     sum      -1   1456.0  184.36  345.68      0   1460.0  183.85  344.73      0
   536870912     134217728     float     sum      -1   2862.3  187.56  351.68      0   2860.0  187.72  351.97      0
  1073741824     268435456     float     sum      -1   5633.5  190.60  357.37      0   5628.0  190.78  357.72      0
  2147483648     536870912     float     sum      -1    11186  191.99  359.98      0    11199  191.75  359.53      0
  4294967296    1073741824     float     sum      -1    22410  191.65  359.35      0    22414  191.62  359.28      0
  8589934592    2147483648     float     sum      -1    44734  192.02  360.04      0    44670  192.30  360.56      0
 17179869184    4294967296     float     sum      -1    89235  192.52  360.98      0    89189  192.62  361.17      0
# Errors with asterisks indicate errors that have exceeded the maximum threshold.
# Out of bounds values : 0 OK
# Avg bus bandwidth    : 142.652 
#
# Collective test concluded: all_reduce_perf

```

```
**RCCL test result with RCCL_AINIC_ROCE=0 where it picks up vanilla/default NET-IB plugin**

#### snip from log ######
smc300x-ccs-aus-gpufc55:452949:453058 [4] NCCL INFO Assigned NET plugin IB to comm                                                                                                                                                                                                                                                                     
smc300x-ccs-aus-gpufc55:452949:453058 [4] NCCL INFO Using network IB                                                                                                                                                                                                                                                                                   
smc300x-ccs-aus-gpufc55:452952:453057 [7] NCCL INFO NET/IB : Using [0]ionic_0:1/RoCE [1]ionic_1:1/RoCE [2]ionic_2:1/RoCE [3]ionic_3:1/RoCE [4]ionic_4:1/RoCE [5]ionic_5:1/RoCE [6]ionic_6:1/RoCE [7]ionic_7:1/RoCE [RO]; OOB ens50f0:10.235.200.207<0>                                                                                                 
smc300x-ccs-aus-gpufc55:452952:453057 [7] NCCL INFO Initialized NET plugin IB 
##########

+ /apps/shared/karthik/repo/ompi/install/bin/mpirun --np 16 -H DCCS-2205-NODE-49:8,DCCS-2205-NODE-51:8 --bind-to numa -x PATH=/apps/shared/karthik/repo/ompi/install/bin:/opt/rocm/bin:/home/karthik/.local/bin:/lib:/apps/shared:/apps/shared/karthik:/apps/shared/karthik/scripts:/apps/shared/karthik/ainic_scripts:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin -x LD_LIBRARY_PATH=/apps/shared/karthik/repo/ompi/install/lib:/opt/rocm/lib:/lib: --mca btl '^vader,openib' -mca btl_tcp_if_include ens50f0 -x NCCL_SOCKET_IFNAME=ens50f0 -x NCCL_IB_QPS_PER_CONNECTION=1 -x NCCL_TOPO_DUMP_FILE=/tmp/system_run2.txt -x NCCL_IB_TC=96 -x NCCL_IB_FIFO_TC=184 -x NCCL_DEBUG=VERSION -x RCCL_AINIC_ROCE=0 -x RCCL_GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING=0 -x NCCL_IGNORE_CPU_AFFINITY=1 -x IONIC_LOCKFREE=all -x RCCL_LL128_FORCE_ENABLE=1 /apps/shared/karthik/repo/rccl-tests/build/all_reduce_perf -b 1K -e 16G -f 2 -g 1 -n 20 -c 1 -w 5
# Collective test starting: all_reduce_perf
# nThread 1 nGpus 1 minBytes 1024 maxBytes 17179869184 step: 2(factor) warmup iters: 5 iters: 20 agg iters: 1 validation: 1 graph: 0
#
rccl-tests: Version develop:5272cd1
# Using devices
#  Rank  0 Group  0 Pid 864363 on smc300x-ccs-aus-gpu611e device  0 [0000:05:00] 
#  Rank  1 Group  0 Pid 864364 on smc300x-ccs-aus-gpu611e device  1 [0000:29:00] 
#  Rank  2 Group  0 Pid 864365 on smc300x-ccs-aus-gpu611e device  2 [0000:49:00] 
#  Rank  3 Group  0 Pid 864366 on smc300x-ccs-aus-gpu611e device  3 [0000:65:00] 
#  Rank  4 Group  0 Pid 864367 on smc300x-ccs-aus-gpu611e device  4 [0000:85:00] 
#  Rank  5 Group  0 Pid 864368 on smc300x-ccs-aus-gpu611e device  5 [0000:a9:00] 
#  Rank  6 Group  0 Pid 864369 on smc300x-ccs-aus-gpu611e device  6 [0000:c9:00] 
#  Rank  7 Group  0 Pid 864370 on smc300x-ccs-aus-gpu611e device  7 [0000:e5:00] 
#  Rank  8 Group  0 Pid 453377 on smc300x-ccs-aus-gpufc55 device  0 [0000:05:00] AMD Instinct MI300X
#  Rank  9 Group  0 Pid 453378 on smc300x-ccs-aus-gpufc55 device  1 [0000:29:00] AMD Instinct MI300X
#  Rank 10 Group  0 Pid 453379 on smc300x-ccs-aus-gpufc55 device  2 [0000:49:00] AMD Instinct MI300X
#  Rank 11 Group  0 Pid 453380 on smc300x-ccs-aus-gpufc55 device  3 [0000:65:00] AMD Instinct MI300X
#  Rank 12 Group  0 Pid 453381 on smc300x-ccs-aus-gpufc55 device  4 [0000:85:00] AMD Instinct MI300X
#  Rank 13 Group  0 Pid 453382 on smc300x-ccs-aus-gpufc55 device  5 [0000:a9:00] AMD Instinct MI300X
#  Rank 14 Group  0 Pid 453383 on smc300x-ccs-aus-gpufc55 device  6 [0000:c9:00] AMD Instinct MI300X
#  Rank 15 Group  0 Pid 453384 on smc300x-ccs-aus-gpufc55 device  7 [0000:e5:00] AMD Instinct MI300X
RCCL version : 2.27.7-ainic_support_in_rccl:915c6a3+
HIP version  : 7.2.25482-8728db54fa
ROCm version : 7.2.0.0-14-8728db54fa
Hostname     : smc300x-ccs-aus-gpu611e.prov.aus.ccs.cpe.ice.amd.com
Librccl path : /opt/rocm/lib/librccl.so.1
#
#                                                              out-of-place                       in-place          
#       size         count      type   redop    root     time   algbw   busbw #wrong     time   algbw   busbw #wrong                               
#        (B)    (elements)                               (us)  (GB/s)  (GB/s)            (us)  (GB/s)  (GB/s)                                      
        1024           256     float     sum      -1    41.68    0.02    0.05      0    34.83    0.03    0.06      0
        2048           512     float     sum      -1    36.51    0.06    0.11      0    36.94    0.06    0.10      0
        4096          1024     float     sum      -1    37.82    0.11    0.20      0    36.78    0.11    0.21      0
        8192          2048     float     sum      -1    41.45    0.20    0.37      0    39.12    0.21    0.39      0
       16384          4096     float     sum      -1    39.89    0.41    0.77      0    38.28    0.43    0.80      0
       32768          8192     float     sum      -1    39.74    0.82    1.55      0    40.79    0.80    1.51      0
       65536         16384     float     sum      -1    46.02    1.42    2.67      0    41.63    1.57    2.95      0
      131072         32768     float     sum      -1    45.87    2.86    5.36      0    45.58    2.88    5.39      0
      262144         65536     float     sum      -1    53.44    4.91    9.20      0    53.40    4.91    9.20      0
      524288        131072     float     sum      -1    69.91    7.50   14.06      0    67.10    7.81   14.65      0
     1048576        262144     float     sum      -1    59.42   17.65   33.09      0    62.19   16.86   31.61      0
     2097152        524288     float     sum      -1    69.97   29.97   56.20      0    70.51   29.74   55.77      0
     4194304       1048576     float     sum      -1    80.77   51.93   97.36      0    82.09   51.09   95.80      0
     8388608       2097152     float     sum      -1    200.3   41.87   78.51      0    202.1   41.50   77.82      0
    16777216       4194304     float     sum      -1    275.5   60.90  114.18      0    284.3   59.02  110.66      0
    33554432       8388608     float     sum      -1    400.2   83.83  157.19      0    405.3   82.79  155.23      0
    67108864      16777216     float     sum      -1    608.1  110.36  206.92      0    612.4  109.59  205.48      0
   134217728      33554432     float     sum      -1   1094.7  122.61  229.89      0   1144.4  117.29  219.91      0
   268435456      67108864     float     sum      -1   1783.0  150.56  282.29      0   1566.4  171.37  321.32      0
   536870912     134217728     float     sum      -1   3951.0  135.88  254.78      0   3856.4  139.21  261.03      0
  1073741824     268435456     float     sum      -1   7532.0  142.56  267.30      0   7735.4  138.81  260.27      0
  2147483648     536870912     float     sum      -1    13622  157.65  295.59      0    13537  158.64  297.45      0
  4294967296    1073741824     float     sum      -1    26186  164.02  307.53      0    26987  159.15  298.41      0
  8589934592    2147483648     float     sum      -1    53869  159.46  298.99      0    54391  157.93  296.12      0
 17179869184    4294967296     float     sum      -1   108076  158.96  298.05      0   108251  158.70  297.57      0
# Errors with asterisks indicate errors that have exceeded the maximum threshold.
# Out of bounds values : 0 OK
# Avg bus bandwidth    : 120.638 
#
# Collective test concluded: all_reduce_perf

karthik@ainic16-headnode:/apps/shared/karthik/anp_within_rccl$ 
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,9 @@ endif()
 # Determine which GPU architectures to build for
 set(GPU_TARGETS "${DEFAULT_GPUS}" CACHE STRING "Target default GPUs if GPU_TARGETS is not defined.")
 
+# ROCM NetIB patch
+include(cmake/rocmIb.cmake)
+
 # Modify GPU architectures for Address Sanitizer builds by appending "xnack+"
 if (BUILD_ADDRESS_SANITIZER)
   SET(amdgpu_targets "")
@@ -571,6 +574,9 @@ set(SRC_FILES
   src/include/mlx5/mlx5dvcore.h
   src/include/mlx5/mlx5dvsymbols.h
   src/include/mlx5/mlx5dvwrap.h
+  src/include/ionic/ionicdvcore.h
+  src/include/ionic/ionicdvsymbols.h
+  src/include/ionic/ionicdvwrap.h
   src/include/msccl/msccl_lifecycle.h
   src/include/msccl/msccl_parser.h
   src/include/msccl/msccl_scheduler.h
@@ -647,6 +653,8 @@ set(SRC_FILES
   src/misc/ipcsocket.cc
   src/misc/mlx5dvsymbols.cc
   src/misc/mlx5dvwrap.cc
+  src/misc/ionicdvsymbols.cc
+  src/misc/ionicdvwrap.cc
   src/misc/npkit.cc
 # src/misc/nvmlwrap.cc
   src/misc/nvmlwrap_stub.cc
@@ -695,6 +703,7 @@ set(SRC_FILES
   src/transport/generic.cc
   src/transport/net.cc
   src/transport/net_ib.cc
+  src/transport/net_ib_rocm.cc
   src/transport/net_socket.cc
   src/transport/nvls.cc
   src/transport/p2p.cc
@@ -862,6 +871,7 @@ target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/device)
 target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/device/network/unpack)
 target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/include)
 target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/include/mlx5)
+target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/include/ionic)
 target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/src/include/plugin)
 target_include_directories(rccl PRIVATE ${HIPIFY_DIR}/gensrc)
 target_include_directories(rccl PRIVATE ${HSA_INCLUDE_PATH})
--- a/ext-src/rocm_netib.patch
+++ b/ext-src/rocm_netib.patch
@@ -1,0 +1,797 @@
+diff --git a/src/transport/net_ib.cc b/src/transport/net_ib.cc
+index 7af56a6c..5c3e3d46 100644
+--- a/src/transport/net_ib.cc
++++ b/src/transport/net_ib.cc
+@@ -28,6 +28,7 @@
+ 
+ #include "ibvwrap.h"
+ #include "mlx5/mlx5dvwrap.h"
++#include "ionic/ionicdvwrap.h"
+ #include "graph/xml.h"
+ 
+ #define MAXSUFFIXSIZE 16
+@@ -107,9 +108,31 @@ struct ncclIbMergedDev ncclIbMergedDevs[MAX_IB_VDEVS];
+ struct ncclIbDev ncclIbDevs[MAX_IB_DEVS];
+ pthread_mutex_t ncclIbLock = PTHREAD_MUTEX_INITIALIZER;
+ static int ncclIbRelaxedOrderingEnabled = 0;
++static bool rcclAinicRoce = 0;
++static bool rcclCtsInlineData = 0;
++static bool rcclCtsOffloadEnabled = 0;
++static bool ncclIbUseInline = 0;
++static int ncclIbGdrFlushDisable = 0;
++
++enum ncclIbChannelType {
++  ncclIbChannelTypeCts  = 0,
++  ncclIbChannelTypeData = 1,
++  ncclIbChannelTypeMax  = 2
++};
++
++struct ncclChannelToUd {
++    int channelId;
++    bool udId;
++    bool udAllocated;
++};
++
++static ncclChannelToUd nccl_channel_ud_map[MAXCHANNELS][ncclIbChannelTypeMax];
++static bool nccl_channel_last_ud[MAX_IB_DEVS][ncclIbChannelTypeMax];
+ 
+ #define NCCL_IB_LLSTR(ll) (((ll) == IBV_LINK_LAYER_INFINIBAND) ? "IB" : (((ll) == IBV_LINK_LAYER_ETHERNET) ? "RoCE" : "UNSPECIFIED"))
+ 
++#define NCCL_CTS_QP_SLOT_INVALID 0xFF
++
+ #define NCCL_IB_SL_DEFAULT 0
+ #define NCCL_IB_TC_DEFAULT 0
+ 
+@@ -131,6 +154,13 @@ NCCL_PARAM(IbEceEnable,"IB_ECE_ENABLE",1);
+ NCCL_PARAM(IbDataDirect,"IB_DATA_DIRECT",1);
+ NCCL_PARAM(IbQpsPerConn, "IB_QPS_PER_CONNECTION", 1);
+ RCCL_PARAM(IbQpsPerP2p, "IB_QPS_PER_P2P", 0);
++NCCL_PARAM(IbGdrFlushDisable, "GDR_FLUSH_DISABLE", 0);
++
++// AMD AINIC
++RCCL_PARAM(CtsInlineData, "CTS_INLINE_DATA", -1);
++RCCL_PARAM(CtsOffloadEnabled, "CTS_OFFLOAD_ENABLED", -1);
++
++extern int64_t rcclParamAinicRoce();
+ 
+ static ncclResult_t ncclIbStatsInit(struct ncclIbStats* stat) {
+   __atomic_store_n(&stat->fatalErrorCount, 0, __ATOMIC_RELAXED);
+@@ -630,6 +660,10 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t pr
+   static int shownIbHcaEnv = 0;
+   if(wrap_ibv_symbols() != ncclSuccess) { return ncclInternalError; }
+   if(wrap_mlx5dv_symbols() != ncclSuccess) { INFO(NCCL_NET, "NET/IB : Failed to open mlx5dv symbols. Advance features like CX-8 Direct-NIC will be disabled."); }
++  if(wrap_ionicdv_symbols() != ncclSuccess) {
++    WARN("NET/IB : Failed to open ionicdv symbols. Advance features like AINIC UD load balancing will be disabled.");
++    return ncclInternalError;
++  }
+ 
+   // Detect IB cards
+   int nIbDevs = 0;
+@@ -783,6 +817,24 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t pr
+     INFO(NCCL_INIT|NCCL_NET, "NET/IB : Using%s %s; OOB %s:%s", line, ncclIbRelaxedOrderingEnabled ? "[RO]" : "",
+           ncclIbIfName, ncclSocketToString(&ncclIbIfAddr, addrline));
+ 
++    ncclIbUseInline = ncclParamIbUseInline();
++    ncclIbGdrFlushDisable = ncclParamIbGdrFlushDisable();
++
++    rcclAinicRoce = ((rcclParamAinicRoce() == 1) ? true : false);
++    if (rcclAinicRoce) {
++      // for AINIC, these params are defaulted to enabled unless user forces it to disable(0).
++      rcclCtsInlineData = ((rcclParamCtsInlineData() == 0) ? false : true);
++      rcclCtsOffloadEnabled = ((rcclParamCtsOffloadEnabled() == 0) ? false : true);
++      // for AINIC IbUseInline is enabled by default always
++      ncclIbUseInline = true;
++      // for AINIC GDR flush is disabled by default
++      ncclIbGdrFlushDisable = 1;
++
++      INFO(NCCL_INIT|NCCL_NET, "NET/IB : AINIC RoCEv2 optimizations enabled: CTS Inline Data: %s; CTS Offload: %s; "
++           "IB Use Inline: enabled; GDR Flush: disabled", rcclCtsInlineData ? "Enabled": "Disabled",
++           rcclCtsOffloadEnabled ? "Enabled": "Disabled");
++    }
++
+     pthread_mutex_unlock(&ncclIbLock);
+   }
+ exit:
+@@ -1112,6 +1164,8 @@ struct ncclIbListenComm {
+   struct ncclIbCommStage stage;
+ };
+ 
++#define MAX_INLINE_DATA_SIZE 24
++
+ struct alignas(64) ncclIbSendFifo {
+   uint64_t addr;
+   uint64_t size;
+@@ -1122,10 +1176,21 @@ struct alignas(64) ncclIbSendFifo {
+   char padding[16];
+ };
+ 
++struct alignas(32) ncclIbSendFifoCtsInline {
++  uint64_t addr;
++  uint32_t rkeys[1];
++  int size;
++  uint8_t nreqs;
++  uint16_t tag;
++  uint32_t idx;
++  char padding[9];
++} __attribute__((packed));
++
+ struct ncclIbQp {
+   struct ibv_qp* qp;
+   int devIndex;
+   int remDevIdx;
++  int8_t ctsQpSlot;
+ };
+ 
+ struct ncclIbRemSizesFifo {
+@@ -1172,6 +1237,7 @@ struct ncclIbSendComm {
+   struct ncclIbNetCommBase base;
+   // Start with fifo and ibv structs as they have alignment restrictions
+   struct ncclIbSendFifo fifo[MAX_REQUESTS][NCCL_NET_IB_MAX_RECVS];
++  struct ncclIbSendFifoCtsInline fifo_inline[MAX_REQUESTS][NCCL_NET_IB_MAX_RECVS];
+   struct ibv_sge sges[NCCL_NET_IB_MAX_RECVS];
+   struct ibv_send_wr wrs[NCCL_NET_IB_MAX_RECVS + 1];
+   // Each dev correlates to a mergedIbDev
+@@ -1187,6 +1253,7 @@ struct ncclIbSendComm {
+ static_assert((sizeof(struct ncclIbNetCommBase) % 32) == 0, "ncclIbNetCommBase size must be 32-byte multiple to ensure fifo is at proper offset");
+ static_assert((offsetof(struct ncclIbSendComm, fifo) % 32) == 0, "ncclIbSendComm fifo must be 32-byte aligned");
+ static_assert((sizeof(struct ncclIbSendFifo) % 32) == 0, "ncclIbSendFifo element size must be 32-byte multiples");
++static_assert((sizeof(struct ncclIbSendFifoCtsInline) % 32) == 0, "ncclIbSendFifoCtsInline element size must be 32-byte multiples");
+ static_assert((offsetof(struct ncclIbSendComm, sges) % 32) == 0, "sges must be 32-byte aligned");
+ static_assert((offsetof(struct ncclIbSendComm, wrs) % 32) == 0, "wrs must be 32-byte aligned");
+ 
+@@ -1201,6 +1268,7 @@ struct ncclIbGpuFlush {
+ 
+ struct ncclIbRemFifo {
+   struct ncclIbSendFifo elems[MAX_REQUESTS][NCCL_NET_IB_MAX_RECVS];
++  struct ncclIbSendFifoCtsInline elems_cts_inline[MAX_REQUESTS][NCCL_NET_IB_MAX_RECVS];
+   uint64_t fifoTail;
+   uint64_t addr;
+   uint32_t flags;
+@@ -1265,20 +1333,59 @@ returning:
+   return res;
+ }
+ 
+-ncclResult_t ncclIbCreateQp(uint8_t ib_port, struct ncclIbNetCommDevBase* base, int access_flags, void* qp_context, struct ncclIbQp* qp) {
++ncclResult_t ncclIbCreateQp(uint8_t ib_port, struct ncclIbNetCommDevBase* base,
++                            int access_flags, void* qp_context, struct ncclIbQp* qp,
++                            int channel_id, bool data_qp, int8_t cts_qp_slot) {
+   struct ibv_qp_init_attr qpInitAttr;
++  enum ncclIbChannelType channel_type = (data_qp ? ncclIbChannelTypeData : ncclIbChannelTypeCts);
+   memset(&qpInitAttr, 0, sizeof(struct ibv_qp_init_attr));
+   qpInitAttr.qp_context = qp_context;
+   qpInitAttr.send_cq = base->cq;
+   qpInitAttr.recv_cq = base->cq;
+   qpInitAttr.qp_type = IBV_QPT_RC;
++
++  if (rcclAinicRoce) {
++    if (!nccl_channel_ud_map[channel_id][channel_type].udAllocated) {
++      bool lud = nccl_channel_last_ud[base->ibDevN][channel_type];
++      nccl_channel_ud_map[channel_id][channel_type].udId = lud;
++      nccl_channel_ud_map[channel_id][channel_type].udAllocated = true;
++      nccl_channel_last_ud[base->ibDevN][channel_type] =
++          !(nccl_channel_last_ud[base->ibDevN][channel_type]);
++    }
++    if (nccl_channel_ud_map[channel_id][channel_type].udId) {
++        wrap_ionicdv_pd_set_udma_mask(base->pd, IONIC_UDMA_MASK_HIGH);
++    } else {
++        wrap_ionicdv_pd_set_udma_mask(base->pd, IONIC_UDMA_MASK_LOW);
++    }
++    qpInitAttr.sq_sig_all |= (1 << 16);
++    if (data_qp) {
++      qpInitAttr.sq_sig_all |= (1 << 17);
++    } else {
++      qpInitAttr.sq_sig_all &= (~(1 << 17));
++    }
++    qpInitAttr.sq_sig_all |= (1 << 18);
++
++    if (rcclCtsOffloadEnabled) {
++      qpInitAttr.sq_sig_all |= (1 << 19);
++    } else {
++      qpInitAttr.sq_sig_all &= (~(1 << 19));
++    }
++  }
++
+   // We might send 2 messages per send (RDMA and RDMA_WITH_IMM)
+   qpInitAttr.cap.max_send_wr = 2*MAX_REQUESTS;
+   qpInitAttr.cap.max_recv_wr = MAX_REQUESTS;
+   qpInitAttr.cap.max_send_sge = 1;
+   qpInitAttr.cap.max_recv_sge = 1;
+-  qpInitAttr.cap.max_inline_data = ncclParamIbUseInline() ? sizeof(struct ncclIbSendFifo) : 0;
++  if (rcclCtsInlineData) {
++    qpInitAttr.cap.max_inline_data = MAX_INLINE_DATA_SIZE;
++  } else {
++    qpInitAttr.cap.max_inline_data = ncclIbUseInline ? sizeof(struct ncclIbSendFifo) : 0;
++  }
+   NCCLCHECK(wrap_ibv_create_qp(&qp->qp, base->pd, &qpInitAttr));
++  if (rcclAinicRoce) {
++    NCCLCHECK(wrap_ionicdv_qp_set_gda(qp->qp, false, true));
++  }
+   struct ibv_qp_attr qpAttr;
+   memset(&qpAttr, 0, sizeof(struct ibv_qp_attr));
+   qpAttr.qp_state = IBV_QPS_INIT;
+@@ -1288,6 +1395,9 @@ ncclResult_t ncclIbCreateQp(uint8_t ib_port, struct ncclIbNetCommDevBase* base,
+   NCCLCHECK(wrap_ibv_modify_qp(qp->qp, &qpAttr, IBV_QP_STATE | IBV_QP_PKEY_INDEX | IBV_QP_PORT | IBV_QP_ACCESS_FLAGS));
+   TRACE(NCCL_NET, "NET/IB : ncclIbCreateQp port=%d dev=%d devName=%s ndevs=%d nmdevs=%d qpn=%u pkey=%u pd=%p",
+     ib_port, base->ibDevN, ncclIbDevs[base->ibDevN].devName, ncclNIbDevs, ncclNMergedIbDevs, qp->qp->qp_num, qpAttr.pkey_index, base->pd);
++  if (rcclAinicRoce) {
++    qp->ctsQpSlot = cts_qp_slot;
++  }
+   return ncclSuccess;
+ }
+ 
+@@ -1371,7 +1481,7 @@ fail:
+   goto exit;
+ }
+ 
+-ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHandle, void** sendComm, ncclNetDeviceHandle_t** /*sendDevComm*/) {
++ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHandle, void** sendComm, ncclNetDeviceHandle_t** sendDevComm) {
+   ncclResult_t ret = ncclSuccess;
+   struct ncclIbHandle* handle = (struct ncclIbHandle*) opaqueHandle;
+   struct ncclIbCommStage* stage = &handle->stage;
+@@ -1379,8 +1489,13 @@ ncclResult_t ncclIbConnect(int dev, ncclNetCommConfig_t* config, void* opaqueHan
+   int ready;
+   uint8_t link_layer = IBV_LINK_LAYER_UNSPECIFIED;
+   int isP2p = 0; 
++  int channel_id = 0;
+   *sendComm = NULL;
+ 
++  if (rcclAinicRoce) {
++    channel_id = ((ncclNet_ctxt_t *)sendDevComm)->chId;
++  }
++
+   if (stage->state == ncclIbCommStateConnect)      goto ib_connect_check;
+   if (stage->state == ncclIbCommStateSendDevList)  goto ib_send_dev_list;
+   if (stage->state == ncclIbCommStateRecvDevList)  goto ib_recv_dev_list;
+@@ -1461,7 +1576,7 @@ ib_recv_dev_list:
+   for (int q = 0; q < comm->base.nqps; q++) {
+     ncclIbSendCommDev* commDev = comm->devs + devIndex;
+     ncclIbDev* ibDev = ncclIbDevs + commDev->base.ibDevN;
+-    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &commDev->base, IBV_ACCESS_REMOTE_WRITE, &comm->base.stats, comm->base.qps + q), ret, fail);
++    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &commDev->base, IBV_ACCESS_REMOTE_WRITE, &comm->base.stats, comm->base.qps + q, channel_id, true, NCCL_CTS_QP_SLOT_INVALID), ret, fail);
+     comm->base.qps[q].devIndex = devIndex;
+     meta.qpInfo[q].qpn      = comm->base.qps[q].qp->qp_num;
+     meta.qpInfo[q].devIndex = comm->base.qps[q].devIndex;
+@@ -1486,7 +1601,11 @@ ib_recv_dev_list:
+     devInfo->lid           = ibDev->portAttr.lid;
+     devInfo->ibv_dev_index = commDev->base.ibDevN;
+     // Prepare my fifo
+-    NCCLCHECKGOTO(wrap_ibv_reg_mr(&commDev->fifoMr, commDev->base.pd, comm->fifo, sizeof(struct ncclIbSendFifo)*MAX_REQUESTS*NCCL_NET_IB_MAX_RECVS, IBV_ACCESS_LOCAL_WRITE|IBV_ACCESS_REMOTE_WRITE|IBV_ACCESS_REMOTE_READ), ret, fail);
++    if (rcclCtsInlineData) {
++      NCCLCHECKGOTO(wrap_ibv_reg_mr(&commDev->fifoMr, commDev->base.pd, comm->fifo_inline, sizeof(struct ncclIbSendFifoCtsInline)*MAX_REQUESTS*NCCL_NET_IB_MAX_RECVS, IBV_ACCESS_LOCAL_WRITE|IBV_ACCESS_REMOTE_WRITE|IBV_ACCESS_REMOTE_READ), ret, fail);
++    } else {
++      NCCLCHECKGOTO(wrap_ibv_reg_mr(&commDev->fifoMr, commDev->base.pd, comm->fifo, sizeof(struct ncclIbSendFifo)*MAX_REQUESTS*NCCL_NET_IB_MAX_RECVS, IBV_ACCESS_LOCAL_WRITE|IBV_ACCESS_REMOTE_WRITE|IBV_ACCESS_REMOTE_READ), ret, fail);
++    }
+     devInfo->fifoRkey = commDev->fifoMr->rkey;
+ 
+     // Pack local GID info
+@@ -1528,7 +1647,11 @@ ib_recv_dev_list:
+       return ncclInternalError;
+     }
+   }
+-  meta.fifoAddr = (uint64_t)comm->fifo;
++  if (rcclCtsInlineData) {
++    meta.fifoAddr = (uint64_t)comm->fifo_inline;
++  } else {
++    meta.fifoAddr = (uint64_t)comm->fifo;
++  }
+   meta.sl = (ncclParamIbSl() != -1) ? ncclParamIbSl() : (config && config->trafficClass != NCCL_NET_TRAFFIC_CLASS_UNDEF) ? config->trafficClass : NCCL_IB_SL_DEFAULT;
+   meta.tc = (ncclParamIbTc() != -1) ? ncclParamIbTc() : (config && config->trafficClass != NCCL_NET_TRAFFIC_CLASS_UNDEF) ? config->trafficClass : NCCL_IB_TC_DEFAULT;
+   strncpy(meta.devName, mergedDev->devName, MAX_MERGED_DEV_NAME);
+@@ -1673,18 +1796,22 @@ ncclResult_t ncclIbCheckVProps(ncclNetVDeviceProps_t* vProps1, ncclNetVDevicePro
+   return ncclSuccess;
+ }
+ 
+-NCCL_PARAM(IbGdrFlushDisable, "GDR_FLUSH_DISABLE", 0);
+ RCCL_PARAM(IbGdrFlushGpuMemNoRelaxedOrdering, "GDR_FLUSH_GPU_MEM_NO_RELAXED_ORDERING", 1);
+ 
+-ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle_t** /*recvDevComm*/) {
++ncclResult_t ncclIbAccept(void* listenComm, void** recvComm, ncclNetDeviceHandle_t** recvDevComm) {
+   ncclResult_t ret = ncclSuccess;
+   struct ncclIbListenComm* lComm = (struct ncclIbListenComm*)listenComm;
+   struct ncclIbCommStage* stage = &lComm->stage;
+   struct ncclIbRecvComm* rComm = (struct ncclIbRecvComm*)stage->comm;
+   int ready;
+   int link_layer = IBV_LINK_LAYER_UNSPECIFIED;
++  int channel_id = 0;
+   *recvComm = NULL;
+ 
++  if (rcclAinicRoce) {
++    channel_id = ((ncclNet_ctxt_t *) recvDevComm)->chId;
++  }
++
+   if (stage->state == ncclIbCommStateAccept)   goto ib_accept_check;
+   if (stage->state == ncclIbCommStateRecvDevList) goto ib_recv_dev_list;
+   if (stage->state == ncclIbCommStateSendDevList) goto ib_send_dev_list;
+@@ -1814,7 +1941,7 @@ ib_recv:
+     // Local ibDevN
+     ibDevN = rComm->devs[devIndex].base.ibDevN;
+     ibDev = ncclIbDevs + ibDevN;
+-    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &rCommDev->base, IBV_ACCESS_REMOTE_WRITE, &rComm->base.stats, qp), ret, fail);
++    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &rCommDev->base, IBV_ACCESS_REMOTE_WRITE, &rComm->base.stats, qp, channel_id, false, q), ret, fail);
+     qp->devIndex = devIndex;
+     devIndex = (devIndex + 1) % rComm->base.vProps.ndevs;
+ 
+@@ -1840,16 +1967,22 @@ ib_recv:
+ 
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -631,6 +634,8 @@ set(SRC_FILES
   src/misc/ipcsocket.cc
   src/misc/mlx5dvsymbols.cc
   src/misc/mlx5dvwrap.cc
+  src/misc/ionicdvsymbols.cc
+  src/misc/ionicdvwrap.cc
   src/misc/npkit.cc
```
nileshnegi, line 658: can you please wrap these src files and the header files in a CMake var that gates support for AINIC plugin?
just so users can build with/without ANP, when needed.
like we have for MSCCL and MSCCLPP: https://github.com/ROCm/rccl/blob/develop/CMakeLists.txt#L698-L712

the default for this var can be set to ON, if that's what we decide.

karthikarum: @nileshnegi  
I shall look into wrapping these under CMake var. 

But just to be clearer, though AINIC support is built by default, it will get exercised with the runtime param  RCCL_AINIC.

I also see there is MLX5 support present outside of CMake var.
```  
src/misc/mlx5dvsymbols.cc
src/misc/mlx5dvwrap.cc
```

src/transport/net_ib.cc
mustafabar, line 1: Is there a strong reason why this isn’t maintained as a separate file from net_ib.cc?
The rcclAinic flag is a boolean that can be checked at a higher level, and the symbols can be dynamically loaded to point to the AINIC plugin.
Keeping AINIC support within net_ib.cc will make future syncs even more challenging and may introduce bugs to other NICs. 
The ANP plugin can be synced with net_ib.cc either after or during the sync process to take advantage of new features as needed.

karthikarum: This gives customers of AINIC a seamless experience for integration with applications using RCCL.
With this part of RCCL default internal IB plugin, installing ROCM will get customers going without have to deal with any additional plugin or other packages to leverage the optimal performance out of AINIC.

I do see [PR 1880 ](https://github.com/ROCm/rccl/pull/1880) that adds support for CX8 to internal IB plugin. I hope this change is in the same lines to help leverage the basic functionality of a given NIC w/o need for additional plugin; which can be reserved for additional/advanced capabilities offered by the NIC.

src/transport/net_ib.cc
```diff
@@ -783,6 +817,19 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t pr
...
+    ncclIbGdrFlushDisable = ncclParamIbGdrFlushDisable();
+
+    rcclAinic = ((rcclParamAinic() == 1) ? true : false);
+    if (rcclAinic) {
+      rcclCtsInlineData = ((rcclParamCtsInlineData() == -1) ? true : false);
+      rcclCtsOffloadEnabled = ((rcclParamCtsOffloadEnabled() == -1) ? true : false);
+      // for AINIC to disable Inline data, set env NCCL_IB_USE_INLINE=-1
+      ncclIbUseInline = ((ncclParamIbUseInline() == -1) ? false : true);
+      // for AINIC GDR flush is disabled by default
+      ncclIbGdrFlushDisable = 1;
```
alexander-sannikov: It worth to have such kind of debug output to confirm that AINIC support is enabled, and identify set of optimizations has been used. Otherwise it is hard to check actual settings used during a run.

```suggestion
      ncclIbGdrFlushDisable = 1;
      INFO(NCCL_INIT|NCCL_NET, "NET/IB : AINIC optimizations enabled: CTS Inline Data: %s; CTS Offload %s;  GDR Flush: disabled",
                rcclCtsInlineData ? "Enabled": "Disabled",
                rcclCtsOffloadEnabled ? "Enabled": "Disabled");                
```

src/transport/net_ib.cc
ddurnov: @mustafabar How frequent are the changes on net_ib level? The out of the box experience of RCCL users with AINIC is the key priority here. If you have major concern and you saw that integration with NCCL's net_ib is expensive, then would be great if we could sync up here. As an undesirable alternative from my perspective we may have second net_ib plugin file which would be part of core RCCL installation (AINIC has to be part of RCCL distribution and not require any additional modules/steps/packages) but ideally it has to be single plug-in which would cover all IB devices so we would be able to integrate optimizations which would cover all NICs rather than propagate them to multiple plug-ins going forward (e.g. QP load balancing, spraying and other future features)

mustafabar: My suggestion does not alter the out-of-the-box experience for RCCL users with AINIC. It is simply about deciding where to place the AINIC-specific changes within the RCCL codebase (e.g., net_ib.cc vs <ainic_file>.cc). I am not proposing that the AINIC file handle all NICs; rather, it should focus exclusively on AMD AINIC, containing the latest and most optimized implementation in a source file that we fully control. These symbols should only be loaded when AINIC is used while remaining part of the RCCL source. Historically, net_ib.cc undergoes moderate to significant changes with each sync. @wenkaidu may have views here as well. I’m not certain this qualifies as a major concern, but it is something to keep in mind. Please feel free to DM me if you need clarification.

src/transport/net_ib.cc
```diff
@@ -783,6 +817,19 @@ ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t pr
...
 
+    ncclIbUseInline = ncclParamIbUseInline();
+    ncclIbGdrFlushDisable = ncclParamIbGdrFlushDisable();
+
+    rcclAinic = ((rcclParamAinic() == 1) ? true : false);
+    if (rcclAinic) {
+      rcclCtsInlineData = ((rcclParamCtsInlineData() == -1) ? true : false);
+      rcclCtsOffloadEnabled = ((rcclParamCtsOffloadEnabled() == -1) ? true : false);
+      // for AINIC to disable Inline data, set env NCCL_IB_USE_INLINE=-1
+      ncclIbUseInline = ((ncclParamIbUseInline() == -1) ? false : true);
```
alexander-sannikov: rcclParamCtsInlineData et al. returns -1 if variable wasn't set.
Thus this check will ignore value of env. variable if one is set.
Should we just set default values to "1", and simplify code below?
Also, why there is a difference in NCCL_IB_USE_INLINE? It would be a hard to describe this to users that -1 is a "false" for AINIC only, but 0 is "true". and at the same time situation is opposite for anything else.

```suggestion
      rcclCtsInlineData = rcclParamCtsInlineData();
      rcclCtsOffloadEnabled = rcclParamCtsOffloadEnabled();
```

src/transport/net_ib.cc
```diff
@@ -131,6 +154,13 @@ NCCL_PARAM(IbEceEnable,"IB_ECE_ENABLE",1);
 NCCL_PARAM(IbDataDirect,"IB_DATA_DIRECT",1);
 NCCL_PARAM(IbQpsPerConn, "IB_QPS_PER_CONNECTION", 1);
 RCCL_PARAM(IbQpsPerP2p, "IB_QPS_PER_P2P", 0);
+NCCL_PARAM(IbGdrFlushDisable, "GDR_FLUSH_DISABLE", 0);
+
+// AMD AINIC
+RCCL_PARAM(CtsInlineData, "CTS_INLINE_DATA", -1);
+RCCL_PARAM(CtsOffloadEnabled, "CTS_OFFLOAD_ENABLED", -1);
```
alexander-sannikov:
```suggestion
RCCL_PARAM(CtsInlineData, "CTS_INLINE_DATA", 1);
RCCL_PARAM(CtsOffloadEnabled, "CTS_OFFLOAD_ENABLED", 1);
```
Please refer my comment below.

src/transport/net_ib.cc
ddurnov: @mustafabar Thanks. Yes, lets sync up tomorrow. I see that separate file could be an option but at the same time I see generic optimizations which could be done everywhere so validation cycle wise it could be more optimal to keep them together. If we are going to provide further serious customizations to the net_ib for AI NIC specifically and we want to have it as a differentiation item, then yes, it is better to put it into separate file on the same level inside RCCL and make this plug-in default for AI NIC (some RCCL high level AI NIC detection logic which would force the new net ib plug-in enabling by default). We are going to pay for it one way or another. The question is how much fallback we want to have. If the strategy is to keep core part of NCCL inside RCCL as close as possible so we won't have to be limited timewise (faster merge time with new NCCL) in case of potential collision with the custom AI NIC part, then I see your position here, but technically AI NIC is one of the key targets and lack of AI NIC will be limitation for RCCL release anyway. Is there anyone among our customers who customize/modify vanilla net ib for their own needs? (if answer is yes, then no doubt we have to go the route with separate file)

mustafabar: @ddurnov - Sure, we can sync up tomorrow. Your point about there being a cost either way is valid—considering the trade-offs, including any overhead from moving code and dynamic symbol loading when AINIC is detected, will help us reach a conclusion. Regarding customization of vanilla net IB, I don’t have a definitive answer beyond ‘it’s possible.’

karthikarum: @ddurnov @mustafabar if timing works out, I shall as well join the discussion. Let me know when you meet

src/transport/net_ib.cc
karthikarum: this will be taken care of and planned to handle differently.

src/transport/net_ib.cc
karthikarum: sure, will add the same

src/transport/net_ib.cc
karthikarum: this will be taken care of and planned to handle differently.

CMakeLists.txt
karthikarum: updated code per latest discussion with @wenkaidu

ext-src/ainic.patch
```diff
@@ -0,0 +1,51 @@
+diff --git a/src/include/net.h b/src/include/net.h
+index 3c8567fc..684365bf 100644
+--- a/src/include/net.h
++++ b/src/include/net.h
```
alexander-sannikov: This changes are minimal, does it makes sense to keep it as an  extra patch?

ext-src/ainic.patch
```diff
@@ -0,0 +1,51 @@
...
+@@ -25,4 +25,9 @@ extern ncclNet_t ncclNetSocket;
+
+ extern ncclResult_t rcclNetP2pPolicy(void* handle, int isP2p);
+
++#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
++extern ncclNet_t anpNetIb;
++extern ncclResult_t rcclAnpNetP2pPolicy(void* handle, int isP2p);
```
alexander-sannikov: `rcclAnpNetP2pPolicy` is declared, but not called from [transport/net.cc](https://github.com/ROCm/rccl/blob/5384a8abb2fefe72a1a3358a63e3cab1c210f2ec/src/transport/net.cc#L780C31-L780C40) in the same way as it has been done for `rcclNetP2pPolicy`

ext-src/ainic_netib.patch
```diff
@@ -0,0 +1,1030 @@
...
+-  ncclResult_t res = ncclIbMakeVDeviceInternal(d, props);
++  ncclResult_t res = anpNetMakeVDeviceInternal(d, props);
+   pthread_mutex_unlock(&ncclIbLock);
+   return res;
+ }
+ 
+ static ncclProfilerCallback_t ncclProfilerFunction;
+ 
+-ncclResult_t ncclIbInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction) {
++ncclResult_t anpNetInit(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction) {
```
alexander-sannikov: I may say right now that this patch won't work after NCCL 2.28 merge - 
`ncclIbInit`, `ncclIbListen`  and `ncclIbConnect` has changed their signatures.
If we follow this way, should sed be used for renaming this symbols? Otherwise we introduce extra problems for ongoing sync process.

ext-src/ainic.patch
karthikarum: Though the changes in `src/include/net.h` are minimal, for the sake of consistency I'm not having the definitions that are not defined in vanilla net-ib.

I'm open to add the changes (in net.h) to the file itself instead of patch.

ext-src/ainic.patch
karthikarum: In the first version of this I'm limiting the scope this to rolling only the ANP plugin changes to RCCL.
I shall add the call to P2pPolicy validation in next commit.

ext-src/ainic_netib.patch
karthikarum: That's right, I was thinking in similar lines after seeing the amount of changes in the patch.
I shall make update to move those ncclNet function pointers name changes to sed.

ext-src/ainic_netib.patch
```diff
@@ -0,0 +1,1030 @@
...
++
+     pthread_mutex_unlock(&ncclIbLock);
+   }
+ exit:
+@@ -793,7 +845,7 @@ fail:
+   goto exit;
+ }
+ 
+-ncclResult_t ncclIbDevices(int* ndev) {
++ncclResult_t anpNetDevices(int* ndev) {
```
wenkaidu: Will it make patch easier if just keeping original function names, i.e. ncclIbDevices, etc?

alexander-sannikov: There is a problem in original NCCL code that symbols which suppose to be static are globally visible. That's why all symbols have to be renamed as a part of a patch or using sed, during build process. Main problem that any new symbol added into net_ib in future and not supposed to be global have to be renamed as well to prevent symbols conflict.

ext-src/ainic.patch
wenkaidu: Agree with @alexander-sannikov. I also think direct change to net.h
 is better.

ext-src/ainic.patch
```diff
@@ -0,0 +1,51 @@
...
+   }
+
++#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
++  if ((rcclParamAinicRoce() == 1)) {
++    netPluginLibs[pluginCounter].ncclNet = &anpNetIb;
++    netPluginLibs[pluginCounter++].ncclNetPluginState = ncclNetPluginStateInitReady;
++  }
++#endif
+   // Add 2 internal ib and socket plugins
+   netPluginLibs[pluginCounter].ncclNet = &ncclNetIb;
```
wenkaidu: what is the reason to have both AINIC plugin and internal IB plugin at same time?

src/transport/net_ib.cc
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

src/transport/net_ib.cc
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

src/transport/net_ib.cc
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

ext-src/ainic.patch
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

ext-src/ainic.patch
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

ext-src/ainic_netib.patch
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

ext-src/ainic_netib.patch
karthikarum: refactored code per discussion on 12/11/2025 and pushed updated to PR

ext-src/ainic.patch
karthikarum: check for Ainic will be added to load the plugin. else will be default to avoid any other cases or when run without AinicRoce flag.

src/transport/net_ib_rocm.cc
```diff
@@ -0,0 +1,3006 @@
...
+NCCL_PARAM(RocmIbSl, "IB_SL", -1);
+NCCL_PARAM(RocmIbTc, "IB_TC", -1);
+NCCL_PARAM(RocmIbArThreshold, "IB_AR_THRESHOLD", 8192);
+NCCL_PARAM(RocmIbPciRelaxedOrdering, "IB_PCI_RELAXED_ORDERING", 2);
+NCCL_PARAM(RocmIbAdaptiveRouting, "IB_ADAPTIVE_ROUTING", -2);
+NCCL_PARAM(RocmIbFifoTc, "IB_FIFO_TC", -1);
+NCCL_PARAM(RocmIbAsyncEvents,"IB_RETURN_ASYNC_EVENTS",1);
+NCCL_PARAM(RocmIbEceEnable,"IB_ECE_ENABLE",1);
+NCCL_PARAM(RocmIbDataDirect,"IB_DATA_DIRECT",1);
+NCCL_PARAM(RocmIbQpsPerConn, "IB_QPS_PER_CONNECTION", 1);
```
vijasrin, line 155: should we leave the existing params as-is ?
For ex:
NCCL_IB_FIFO_TC will become NCCL_ROCM_IB_FIFO_TC

This means customers will have to modify the existing RCCL scripts

src/transport/net_ib_rocm.cc
alexander-sannikov, line 1: If this file only used for reference, should we rename it accordingly, e.g. `reference-net_ib_rocm.cc`? 
It may confuse if someone will modify this file in attempt to introduce some changes in `net_ib_rocm.cc`.

src/transport/net_ib.cc
karthikarum: comment addressed in latest commit.

src/transport/net_ib_rocm.cc
karthikarum: as discussed in person, the MACRO ENV name would still remain as before. Only the 1st argument passed to NCCL_PARAM() macro will now be having prefix RocmIb which is only used to defined the ncclParam<api-name>.
So this shouldn't affect the way how env is passed from the cmdline

src/transport/net_ib_rocm.cc
karthikarum: this is the same auto-generated file (thru the cmake) that is compiled with librccl.so. So we can't rename this file.

If it helps, I shall add header (warning) at top of file as part of the patch generation to indicate that this file is auto-generated file and shouldn't be edited manually.

alexander-sannikov (approved):

src/misc/ionicdvsymbols.cc
```diff
@@ -0,0 +1,60 @@
...
+  LOAD_SYM(ionicdvhandle, "ionic_dv_pd_set_udma_mask", ionicdvSymbols->ionicdv_internal_pd_set_udma_mask);
+  INFO(NCCL_INIT, "Loaded dlvsym from libionic.so[.1]");
+
+  return ncclSuccess;
+
+teardown:
+  ionicdvSymbols->ionicdv_internal_qp_set_gda = NULL;
+  ionicdvSymbols->ionicdv_internal_pd_set_udma_mask = NULL;
+
+  if (ionicdvSymbols != NULL) dlclose(ionicdvSymbols);
```
mustafabar: Need to close handle not symbols.. I think this may seg-fault.

src/include/net.h
```diff
@@ -25,4 +25,9 @@ extern ncclNet_t ncclNetSocket;
 
 extern ncclResult_t rcclNetP2pPolicy(void* handle, int isP2p);
 
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
+extern ncclNet_t rocmNetIb;
+extern ncclResult_t rcclRocmNetP2pPolicy(void* handle, int isP2p);
```
mustafabar, line 30: I don't see this being called anywhere. It is needed to support P2P QP capping.

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
+diff --git a/src/transport/net_ib.cc b/src/transport/net_ib.cc
+index 7af56a6c..5c3e3d46 100644
```
mustafabar, line 2: Do we need this patch given that the full source is there?

mustafabar (approved):

wenkaidu (approved):

vijasrin (approved):
lgtm

src/misc/ionicdvsymbols.cc
karthikarum: That's a miss on my end ! good catch @mustafabar. 
addressed and pushing the update to PR.

src/include/net.h
karthikarum: This will be called from the generated (from the patch file) net_ib_rocm.cc.

ext-src/rocm_netib.patch
karthikarum: This file net_ib_rocm.cc is generated (from the patch file) and doing away with the patch defeats the purpose of keeping upto date with NCCL sync of net_ib.cc

wenkaidu (approved):

mustafabar (approved):

alexander-sannikov (approved):

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
++      nccl_channel_ud_map[channel_id][channel_type].udAllocated = true;
++      nccl_channel_last_ud[base->ibDevN][channel_type] =
++          !(nccl_channel_last_ud[base->ibDevN][channel_type]);
++    }
++    if (nccl_channel_ud_map[channel_id][channel_type].udId) {
++        wrap_ionicdv_pd_set_udma_mask(base->pd, IONIC_UDMA_MASK_HIGH);
++    } else {
++        wrap_ionicdv_pd_set_udma_mask(base->pd, IONIC_UDMA_MASK_LOW);
++    }
++    qpInitAttr.sq_sig_all |= (1 << 16);
```
ddurnov, line 179: Just for the future iterations of changes would be great if we had these bits defined somewhere in header file.
Ideally I would recommend to introduce set_bit/clear_bit primitive if the functionality going to be extended

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
++      qpInitAttr.sq_sig_all |= (1 << 17);
++    } else {
++      qpInitAttr.sq_sig_all &= (~(1 << 17));
++    }
++    qpInitAttr.sq_sig_all |= (1 << 18);
++
++    if (rcclCtsOffloadEnabled) {
++      qpInitAttr.sq_sig_all |= (1 << 19);
++    } else {
++      qpInitAttr.sq_sig_all &= (~(1 << 19));
```
ddurnov, line 190: For the future (very minor comment). If there is a zeroing of the sq_sig_all at line 160, then bit reset may be redundant in general. I see the point of the altered behavior (set/clear) in cases where you modify existing attribute and you don't track its state. Here it looks like a redundant step

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
+   uint64_t addr;
+   uint32_t flags;
+@@ -1265,20 +1333,59 @@ returning:
+   return res;
+ }
+ 
+-ncclResult_t ncclIbCreateQp(uint8_t ib_port, struct ncclIbNetCommDevBase* base, int access_flags, void* qp_context, struct ncclIbQp* qp) {
++ncclResult_t ncclIbCreateQp(uint8_t ib_port, struct ncclIbNetCommDevBase* base,
++                            int access_flags, void* qp_context, struct ncclIbQp* qp,
++                            int channel_id, bool data_qp, int8_t cts_qp_slot) {
```
ddurnov, line 157: cts_qp_slot could be just one of extra attributes in the future for ncclIbCreateQp so ideally we need to put into the structure in the future

ddurnov: For the future. ```int channel_id, bool data_qp, int8_t cts_qp_slot``` all have to be part of qp attr structure. we are going to add extensions and we shouldn't touch function signature for it

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
++
+   if (stage->state == ncclIbCommStateConnect)      goto ib_connect_check;
+   if (stage->state == ncclIbCommStateSendDevList)  goto ib_send_dev_list;
+   if (stage->state == ncclIbCommStateRecvDevList)  goto ib_recv_dev_list;
+@@ -1461,7 +1576,7 @@ ib_recv_dev_list:
+   for (int q = 0; q < comm->base.nqps; q++) {
+     ncclIbSendCommDev* commDev = comm->devs + devIndex;
+     ncclIbDev* ibDev = ncclIbDevs + commDev->base.ibDevN;
+-    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &commDev->base, IBV_ACCESS_REMOTE_WRITE, &comm->base.stats, comm->base.qps + q), ret, fail);
++    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &commDev->base, IBV_ACCESS_REMOTE_WRITE, &comm->base.stats, comm->base.qps + q, channel_id, true, NCCL_CTS_QP_SLOT_INVALID), ret, fail);
```
ddurnov, line 250: For future I would recommend ```channel_id, true, NCCL_CTS_QP_SLOT_INVALID``` values fields passed via qp attr structure as mentioned earlier. It looks like we may need more attributes in the future and it is better if we wouldn't have to touch function signature for each new attribute we introduce

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
+     devInfo->fifoRkey = commDev->fifoMr->rkey;
+ 
+     // Pack local GID info
+@@ -1528,7 +1647,11 @@ ib_recv_dev_list:
+       return ncclInternalError;
+     }
+   }
+-  meta.fifoAddr = (uint64_t)comm->fifo;
++  if (rcclCtsInlineData) {
++    meta.fifoAddr = (uint64_t)comm->fifo_inline;
```
ddurnov, line 273: If we introduce this meta.fifoAddr anyway, then is it possible to do it earlier so we would have ibv_reg_mr called with meta.fifoAddr and some new meta.fifoSize?
From my perspective we introduce extra stuff and don't leverage it. E.g.
```NCCLCHECKGOTO(wrap_ibv_reg_mr(&commDev->fifoMr, commDev->base.pd, meta.fifoAddr, meta.fifoSize, IBV_ACCESS_LOCAL_WRITE|IBV_ACCESS_REMOTE_WRITE|IBV_ACCESS_REMOTE_READ), ret, fail);```

File: ext-src/rocm_netib.patch [hunk 19]
```diff
@@ -0,0 +1,797 @@
...
++
+   if (stage->state == ncclIbCommStateAccept)   goto ib_accept_check;
+   if (stage->state == ncclIbCommStateRecvDevList) goto ib_recv_dev_list;
+   if (stage->state == ncclIbCommStateSendDevList) goto ib_send_dev_list;
+@@ -1814,7 +1941,7 @@ ib_recv:
+     // Local ibDevN
+     ibDevN = rComm->devs[devIndex].base.ibDevN;
+     ibDev = ncclIbDevs + ibDevN;
+-    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &rCommDev->base, IBV_ACCESS_REMOTE_WRITE, &rComm->base.stats, qp), ret, fail);
++    NCCLCHECKGOTO(ncclIbCreateQp(ibDev->portNum, &rCommDev->base, IBV_ACCESS_REMOTE_WRITE, &rComm->base.stats, qp, channel_id, false, q), ret, fail);
```
ddurnov, line 310: Ideally there should be structure with explicit fields initialization e.g. ```qp_ext_attr = { .channel_id = channel_id; .data_qp = false; .cts_qp_slot = q;}``` rather than 3 function arguments (btw, ```is_data_qp``` if it is boolean). I would recommend one function argument with attr (we may use some ext in name in order to avoid collision with ibv_qp_attr approach) where we wouldn't have to get to function signature and check for order of arguments. Yes, IDEs have hints nowadays but going forward we may introduce more attributes and it will lead to more differences on code level

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
+   if (nreqs > NCCL_NET_IB_MAX_RECVS) return ncclInternalError;
+ 
+   uint64_t wr_id = 0ULL;
+@@ -2130,7 +2268,11 @@ ncclResult_t ncclIbMultiSend(struct ncclIbSendComm* comm, int slot) {
+     sge->addr=(uintptr_t)reqs[r]->send.data;
+     wr->opcode = IBV_WR_RDMA_WRITE;
+     wr->send_flags = 0;
+-    wr->wr.rdma.remote_addr = slots[r].addr;
++    if (rcclCtsOffloadEnabled) {
++      wr->wr.rdma.remote_addr = 0xdeadbeef;
```
ddurnov, line 373: I recommend to define magic names/numbers/bits and give them more meaningful context/name in the place where they got used if necessary

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
+   lastWr->next = NULL;
+   lastWr->send_flags = IBV_SEND_SIGNALED;
+ 
+@@ -2182,7 +2326,11 @@ ncclResult_t ncclIbMultiSend(struct ncclIbSendComm* comm, int slot) {
+       //ncclIbAddEvent(reqs[r], devIndex, &comm->devs[devIndex].base);
+ 
+       // Select proper rkey (needed even for 0-size send)
+-      comm->wrs[r].wr.rdma.rkey = slots[r].rkeys[qp->remDevIdx];
++      if (rcclCtsOffloadEnabled) {
++        comm->wrs[r].wr.rdma.rkey = 0xbade;
```
ddurnov, line 433: Same here. Better to define as something. If it is DUMMY value/variable, then let it be DUMMY or STUB for debug purposes. Otherwise it looks like special/magic value which definitely requires to be defined somewhere

ext-src/rocm_netib.patch
```diff
@@ -0,0 +1,797 @@
...
++      } else {
++        comm->wrs[r].wr.rdma.rkey = slots[r].rkeys[qp->remDevIdx];
```
ddurnov, line 435: I would prefer to keep original value initialization (line 431) and override if diff minimization is one of the goals here

ddurnov (commented):
Minor comments for future. We may have separate round of refactoring of several extensions/attributes handling parts in order to have more flexibility for future iterations.

ddurnov (approved):

mustafabar:
@ddurnov 's comments will be addressed in a future refactor.

PR #2085, opened 2025-12-03
Virtual device enablement 

PJAvinash:
## Details 
**Work item:** "Internal"

**What were the changes?**  
- virtual device enablement through change in function input
- Code refactoring to follow separation of concerns (ncclTopoGetSystemNew)

- makeVDevice function for net plugin gets invoked during init
- ncclTopoGetSystem has different logic to create comm-> topo from xml depending on dumpXmlFile value . code is refactored to address this. 

- only additive changes for new functions
- few changes on existing helper functions to indicate/enforce xml input remains unchanged
**Additional Documentation:**  
- using this feature is giving lower internode performance than not using it ( in a 2 node system net/ib) 
- This PR only serves enablement, not optimization

### DIFF
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -1464,7 +1464,7 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
   NCCLCHECKGOTO(ncclTopoGetSharedState(&state, comm->ncclNet->name, netStates), ret, fail);
   // [RCCL] Disabled virtual devices
   NCCLCHECKGOTO(ncclTopoProcessNet(xml, 0, dumpXmlFile, state,
-    comm->ncclNet->getProperties, nullptr /*comm->ncclNet->makeVDevice*/, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
+    comm->ncclNet->getProperties, /*nullptr*/ comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
   pthread_mutex_unlock(&netLock);
   netLockHeld = 0;
 
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -523,6 +523,9 @@ ncclResult_t ncclTopoAddCpu(struct ncclXmlNode* xmlCpu, struct ncclTopoSystem* s
   int systemId;
   NCCLCHECK(ncclGetSystemId(system, xmlCpu, &systemId));
   struct ncclTopoNode* cpu;
+  /*
+  system->nodes[NET] and other fields are initialized in the below function
+  */
   NCCLCHECK(ncclTopoCreateNode(system, &cpu, CPU, NCCL_TOPO_ID(systemId, numaId)));
   const char* str;
   NCCLCHECK(xmlGetAttr(xmlCpu, "affinity", &str));
@@ -1327,7 +1330,7 @@ static ncclResult_t ncclTopoPopulateNics(ncclXml* xml, int startIndex, int endIn
       if (net == NULL) NCCLCHECK(ncclTopoGetVNicParent(xml, getProperties, &props.vProps, &parent));
     }
 
-    NCCLCHECK(ncclTopoFillNet(xml, props.pciPath, props.name, &netNode, parent));
+    NCCLCHECK(ncclTopoFillNet(xml, props.pciPath, props.name, &netNode, parent));          /*parent->name is 'cpu'*/
 
     const char* colAttr;
     NCCLCHECK(xmlGetAttr(netNode, "coll", &colAttr));
@@ -1579,6 +1582,9 @@ ncclResult_t getLocalNetCountByBw(struct ncclTopoSystem* system, int gpu, int *c
   return ncclSuccess;
 }
 
+/*
+** the following function modifies system->nodes[NET]
+*/
 ncclResult_t ncclTopoGetLocalNet(struct ncclTopoSystem* system, int rank, int channelId, int64_t* id, int* dev) {
   int gpu;
   NCCLCHECK(ncclTopoRankToIndex(system, rank, &gpu, /*showWarn=*/true));
--- a/src/graph/topo.cc
+++ b/src/graph/topo.cc
@@ -563,7 +563,7 @@ ncclResult_t ncclTopoAddCpu(struct ncclXmlNode* xmlCpu, struct ncclTopoSystem* s
   }
   for (int s=0; s<xmlCpu->nSubs; s++) {
     struct ncclXmlNode* node = xmlCpu->subs[s];
-    if (strcmp(node->name, "pci") == 0) NCCLCHECK(ncclTopoAddPci(node, system, cpu, systemId, numaId));
+    if (strcmp(node->name, "pci") == 0) NCCLCHECK(ncclTopoAddPci(node, system, cpu, systemId, numaId)); /* Adds pci in system.nodes[Net]*/
     if (strcmp(node->name, "nic") == 0) {
       struct ncclTopoNode* nic = NULL;
       int64_t localNicId = NCCL_TOPO_LOCAL_NIC_ID(numaId, 0);
@@ -1328,6 +1328,8 @@ static ncclResult_t ncclTopoPopulateNics(ncclXml* xml, int startIndex, int endIn
       // In the event of multithreaded use case, we need to re-discover the shared parent of the given devices for this vNIC
       // Only run this if the net doesn't exist locally - this may alter the XML state
       if (net == NULL) NCCLCHECK(ncclTopoGetVNicParent(xml, getProperties, &props.vProps, &parent));
+      INFO(NCCL_GRAPH, "vNIC populate: props.name=%s pciPath=%s", props.name, props.pciPath);
+      if (parent == NULL) WARN("vNIC parent discovery failed for %s (check pciPath + ordering)", props.name);
     }
 
     NCCLCHECK(ncclTopoFillNet(xml, props.pciPath, props.name, &netNode, parent));          /*parent->name is 'cpu'*/
@@ -1366,6 +1368,7 @@ ncclResult_t ncclTopoProcessNet(ncclXml* xml, int coll, const char* dumpXmlFile,
   int usePhysicalDevices = (dumpXmlFile || makeVDevice == NULL);
   if (state->nPhysicalNics == -1) NCCLCHECK(devices(&state->nPhysicalNics));
   // Enumerate physical devices
+  ncclTopoDumpXmlToFile("/home/apotnuru/LWPCOMMLIBS-706/ncclTopoProcessNet_1371.xml", xml);
   NCCLCHECK(ncclTopoPopulateNics(xml, 0, state->nPhysicalNics, getProperties, netName, coll, false, dmaBufSupport));
   if (!usePhysicalDevices) {
     if (state->nVirtualNics == -1) {
@@ -1376,7 +1379,9 @@ ncclResult_t ncclTopoProcessNet(ncclXml* xml, int coll, const char* dumpXmlFile,
     }
     if (state->nVirtualNics > 0) {
       // Populate new devices
+      ncclTopoDumpXmlToFile("/home/apotnuru/LWPCOMMLIBS-706/ncclTopoProcessNet_1382.xml", xml);
       NCCLCHECK(ncclTopoPopulateNics(xml, state->nPhysicalNics, state->nPhysicalNics+state->nVirtualNics, getProperties, netName, coll, true, dmaBufSupport));
+      ncclTopoDumpXmlToFile("/home/apotnuru/LWPCOMMLIBS-706/ncclTopoProcessNet_1384.xml", xml);
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/graph/topo.cc
```diff
@@ -1404,6 +1408,195 @@ ncclResult_t ncclTopoGetSharedState(ncclTopoNetState** state, const char* name,
...
+ * The caller is responsible for:
+ *  - XML object's memory ownership
+ *  - Calling this function before XML trimming and XML fusion steps
+*/
+ncclResult_t ncclTopoImportNetPlugins(ncclXml* xml,const struct ncclComm* comm){
+  // Auto-detect NICs if needed. net/collnet share the same xml/graph nodes,
+  // so we start with collnet so that it has precedence.
+  ncclResult_t ret = ncclSuccess;
+  int netLockHeld = 0;
+  pthread_mutex_lock(&netLock);
```
alex-breslow-amd, line 1429: This is guaranteed to succeed.  Probably check for zero status: https://pubs.opengroup.org/onlinepubs/7908799/xsh/[pthread_mutex_lock](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_lock.html).html

src/graph/topo.cc
```diff
@@ -1404,6 +1408,195 @@ ncclResult_t ncclTopoGetSharedState(ncclTopoNetState** state, const char* name,
...
+  if (collNetSupport(comm)) {
+    NCCLCHECKGOTO(ncclTopoGetSharedState(&state, comm->ncclCollNet->name, collNetStates), ret, fail);
+    NCCLCHECKGOTO(ncclTopoProcessNet(xml, 1, NULL, state,
+      comm->ncclCollNet->getProperties, comm->ncclCollNet->makeVDevice, comm->ncclCollNet->devices, comm->ncclCollNet->name, comm->dmaBufSupport), ret, fail);
+  }
+  NCCLCHECKGOTO(ncclTopoGetSharedState(&state, comm->ncclNet->name, netStates), ret, fail);
+  // [RCCL] Disabled virtual devices
+  NCCLCHECKGOTO(ncclTopoProcessNet(xml, 0, NULL, state,
+    comm->ncclNet->getProperties, /*nullptr*/ comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
+  pthread_mutex_unlock(&netLock);
```
alex-breslow-amd, line 1443: Check here too and other pthread calls.  See reference I linked for what the macro check probably should examine.  If this returns nonzero, we're in a bad place, but we should catch it at least.

src/graph/topo.cc
```diff
@@ -1404,6 +1408,195 @@ ncclResult_t ncclTopoGetSharedState(ncclTopoNetState** state, const char* name,
...
+    nLocalRanks = comm->clique.size;
+    localRank = comm->cliqueRank;
+    localRanks = comm->clique.ranks;
+  } else {
+    // Intra-node fusion.  Much of the comm is not initialized yet at this point so we need to do our own calculations.
+    NCCLCHECKGOTO(ncclCalloc(&localRanks, comm->nRanks), ret, fail);
+    for (int i = 0; i < comm->nRanks; i++) {
+      if (comm->peerInfo[i].hostHash == comm->peerInfo[comm->rank].hostHash) {
+        if (i == comm->rank)
+          localRank = nLocalRanks;
```
alex-breslow-amd: Can you put this in curly braces?  That should help avoid a common cause of errors.

src/graph/topo.cc
```diff
@@ -1404,6 +1408,195 @@ ncclResult_t ncclTopoGetSharedState(ncclTopoNetState** state, const char* name,
...
+  return ret;
+}
+
+/**
+ * @brief  If xml is empty (maxIndex == 0), create a <system> root node and set version.
+ * @param xml [in,out]
+ */
+static ncclResult_t ncclTopoEnsureSystemRoot(struct ncclXml* xml) {
+  ncclResult_t ret = ncclSuccess;
+  if (xml->maxIndex == 0) {
```
alex-breslow-amd, line 1480: Check for nullness too?

src/graph/topo.cc
```diff
@@ -1313,7 +1315,7 @@ ncclResult_t ncclTopoMakeVNics(struct ncclXml* xml, ncclResult_t (*makeVDevice)(
   return res;
 }
 
-static ncclResult_t ncclTopoPopulateNics(ncclXml* xml, int startIndex, int endIndex, ncclResult_t (*getProperties)(int, ncclNetProperties_t*), const char* netName, int coll, int virtualNics, bool dmaBufSupport) {
+static ncclResult_t ncclTopoPopulateNics(ncclXml* xml, int startIndex, int endIndex, ncclResult_t (*getProperties)(int, ncclNetProperties_t*), const char* netName, int coll, bool virtualNics, bool dmaBufSupport) {
```
alex-breslow-amd, line 1318: Ok, I guess it makes sense if it's being used as a boolean.

src/graph/topo.cc
```diff
@@ -1360,7 +1364,7 @@ static ncclResult_t ncclTopoPopulateNics(ncclXml* xml, int startIndex, int endIn
 
 // Calls to network plugin APIs should be protected. This function should be called inside a per-process lock.
 ncclResult_t ncclTopoProcessNet(ncclXml* xml, int coll, const char* dumpXmlFile, ncclTopoNetState* state, ncclResult_t (*getProperties)(int, ncclNetProperties_t*), ncclResult_t (*makeVDevice)(int*, ncclNetVDeviceProps_t*), ncclResult_t (*devices)(int*), const char* netName, bool dmaBufSupport) {
-  int usePhysicalDevices = (dumpXmlFile || makeVDevice == NULL);
+  int usePhysicalDevices = (makeVDevice == NULL); /*(dumpXmlFile || makeVDevice == NULL);*/
```
corey-derochie-amd, line 1367:
```suggestion
  int usePhysicalDevices = (dumpXmlFile || makeVDevice == NULL);
```
We should maintain the old xml dump functionality.

src/graph/topo.cc
```diff
@@ -1464,13 +1657,11 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
   NCCLCHECKGOTO(ncclTopoGetSharedState(&state, comm->ncclNet->name, netStates), ret, fail);
   // [RCCL] Disabled virtual devices
   NCCLCHECKGOTO(ncclTopoProcessNet(xml, 0, dumpXmlFile, state,
-    comm->ncclNet->getProperties, nullptr /*comm->ncclNet->makeVDevice*/, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
+    comm->ncclNet->getProperties, /*nullptr*/ comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
```
corey-derochie-amd:
```suggestion
    comm->ncclNet->getProperties, comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
```

src/transport/net_ib.cc
```diff
@@ -561,7 +546,7 @@ static bool ncclMlx5dvDmaBufCapable(ibv_context *context){
 
 ncclResult_t ncclIbMakeVDeviceInternal(int* d, ncclNetVDeviceProps_t* props) {
   if (ncclParamIbMergeNics() == 0 && props->ndevs > 1) {
-    INFO(NCCL_NET, "NET/IB : Skipping makeVDevice, NCCL_IB_MERGE_NICS=0");
+    WARN("NET/IB : Skipping makeVDevice, Please set NCCL_IB_MERGE_NICS=1");
```
corey-derochie-amd, line 549:
```suggestion
    INFO(NCCL_NET, "NET/IB : Skipping makeVDevice, NCCL_IB_MERGE_NICS=0");
```
This isn't a warning case, the user may rightfully choose to set NCCL_IB_MERGE_NICS to 0. Let's leave the INFO log as-is.

src/graph/xml.h
corey-derochie-amd, line 1: I appreciate the const-correctness, but I don't think this is necessary for this PR, this will lead to unnecessary NCCL merge conflicts, and there are several places that we are casting out of const-ness anyway. Let's undo these.

src/graph/topo.cc
```diff
@@ -1464,13 +1657,11 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
...
   pthread_mutex_unlock(&netLock);
   netLockHeld = 0;
```
corey-derochie-amd, line 1667: No need for unnecessary whitespace changes.

src/init.cc
```diff
@@ -1241,11 +1224,14 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
...
-    NCCLCHECKGOTO(ncclTopoGetSystem(comm, NULL, dumpXmlFile), ret, fail);
+    //NCCLCHECKGOTO(ncclTopoGetSystem(comm, NULL, dumpXmlFile), ret, fail);
+    NCCLCHECKGOTO(ncclTopoGetSystemNew(comm, &comm->topo, dumpXmlFile), ret, fail);
+  }else{
+    NCCLCHECKGOTO(ncclTopoGetSystemNew(comm, &comm->topo, NULL), ret, fail);
   }
 
   // Topo detection / System graph creation
-  NCCLCHECKGOTO(ncclTopoGetSystem(comm, &comm->topo), ret, fail);
+  //NCCLCHECKGOTO(ncclTopoGetSystem(comm, &comm->topo), ret, fail);
```
corey-derochie-amd:
```suggestion

  // Topo detection / System graph creation
  NCCLCHECKGOTO(ncclTopoGetSystemNew(comm, &comm->topo, dumpXmlFile), ret, fail);
```
This should be equivalent. I'm wondering, is `ncclTopoGetSystem` called anywhere now? If so, why not modify it instead of adding `ncclTopoGetSystemNew`?

src/graph/topo.cc
```diff
@@ -560,7 +562,7 @@ ncclResult_t ncclTopoAddCpu(struct ncclXmlNode* xmlCpu, struct ncclTopoSystem* s
   }
   for (int s=0; s<xmlCpu->nSubs; s++) {
     struct ncclXmlNode* node = xmlCpu->subs[s];
-    if (strcmp(node->name, "pci") == 0) NCCLCHECK(ncclTopoAddPci(node, system, cpu, systemId, numaId));
+    if (strcmp(node->name, "pci") == 0) NCCLCHECK(ncclTopoAddPci(node, system, cpu, systemId, numaId)); /* Adds pci in system.nodes[Net]*/
```
corey-derochie-amd: This comment probably applies to `ncclTopoAddPci` function rather than just here.
```suggestion
    if (strcmp(node->name, "pci") == 0) NCCLCHECK(ncclTopoAddPci(node, system, cpu, systemId, numaId));
```

wenkaidu:
I had discussion with Avinash. My suggestion is we need to make sure the xml generated is logically compatible with NCCL does to avoid confusion to customers who work on both platforms. i.e. if NCCL dumps XML before merging NICs, RCCL should be doing the same.

src/transport/net_ib.cc
PJAvinash: When user is choosing NCCL_IB_MERGE_NICS = 0 and NCCL_NET_FORCE_MERGE="a,b;c,d;e,f .."
RCCL is failing due to conflicting inputs from user. It was difficult to trace the root cause without this warning. One of these environment variables should be of higher priority and In case of conflict, a warning or silently succeeding should be implemented.

src/graph/topo.cc
```diff
@@ -1464,7 +1663,7 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
   NCCLCHECKGOTO(ncclTopoGetSharedState(&state, comm->ncclNet->name, netStates), ret, fail);
   // [RCCL] Disabled virtual devices
```
corey-derochie-amd:
```suggestion
```

src/graph/topo.cc
```diff
@@ -1464,7 +1663,7 @@ ncclResult_t ncclTopoGetSystem(struct ncclComm* comm, struct ncclTopoSystem** sy
...
   NCCLCHECKGOTO(ncclTopoProcessNet(xml, 0, dumpXmlFile, state,
-    comm->ncclNet->getProperties, nullptr /*comm->ncclNet->makeVDevice*/, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
+    comm->ncclNet->getProperties,comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
```
corey-derochie-amd:
```suggestion
    comm->ncclNet->getProperties, comm->ncclNet->makeVDevice, comm->ncclNet->devices, comm->ncclNet->name, comm->dmaBufSupport), ret, fail);
```

src/graph/xml.h
```diff
@@ -173,7 +173,7 @@ static ncclResult_t xmlFindNextTag(struct ncclXml* xml, const char* tagName, str
   return ncclSuccess;
 }
 
-static ncclResult_t xmlFindTagKv(struct ncclXml* xml, const char* tagName, struct ncclXmlNode** node, const char* attrName, const char* attrValue) {
+static ncclResult_t xmlFindTagKv( struct ncclXml* xml, const char* tagName, struct ncclXmlNode** node, const char* attrName, const char* attrValue) {
```
corey-derochie-amd:
```suggestion
static ncclResult_t xmlFindTagKv(struct ncclXml* xml, const char* tagName, struct ncclXmlNode** node, const char* attrName, const char* attrValue) {
```

src/graph/topo.cc
corey-derochie-amd, line 1: I've reviewed this added code, and I see no functional changes from original other than just re-enabling via passing `makeVDevice` and the changes to xml dump. We don't want to refactor this code as this poses a risk to NCCL sync merges, so all these changes should be undone.

corey-derochie-amd (changes requested):

src/graph/topo.cc
corey-derochie-amd: @alex-breslow-amd can you also comment on this?

corey-derochie-amd:
Replaced by #2110

PR #2087, opened 2025-12-03
Re-enable virtual devices

corey-derochie-amd:
Enabling virtual devices, as cherry-pick against 10-31 drop.


corey-derochie-amd:
Replaced by #2110

PR #2091, opened 2025-12-05
Removes default visibility in debug mode and updates unit tests for alt_rsmi impl

atulkulk:
**Work item:** Internal

**What were the changes?**  
- Create distinct test executable for alt_rsmi testing
- Updated alt_rsmi tests to use public methods
- Compiles alt_rsmi.cc with ARSMI_TEST_BUILD
- Enables external linkage of internal variables
- Only for AltRsmiTests.cpp that manipulates internals
- Clean separation for test behavior

To improve the maintainability of the unit tests

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/misc/alt_rsmi.cc
+++ b/src/misc/alt_rsmi.cc
@@ -48,17 +48,25 @@ struct ARSMI_systemNode {
     std::string s_card;
 };
 
-static const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
+// Internal implementation details
+// When compiled into the test binary (ARSMI_TEST_BUILD defined), these have external linkage
+// so AltRsmiTestUtils.cpp can access them via extern declarations.
+// When compiled into the production library (ARSMI_TEST_BUILD not defined), these are static.
+#ifdef ARSMI_TEST_BUILD
+  // Test build: external linkage for cross-TU access
+  thread_local const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
+  thread_local std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
+  thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
+  thread_local int ARSMI_num_devices=-1;
+#else
+  // Production build: static/internal linkage
+  static thread_local const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
+  static thread_local std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
+  static thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
+  static thread_local int ARSMI_num_devices=-1;
+#endif
+
 static const uint32_t kAmdGpuId = 0x1002;
-
-// Vector containing data about each node, ordered by bdf ID
-static thread_local std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
-
-// 2-D matrix with link information between each pair of nodes.
-static thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
-
-// Number of devices recognized
-static thread_local int ARSMI_num_devices=-1;
 
 
 // Public API functions
--- a/src/misc/alt_rsmi.cc
+++ b/src/misc/alt_rsmi.cc
@@ -54,20 +54,23 @@ struct ARSMI_systemNode {
 // When compiled into the production library (ARSMI_TEST_BUILD not defined), these are static.
 #ifdef ARSMI_TEST_BUILD
   // Test build: external linkage for cross-TU access
-  thread_local const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
-  thread_local std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
-  thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
-  thread_local int ARSMI_num_devices=-1;
+  #define ARSMI_LOCAL thread_local
 #else
   // Production build: static/internal linkage
-  static thread_local const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
-  static thread_local std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
-  static thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
-  static thread_local int ARSMI_num_devices=-1;
+  #define ARSMI_LOCAL static thread_local
 #endif
 
+ARSMI_LOCAL const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
 static const uint32_t kAmdGpuId = 0x1002;
 
+// Vector containing data about each node, ordered by bdf ID
+ARSMI_LOCAL std::vector<ARSMI_systemNode> ARSMI_orderedNodes;
+
+// 2-D matrix with link information between each pair of nodes.
+ARSMI_LOCAL std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
+
+// Number of devices recognized
+ARSMI_LOCAL int ARSMI_num_devices=-1;
 
 // Public API functions
 int ARSMI_init(void)
--- a/test/AltRsmiTests.cpp
+++ b/test/AltRsmiTests.cpp
@@ -4,6 +4,7 @@
  * See LICENSE.txt for license information
  ************************************************************************/
 #include "alt_rsmi.h"
+#include "common/ProcessIsolatedTestRunner.hpp"
 
 #include <cerrno>
 #include <cstdio>
@@ -18,6 +19,8 @@
 #include <unistd.h>
 #include <vector>
 #include <cstdlib>
+#include <limits>
+#include <filesystem>
 
 // ============================================================================
 // Internal structures and variables from alt_rsmi.cc (TEST USE ONLY)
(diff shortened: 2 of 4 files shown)


Kapil-Shyam-Pawar (approved):

venksubramd:
Looks good to me, @atulkulk

src/misc/alt_rsmi.cc
```diff
@@ -48,17 +48,25 @@ struct ARSMI_systemNode {
...
+  static thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
+  static thread_local int ARSMI_num_devices=-1;
+#endif
 
-// 2-D matrix with link information between each pair of nodes.
-static thread_local std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;
-
-// Number of devices recognized
-static thread_local int ARSMI_num_devices=-1;
+static const uint32_t kAmdGpuId = 0x1002;
```
corey-derochie-amd, line 64:
```suggestion
// Internal implementation details
// When compiled into the test binary (ARSMI_TEST_BUILD defined), these have external linkage
// so AltRsmiTestUtils.cpp can access them via extern declarations.
// When compiled into the production library (ARSMI_TEST_BUILD not defined), these are static.
#ifdef ARSMI_TEST_BUILD
  // Test build: external linkage for cross-TU access
  #define ARSMI_LOCAL thread_local
#else
  // Production build: static/internal linkage
  #define ARSMI_LOCAL static thread_local
#endif

ARSMI_LOCAL const char *kKFDNodesPathRoot = "/sys/class/kfd/kfd/topology/nodes";
static const uint32_t kAmdGpuId = 0x1002;

// Vector containing data about each node, ordered by bdf ID
ARSMI_LOCAL std::vector<ARSMI_systemNode> ARSMI_orderedNodes;

// 2-D matrix with link information between each pair of nodes.
ARSMI_LOCAL std::vector<std::vector<ARSMI_linkInfo>> ARSMI_orderedLinks;

// Number of devices recognized
ARSMI_LOCAL int ARSMI_num_devices=-1;
```
This would avoid redundant definitions and maintain code similarity.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+    // Create the final directory
+    if (mkdir(path.c_str(), 0700) == -1 && errno != EEXIST) {
+      return -1; // Return error if directory creation fails
+    }
+
+    return 0; // Success
+  }
+
+  // Helper function to remove a directory recursively
+  int removeDirectory(const std::string &path) {
```
corey-derochie-amd: `createDirectory` and `removeDirectory` can be `static`
```suggestion
  static int createDirectory(const std::string &path) {
    size_t pos = 0;
    std::string currentPath;

    // Iterate through each component of the path
    while ((pos = path.find('/', pos)) != std::string::npos) {
      currentPath = path.substr(0, pos++);
      if (!currentPath.empty() && mkdir(currentPath.c_str(), 0700) == -1 &&
          errno != EEXIST) {
        return -1; // Return error if directory creation fails
      }
    }

    // Create the final directory
    if (mkdir(path.c_str(), 0700) == -1 && errno != EEXIST) {
      return -1; // Return error if directory creation fails
    }

    return 0; // Success
  }

  // Helper function to remove a directory recursively
  static int removeDirectory(const std::string &path) {
```

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+      std::cerr << "Failed to remove directory: " << path
+                << " (errno: " << errno << ")" << std::endl;
+      return -1;
+    }
+
+    return 0; // Success
+  }
+
+  // Helper function to create a file with content
+  void createFile(const std::string &path, const std::string &content) {
```
corey-derochie-amd:
```suggestion
  static void createFile(const std::string &path, const std::string &content) {
```

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+    if (!file) {
+      std::cerr << "Failed to create file: " << path << ", errno: " << errno << std::endl;
+      return;
+    }
+    file << content;
+    file.close();
+  }
+
+  // Helper function to remove a file
+  int removeFile(const std::string &path) {
```
corey-derochie-amd:
```suggestion
  static int removeFile(const std::string &path) {
```

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+    if (unlink(path.c_str()) == -1) {
+      std::cerr << "Failed to remove file: " << path << " (errno: " << errno
+                << ")" << std::endl;
+      return -1; // Return error if file removal fails
+    }
+    return 0; // Success
+  }
+
+  // Function to create the test directory structure and files
+  void setupTestFiles() {
```
corey-derochie-amd, line 236:
```suggestion
  static void setupTestFiles() {
```

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+             "min_latency 0\n"
+             "max_latency 0\n"
+             "min_bandwidth 0\n"
+             "max_bandwidth 64000\n"
+             "recommended_transfer_size 0\n"
+             "recommended_sdma_engine_id_mask 0\n"
+             "flags 0\n");
+
+  int result = ARSMI_init();
+  ASSERT_NE(result, 0); // Expect non-zero error code when required property is missing
```
corey-derochie-amd: This test would be stronger if we knew what error to expect if a required property was missing, but this is fine if the error code is generic.

test/AltRsmiTests.cpp
corey-derochie-amd: You always call every file function within the test directory, so why not bake it into the file functions? That will also force future tests to do the same, which is safer that accidentally modifying other files on the system.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+}
+
+TEST_F(AltRsmiTest, ARSMIInitTypePCIeProperty) {
+  int result = ARSMI_init();
+  ASSERT_EQ(result, 0);
+
+  // Verify link info can be retrieved for PCIe connections
+  ARSMI_linkInfo info;
+  ASSERT_EQ(ARSMI_topo_get_link_info(0, 1, &info), 0);
+  // The link should be detected (type will be set based on properties)
```
corey-derochie-amd: Is this test incomplete?

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+int GetNumDevices() {
+    return ARSMI_num_devices;
+}
+
+} // namespace AltRsmiTestUtils
+
+// Test paths for creating mock KFD filesystem
+static const char* kTestKFDBasePath = "/tmp/test_kfd_arsmi";
+// IMPORTANT: This must match the default value in alt_rsmi.cc
+// Always restore this path after tests to avoid affecting other test suites
```
corey-derochie-amd: Should these tests be process-isolated to avoid this constraint?

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+static const char* kSystemKFDPath = "/sys/class/kfd/kfd/topology/nodes";
+static const char* kTestKFDPath = "/tmp/test_kfd_arsmi/topology/nodes";
```
corey-derochie-amd: Consider using `std::filesystem`, and consider using `std::filesystem::temp_directory_path()` to get a temp directory, instead of hard-coding a temp directory path that may or may not be valid.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+             "recommended_transfer_size 0\n"
+             "recommended_sdma_engine_id_mask 0\n"
+             "flags 0\n");
+
+  int result = ARSMI_init();
+  ASSERT_NE(result, 0); // Expect non-zero error code when max_bandwidth is missing
+}
+
+TEST_F(AltRsmiTest, ARSMIGetNumDevicesUninitialized) {
+  // Don't call ARSMI_init, let ARSMI_get_num_devices initialize
```
corey-derochie-amd: Is there a way to verify that `ARSMI_init` was called after `ARSMI_get_num_devices` is called? If so, you may want to assert that here.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+  ASSERT_EQ(ARSMI_get_num_devices(&num_devices), 0);
+  ASSERT_GT(num_devices, 0);
+
+  // Try to get BDF for out-of-range device index
+  uint64_t bdfid = 0;
+  uint32_t invalid_index = num_devices + 10;
+
+  // Note: ARSMI_dev_pci_id_get does not perform bounds checking
+  // This test documents that behavior - in production, callers must validate indices
+  // Accessing out-of-range index will cause undefined behavior (likely crash or garbage data)
```
corey-derochie-amd: Should this just be fixed? Otherwise, I don't know if a no-op test is the best way to document the behaviour.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+
+  // Try to get link info between the two devices (no direct link defined)
+  ARSMI_linkInfo info;
+  int result = ARSMI_topo_get_link_info(0, 1, &info);
+
+  // Should succeed but return default values since no io_links are defined
+  ASSERT_EQ(result, 0);
+  ASSERT_EQ(info.hops, 2); // Default hops
+  ASSERT_EQ(info.type, ARSMI_IOLINK_TYPE_PCIEXPRESS); // Default type
+  ASSERT_EQ(info.weight, 40); // Default weight
```
corey-derochie-amd: Is this actually correct behaviour for this call? Seems questionable.

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+  uint64_t bdf0 = 0, bdf1 = 0;
+  ASSERT_EQ(ARSMI_dev_pci_id_get(0, &bdf0), 0);
+  ASSERT_EQ(ARSMI_dev_pci_id_get(1, &bdf1), 0);
+
+  // BDFs should be ordered (lower BDF first)
+  // Based on BDF ordering: node0 (domain=0, location_id=23552) comes before
+  // node1 (domain=1, location_id=23553)
+  // This verifies that ordering is working correctly
+  ASSERT_NE(bdf0, 0);
+  ASSERT_NE(bdf1, 0);
```
corey-derochie-amd: This doesn't test the ordering, right?

test/AltRsmiTests.cpp
```diff
@@ -0,0 +1,932 @@
...
+}
+
+TEST_F(AltRsmiTest, FileExistsCheck) {
+  // Test fileExists() indirectly by verifying behavior when files don't exist
+  // This covers the fileExists(char const*) internal function
```
corey-derochie-amd: I think `fileExists` probably doesn't need to be tested, especially since you are forced to test it indirectly anyway. ARSMI is the API under test, and `fileExists` should already have been implicitly tested by other tests. You should check your coverage report to confirm.

corey-derochie-amd (approved):
Seems fine to me, with a bunch of suggestions and considerations. Please have a look.

src/misc/alt_rsmi.cc
atulkulk: Applied.

test/AltRsmiTests.cpp
atulkulk: Moved to anonymous namespace

test/AltRsmiTests.cpp
corey-derochie-amd: Test is still valid, keeping.

corey-derochie-amd (approved):

test/AltRsmiTests.cpp
atulkulk: Yes, the test is still valid and need to be kept. The function is just not being used anywhere at the moment but it is a public function.

PR #2093, opened 2025-12-08
Tuning: fixed out-of-bound access

alexander-sannikov:
## Details
For messages with size less than `2^6`, `log2i(nBytes>>6);` [returns -1](https://github.com/ROCm/rccl/blob/29e1567b95e28823b0beb1a988adc587bfab5b4f/src/include/bitops.h#L272).
Thus access to CorrectionFactor (`rcclTuningModel[comm->topo->tuning].treeCorrectionFactor[protocol][logSize];`) targets invalid location - last value value from section defining prior protocol, or memory value right before treeCorrectionFactor table.

**What were the changes?**  
Added CorrectionFactor table indexes correction to prevent out-of-bound access.

Prevents improper algorithm selection logic, or algo estimation errors leading to error like:
```
failed: invalid usage (run with NCCL_DEBUG=WARN for details). Last NCCL warning(error) log entry (may be unrelated) 'Error : no algorithm/protocol available for function AllReduce with datatype ncclFloat32. NCCL_ALGO was set to TREE. NCCL_PROTO was set to LL128.'.
```

Added check for negative `log2i` return value.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/graph/tuning.cc
+++ b/src/graph/tuning.cc
@@ -143,16 +143,24 @@ static const float baseLat  [NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
 #define NCCL_HW_PCI 1
 #define NCCL_HW_NET 2
 
-
-
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
+#define RCCL_FACTOR_TABLE_SIZE 27
+#define RCCL_FACTOR_TABLE_MAX_INDEX (RCCL_FACTOR_TABLE_SIZE - 1)
+#endif
 struct tuningModel {
   float hwLat [3][NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS];
   float bwRatio [2][NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS];
+#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
+  float treeCorrectionFactor[NCCL_NUM_PROTOCOLS][RCCL_FACTOR_TABLE_SIZE];
+  float ringCorrectionFactor[NCCL_NUM_PROTOCOLS][RCCL_FACTOR_TABLE_SIZE];
+#else
   float treeCorrectionFactor[NCCL_NUM_PROTOCOLS][27];
   float ringCorrectionFactor[NCCL_NUM_PROTOCOLS][27];
+#endif
   uint64_t llProtoRanges[RCCL_TUNABLE_COLLS][NCCL_NUM_PROTOCOLS - 1][RCCL_PROTOCOL_ENTRY_SIZE];
   uint64_t channelThresholds[RCCL_TUNABLE_COLLS][RCCL_CHANNELS_TUNABLE_ENTRIES][3]; //for each collective, set for 5 channel-counts: 2,4,8,16,32,40,48,56,64, {min,max,nchannels}
 };
+
 
 static struct tuningModel tuning_model_0 {
   .hwLat = {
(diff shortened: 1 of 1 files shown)


src/graph/tuning.cc
```diff
@@ -888,14 +888,16 @@ ncclResult_t ncclTopoGetAlgoTime(struct ncclComm* comm, int coll, int algorithm,
   int logSize = log2i(nBytes>>6);
```
gilbertlee-amd, line 896: @mustafabar / @alexander-sannikov 
Can we introduce a constant for the number of elements in tree/ring CorrectionFactor then
make this int logSize = std::max(0, std::min(log2i(nBytes>>6), MAX_FACTOR_TABLE_SIZE);

alexander-sannikov: Was trying to keep it style-aligned/minimal change.
But if there are no objections from @mustafabar - I may clean it up a bit.

UPD: as far as I see, such change wont interfere with a [merge process](https://github.com/ROCm/rccl/pull/2040/files#diff-8029994d1ac745bdd5aa9db869bc52ccd10ad396e54b88acbb4d52e4cb343085), so should be safe.

src/graph/tuning.cc
```diff
@@ -888,14 +888,16 @@ ncclResult_t ncclTopoGetAlgoTime(struct ncclComm* comm, int coll, int algorithm,
...
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
 
+  if (logSize < 0) logSize = 0;
+  if (logSize > 26) logSize = 26;
```
mustafabar: is this check needed? given that the original code already has it in line 892. I am more of a minimal and equally robust fix mindset, so having just your change in line 891 should be enough. We can still introduce `MAX_FACTOR_TABLE_SIZE` constant and use it instead of 27 and 26 references. 
We should also change how the array is defined here https://github.com/ROCm/rccl/blob/7e10267dfd3f489ce7df96f6b5c8840408e12c84/src/graph/tuning.cc#L151

mustafabar: Since this is AMD only portion, we can go with the cleaner change suggested from @gilbertlee-amd . I thought it was a shared portion of code, but please revisit all direct hardcodes of numeric values for `MAX_FACTOR_TABLE_SIZE`

alexander-sannikov: Minimal required fix is:
```
if (logSize < 0) logSize = 0;
```

As soon as whole changed block is RCCL specific - proposed change would not interfere with  NCCL baseline,
although  changes in line 151 - will do that.
So we may have something like this:
```
#define RCCL_TUNING_MODEL_CORRECTION_FACTOR_SIZE 27
static_assert(sizeof(rcclTuningModel[0].treeCorrectionFactor[0]) == RCCL_TUNING_MODEL_CORRECTION_FACTOR_SIZE * sizeof(float), "treeCorrectionFactor size missmatch");
```
And use `RCCL_TUNING_MODEL_CORRECTION_FACTOR_SIZE ` instead of 27 in RCCL code. That would eliminate any modifications in original code.

mustafabar (approved):

alex-breslow-amd (approved):
Thank you for fixing this! :-)

src/graph/tuning.cc
```diff
@@ -887,15 +895,14 @@ ncclResult_t ncclTopoGetAlgoTime(struct ncclComm* comm, int coll, int algorithm,
   }
   int logSize = log2i(nBytes>>6);
 #if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
-
+  logSize = std::max(0, std::min(RCCL_FACTOR_TABLE_MAX_INDEX, logSize));
```
alex-breslow-amd, line 898: I'm wondering if we should log a fatal here if the logSize is out of bounds.  The risk here is that it can silently get thresholded.  Is silent thresholding really the correct behavior?  If it's out of bounds, something is really wrong.  Thoughts @mustafabar and @gilbertlee-amd?

alexander-sannikov: That's correct behavior - message size below 64B (2^6) and above 4GB (2^(26+6)) are valid but not covered by corresponding table entries and should just fall into first/last bucket. I assume we have no big interest in messages below 64 bytes to maintain tuning with several-byte granularity, and there is no algorithm switch point beyond 4Gb threshold.

corey-derochie-amd:
Hi @alexander-sannikov . Please use squash merge when merging to rccl in the future.

PR #2095, opened 2025-12-08
Navi4 LL enablement and tuning

PJAvinash:
## Details
Navi4 Perfromance tuning 
- Single and Multinode

**Work item:** "Internal"

**What were the changes?**  
- Minor improvements to Tuning table isolation for different architecture
- Single node LL/Simple protocol selection range update
- LL enablement for gfx1201

- Tuning protocol selection for Single node, as the default choice is no optimal

Experimentation with Single and two node system with 8x GPUs each

**Additional Documentation:**  
- The PR is applicable only if HIP/HSA uncached memory allocation changes are available
- The existing tuning table for gfx1201 may need additional tuning as we get cluster with more nodes


src/graph/tuning.cc
```diff
@@ -758,8 +809,8 @@ ncclResult_t ncclTopoTuneModel(struct ncclComm* comm, int minCompCap, int maxCom
   }
 
   for (int c=0; c<NCCL_NUM_FUNCTIONS; c++) for (int a=0; a<NCCL_NUM_ALGORITHMS; a++) for (int p=0; p<NCCL_NUM_PROTOCOLS; p++) {
-    // Disable LL protocol on gfx12xx
-    int pEnable = (p == NCCL_PROTO_LL && IsArchMatch(comm->topo->nodes[GPU].nodes[0].gpu.gcn, "gfx12")) ? 0 : protoEnable[c*NCCL_NUM_PROTOCOLS+p];
+    // Re-enable LL protocol on gfx12xx
```
corey-derochie-amd:
```suggestion
```
The comment only makes sense in the context of the code change, so it can be removed.

src/graph/tuning.cc
```diff
@@ -908,3 +959,22 @@ ncclResult_t ncclTopoGetAlgoTime(struct ncclComm* comm, int coll, int algorithm,
...
+ * takes gfx arch name as C-style string and returns a tuning index to
+ */
+int rcclGetTuningIndexForArch(const char* gfxarch) {
+    static const std::unordered_map<std::string, int> tuningIndexMap = {
+        {"gfx906", 0}, {"gfx908", 0}, {"gfx90a", 0}, {"gfx942", 5}, 
+        {"gfx950", 6}, {"gfx1030", 0}, {"gfx1100", 0}, {"gfx1102", 0}, 
+        {"gfx1200", 7}, {"gfx1201", 7}
+    };
+    // Lookup requires converting the const char* to a std::string 
+    // (potentially creating a temporary object and allocating memory)
```
corey-derochie-amd:
```suggestion
```
This code won't be called on the hot path, and I expect std::string should use the small string optimization anyway. This probably doesn't need to be noted.

corey-derochie-amd (approved):
lgtm
@wenkaidu @alex-breslow-amd do you think we need the extra INFO logging? I guess more communication is often better, but just want your opinions.

wenkaidu (approved):

PR #2097, opened 2025-12-09
[ci] Bumping TheRock CI commit hash

geomin12:
Bumping TheRock CI commit hash

### DIFF
--- a/.github/workflows/therock-ci-linux.yml
+++ b/.github/workflows/therock-ci-linux.yml
@@ -34,7 +34,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           repository: "ROCm/TheRock"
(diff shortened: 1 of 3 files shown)


araravik-psd (approved):

geomin12:
Landing as this only affects TheRock CI and single-node tests pass. We need to land this in order to add more capacity for mi325 (and decommission the old namespace)

.github/workflows/therock-ci-linux.yml
```diff
@@ -117,7 +117,7 @@ jobs:
         run: |
           python3 build_tools/github_actions/post_build_upload.py \
             --run-id ${{ github.run_id }} \
-            --amdgpu-family ${{ env.AMDGPU_FAMILIES }} \
+            --artifact-group ${{ env.AMDGPU_FAMILIES }} \
```
atulkulk, line 120: @geomin12 This does not look correct to me. Why `--artifact-group` is assigned `env.AMDGPU_FAMILIES`?

PR #2098, opened 2025-12-10
[TheRock CI] Correct runner name for mi325

geomin12:
Correct runner name for mi325


araravik-psd (approved):
Was there a change in the runner name?

geomin12:
Landing as the old runner name will be discontinued and don't want to cause delays

PR #2099, opened 2025-12-10
GDA support for alltoall via rocshmem integration

nusislam:
## Details
Experimental enablement of GDA-based alltoall in RCCL via rocshmem integration. Tested on MI300 with Thor2 NICs.

**Work item:** AICOMRCCL-332 & AIROCSHMEM-198

**What were the changes?**  
Developed an alltoall kernel in RCCL that invokes rocshmem operations for GDA support.

GDA can significantly improve small message alltoall performance.

rocshmem has been integrated with RCCL for GDA support. 

**Additional Documentation:**  
This is currently an experimental feature, enabled by using `--rocshmem` flag during build with the `install.sh` script. The runtime environment variable to use GDA for alltoall is `RCCL_ROCSHMEM_ENABLE=1`.
## Approval Checklist


src/device/generate.py
```diff
@@ -14,6 +14,11 @@
 all_pipelines = ["0", "1"]
 all_unrolls   = ["1", "2", "4"]
 
+# Pipelining is not supported for LL/LL64 prims, so "1" is not a valid value for low latency protocols.
+# However, if it needs to be supported, equivalent_primary() can be modified to avoid the "non-zero"->"0" mapping.
+all_pipeline = ["0", "1"]
+pipelined_types = ["bf16"]
```
atulkulk: Is `all_pipeline` just a placeholder for now? I do not see it being used. OR is it a dup of line 14?
`pipelined_types` is also defined again below on line 162. Duplicate?

src/device/alltoall_gda.h
```diff
@@ -0,0 +1,39 @@
+/*************************************************************************
```
atulkulk, line 1: What is the diff between `alltoall_gda` and `alltoall_gda1` impl?

alex-breslow-amd: I had the same question.

nusislam: `alltoall_gda1` is a kernel for large messages where we use multiple CUs for the data copies.

src/init.cc
```diff
@@ -56,6 +56,12 @@
 #include "rccl_common.h"
 // [/RCCL]
 
+#ifdef ENABLE_ROCSHMEM
+#include <rocshmem/rocshmem.hpp>
+#define NUM_SYM_BUF 8
```
alex-breslow-amd, line 61: This seems potentially dangerous.  Can we get this from ROCSHMEM?

src/init.cc
```diff
@@ -2094,6 +2106,60 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
 
   // RCCL: determine and set unroll factor for comm
   NCCLCHECK(commSetUnrollFactor(comm));
+  //comm->isA2a = 0;
+
+#ifdef ENABLE_ROCSHMEM
+  /* --- sanity-check print statement for development purposes --- */
+  if (rcclParamRocshmemEnabled()) { // @TODO - This doesn't seem to disable when I set ROCSHMEM_ENABLE=0 on command line
+    printf("Initializing rocSHMEM inside of RCCL\n");
```
alex-breslow-amd: INFO debug print instead?

src/init.cc
nusislam: This is a rccl defined number to control the number of symmetric buffers used in rccl.

src/init.cc
```diff
@@ -2094,6 +2106,60 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
...
+      abort();
+    }
+   
+   
+    comm->sourceRshmem = (void**) malloc(NUM_SYM_BUF * sizeof(void *));
+    comm->destRshmem = (void**) malloc(NUM_SYM_BUF * sizeof(void *));
+ 
+    for (int i = 0; i < NUM_SYM_BUF; i++) { 
+    	comm->sourceRshmem[i] = (void *)rocshmem::rocshmem_malloc((size_t)(1*1024*1024));
+    	comm->destRshmem[i] = (void *)rocshmem::rocshmem_malloc((size_t)(1*1024*1024));
```
alex-breslow-amd, line 2157: This constant looks like it should be pulled out and given a name.  What are the implications of this sizing?

.gitmodules
```diff
@@ -8,3 +8,7 @@
 	url = https://github.com/nlohmann/json.git
 	ignore = dirty
 	shallow = true
+[submodule "ext-src/rocSHMEM"]
```
alex-breslow-amd, line 11: To be honest, I don't hate this.

src/init.cc
nusislam: This is a good point. This size should be `>= threshold`. I will change it.

src/device/generate.py
```diff
@@ -79,14 +84,16 @@ def paste(sep, *args):
 if func_pattern and func_pattern[0]:
   func_pattern = func_pattern[0]
 else:
-  func_pattern = "AllGather|AllReduce|AllToAllPivot|Broadcast|Reduce|ReduceScatter|SendRecv"
+  func_pattern = "AllGather|AllReduce|AllToAllPivot|AllToAllGda|AllToAllGda1|Broadcast|Reduce|ReduceScatter|SendRecv"
```
mustafabar: are these new functions built if rocshmem is disabled? I see the guards in the source code, but will this take care of avoiding the build?

alex-breslow-amd:
Can you comment on how this affects compilation time for gfx942 and gfx950?

cmake/Findrocshmem_static.cmake
```diff
@@ -0,0 +1,40 @@
...
+    NAMES rocshmem
+        HINTS ${ROCSHMEM_INSTALL_DIR}/lib)
+  
+## -- todo --- what to do with verbs? add to handle args call below? -- ##
+find_library(IBVERBS ibverbs)
+
+find_package_handle_standard_args(rocshmem_static DEFAULT_MSG ROCSHMEM_INCLUDE_DIR ROCSHMEM_LIBRARY)
+## mark_as_advanced(MSCCLPP_INCLUDE_DIRS MSCCLPP_NCCL_STATIC_LIB) add this for Rocshmem?
+
+## --- TODO --- Remove this, just use for testing purposes -- ###
```
mustafabar: It is a HARD fatal error. Do you want to still keep it?

src/enqueue.cc
```diff
@@ -29,13 +30,19 @@
...
+#endif
+
 using namespace rccl;
 
 struct ncclKernelMatch {
   void* kernelFn;
   bool specialized;
 };
 
+static int symId = 0;
```
mustafabar: Looks like a race. It is shared across all communicators, and can err in multi-threaded mode.

src/init.cc
```diff
@@ -2923,6 +2989,18 @@ ncclResult_t ncclCommDestroy_impl(ncclComm_t comm) {
   }
 #endif
 
+#ifdef ENABLE_ROCSHMEM
+  if (comm->enableRocshmem) {
+     for (int i = 0; i < NUM_SYM_BUF; i++) {	  
+     	rocshmem::rocshmem_free(comm->sourceRshmem[i]);
+     	rocshmem::rocshmem_free(comm->destRshmem[i]);	  
+     }
+     rocshmem::rocshmem_finalize();
```
mustafabar: Is `rocshmem::rocshmem_finalize();` safe when multiple comms call finalize?

src/init.cc
mustafabar: what happens if threshold is set to greater than 1MB? do you clamp it?

.gitmodules
alex-breslow-amd: I guess the question I had is about how we are planning on rolling out rocSHMEM changes.  Are we planning on using a pinned version of rocSHMEM or running part of RCCL CI tests as part of rocSHMEM's release?  I think we need something like that for safety if we're planning on enabling this code path by default as part of our release.

nusislam: We plan to use a pinned version of rocSHMEM.

thomas-huber:
> Can you comment on how this affects compilation time for gfx942 and gfx950?

Hey @alex-breslow-amd! 

On gfx942, a traditional build (`./install.sh -l`) takes `603.045s` and the "with rocshmem" build takes `605.864s` (`./install.sh -l --rocshmem`).

alex-breslow-amd:
> > Can you comment on how this affects compilation time for gfx942 and gfx950?
> 
> Hey @alex-breslow-amd!
> 
> On gfx942, a traditional build (`./install.sh -l`) takes `603.045s` and the "with rocshmem" build takes `605.864s` (`./install.sh -l --rocshmem`).

Great, that is good to hear @thomas-huber and @nusislam!

alex-breslow-amd:
Sorry for closing by accident.

src/device/generate.py
nusislam: Currently, it is always built. The build time is not increased much as @thomas-huber indicated.

nusislam:
I have addressed the review comments.

cmake/Findrocshmem_static.cmake
thomas-huber: removed this, thanks for catching!

src/enqueue.cc
```diff
@@ -1770,7 +1793,9 @@ ncclResult_t ncclLaunchKernel(struct ncclComm* comm, struct ncclKernelPlan* plan
   if (planner->numStreams == 1 && !plan->persistent) {
     latency_profiler::collTraceRecordStartEvent(comm, launchStream, event.get());
     comm->lastStream = planner->streams->stream;
+
```
alex-breslow-amd: Remove additional blank lines

wenkaidu:
Some general comments:
I saw some workloads actually use alltoallv. Is it possible to simply passing down all send/recv requests directly to rocshmem so alltoallv can be supported?
What are minimal required CU from rocshmem? If it will be much smaller, we should target minimal CU usage as well.

nusislam:
> Some general comments: I saw some workloads actually use alltoallv. Is it possible to simply passing down all send/recv requests directly to rocshmem so alltoallv can be supported? What are minimal required CU from rocshmem? If it will be much smaller, we should target minimal CU usage as well.

We have plans to implement an `alltoallv` using rocshmem `put()` operation. The current PR is for `alltoall`, `alltoallv` should come in the next one.

The CU requirement for rocshmem alltoall is 1. For message sizes <= GDA threshold, we are using a single CU.

wenkaidu:
> > Some general comments: I saw some workloads actually use alltoallv. Is it possible to simply passing down all send/recv requests directly to rocshmem so alltoallv can be supported? What are minimal required CU from rocshmem? If it will be much smaller, we should target minimal CU usage as well.
> 
> We have plans to implement an `alltoallv` using rocshmem `put()` operation. The current PR is for `alltoall`, `alltoallv` should come in the next one.
> 
> The CU requirement for rocshmem alltoall is 1. For message sizes <= GDA threshold, we are using a single CU.

where is the line setting CU to 1? I wasn't able to find it

nusislam:
> > > Some general comments: I saw some workloads actually use alltoallv. Is it possible to simply passing down all send/recv requests directly to rocshmem so alltoallv can be supported? What are minimal required CU from rocshmem? If it will be much smaller, we should target minimal CU usage as well.
> > 
> > 
> > We have plans to implement an `alltoallv` using rocshmem `put()` operation. The current PR is for `alltoall`, `alltoallv` should come in the next one.
> > The CU requirement for rocshmem alltoall is 1. For message sizes <= GDA threshold, we are using a single CU.
> 
> where is the line setting CU to 1? I wasn't able to find it

For `<= 256KB`, rccl uses a single CU for alltoall kernel launch.

wenkaidu:
> > > > Some general comments: I saw some workloads actually use alltoallv. Is it possible to simply passing down all send/recv requests directly to rocshmem so alltoallv can be supported? What are minimal required CU from rocshmem? If it will be much smaller, we should target minimal CU usage as well.
> > > 
> > > 
> > > We have plans to implement an `alltoallv` using rocshmem `put()` operation. The current PR is for `alltoall`, `alltoallv` should come in the next one.
> > > The CU requirement for rocshmem alltoall is 1. For message sizes <= GDA threshold, we are using a single CU.
> > 
> > 
> > where is the line setting CU to 1? I wasn't able to find it
> 
> For `<= 256KB`, rccl uses a single CU for alltoall kernel launch.

got it. thanks

wenkaidu (approved):

src/enqueue.cc
nusislam: Done, @alex-breslow-amd

thomas-huber (approved):

src/init.cc
```diff
@@ -2095,6 +2108,61 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
 
   // RCCL: determine and set unroll factor for comm
   NCCLCHECK(commSetUnrollFactor(comm));
+  //comm->isA2a = 0;
```
atulkulk: Is this required?

src/init.cc
```diff
@@ -2095,6 +2108,61 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
...
+    INFO(NCCL_INIT,"Initializing rocSHMEM inside of RCCL");
+    int ret;
+    rocshmem::rocshmem_uniqueid_t rocshmemUniqueId;
+    rocshmem::rocshmem_init_attr_t rocshmemAttr;
+
+    if(comm->rank == 0 ) {
+      ret = rocshmem::rocshmem_get_uniqueid (&rocshmemUniqueId);
+      if (ret != rocshmem::ROCSHMEM_SUCCESS) {
+        WARN("Error in rocshmem_get_uniqueid, Aborting.");
+        abort();
```
atulkulk: Why are we using `abort()` here? Should we be returning instead?

thomas-huber: Hey @atulkulk, I believe we are using abort() here because we want a hard failure if rocshmem cannot perform it's basic setup tasks. If the user has gotten to this block, they have specifically requested rocshmem at build time. If user requests rocshmem support at build and runtime and can't get it due to rocshmem installation/usage issues, we want that to be apparent. I could see users being confused if we fallback to RCCL silently because they do not have logging on. 

@nusislam

thomas-huber: Although, it seems the MSCCL branch does a nice job of WARN'ing and cleaning up, instead of just aborting.

nusislam: Rocshmem is not enabled by default. Let's keep abort for now.

atulkulk: We should be returning `ncclSystemError` as we have done between lines 150 - 156 as an example. RCCL should not abort. It should be handled at the application level and not at the library level.

nusislam: @atulkulk - can you check if the latest commit addresses your concern?

atulkulk: Thanks @nusislam! It does. Small nit: I would suggest to use `ERROR()` instead of `WARN()`.

src/device/generate.py
atulkulk: > Is `all_pipeline` just a placeholder for now? I do not see it being used. OR is it a dup of line 14? `pipelined_types` is also defined again below on line 162. Duplicate?

Pls check on this.

nusislam: Done @atulkulk

atulkulk (approved):

PR #2103, opened 2025-12-11
Direct Reduce Scatter Implementation 

KawtharShafie:
## Details

**What were the changes?**  
New implementation of a direct send/recv reduce scatter. 

**Additional Documentation:**  

## Approval Checklist


src/enqueue.cc
```diff
@@ -390,6 +390,13 @@ ncclResult_t ncclTasksRegAndEnqueue(struct ncclComm* comm) {
     devWork.rcclUseOneSlice = comm->rcclUseOneSlice;
     //[Added-comment] opCount is missing for collDevWork, adding here
     devWork.opCount = task->opCount;
+    // Direct Reduce Scatter
+    if (comm->enableDirectReduceScatter) {
```
nusislam: The message size threshold should be checked here as well.

src/device/reduce_scatter.h
```diff
@@ -19,7 +19,8 @@ namespace {
     // Direct Reduce Scatter
     // TODO: Move Direct RS out of rungRing into separate kernel 
     // 2097152 = 2MB the Direct reduce scatter limit
-    if (work->enableDirectReduceScatter && (work->count * sizeof(T) * ncclShmem.comm.nRanks) <= 2097152) { 
+    size_t msgSize = work->count * sizeof(T) * ncclShmem.comm.nRanks;
+    if (work->enableDirectReduceScatter && msgSize <= 2097152) {
```
nusislam: Need to add a check to enable direct RS for <=8 nodes.

src/collectives.cc
```diff
@@ -17,6 +17,8 @@
 #include <rocshmem/rocshmem.hpp>
 #endif
 
+#define TEMP_BUFF_SIZE (4 * 1024 * 1024) // Define Size for Temporary Buffer for Direct RS
```
KawtharShafie: This line can be removed. TEMP_BUFF_SIZE is defined and used in init but no longer necessary here. The value here is not being used.

src/include/comm.h
```diff
@@ -740,6 +740,12 @@ struct ncclComm {
   int symId;
 #endif
 
+  // Direct Reduce Scatter [RCCL]
+  bool enableDirectReduceScatter;
+  // Temporary Buffer [RCCL]
+  // TODO: Limit tempbuff creation to Direct RS alg
```
KawtharShafie: This TODO comment can be removed

src/rccl_wrap.cc
```diff
@@ -450,6 +483,10 @@ void rcclSetPxn(struct ncclComm* comm,  int& rcclPxnDisable) {
   }
   rcclPxnDisable = pxnDisable;
   comm->enableCustColl = !pxnDisable;
+
+  // Enable Direct Reduce Scatter
+  // TODO: Add condition and Limit for when direct RS will be used
```
KawtharShafie: This TODO line can be removed

src/rccl_wrap.cc
```diff
@@ -450,6 +483,10 @@ void rcclSetPxn(struct ncclComm* comm,  int& rcclPxnDisable) {
...
+  comm->enableDirectReduceScatter = 0;
```
KawtharShafie: This line can probably be removed or taken up into the rcclUseDirectReduceScatter function you worked on. I am resetting this in collectives.cc before the algorithm is entered and then setting it to 1 inside where it would then be propagated to the 'work' item in the kernel for the check there.

src/include/rccl_common.h
```diff
@@ -112,12 +113,16 @@ NCCL_API(ncclResult_t, rcclGetAlgoInfo, struct ncclComm* comm, ncclFunc_t coll,
...
+bool rcclUseReduceScatterDirect(struct ncclComm* comm, size_t& msgSize);
 bool rcclUseAllToAllGda(struct ncclComm* comm);
 void rcclSetPxn(struct ncclComm* comm,  int& rcclPxnDisable);
 void rcclSetP2pNetChunkSize(struct ncclComm* comm,  int& rcclP2pNetChunkSize);
 ncclResult_t rcclFuncMaxSendRecvCount(ncclFunc_t func, int nRanks, size_t count, size_t& maxCount);
 ncclResult_t commSetUnrollFactor(struct ncclComm* comm);
 bool validHsaScratchEnvSetting(const char*hsaScratchEnv, int hipRuntimeVersion, int firmwareVersion, const char* archName);
+
+// Direct ReduceScatter Limit
+RCCL_PARAM_DECLARE(DirectReduceScatterLimit);
```
KawtharShafie: This can be changed from Limit to Threshold to match the syntax used elsewhere (the correct terminology would be limit but seems like threshold is used more often and the norm here). Some places we are referring to it as a limit but then most places it is being called a threshold.

src/rccl_wrap.cc
```diff
@@ -34,6 +34,7 @@ RCCL_PARAM(PipelineAllDTypes, "PIPELINE_ALL_DATA_TYPES", 0);
 // Otherwise, it is automatically set for certain archs, datatypes and reduction collectives
 RCCL_PARAM(disableReduceCopyPipelining, "DISABLE_REDUCE_COPY_PIPELINING", 0);
 RCCL_PARAM(DirectAllGatherThreshold, "DIRECT_ALLGATHER_THRESHOLD", 75497472);
+RCCL_PARAM(DirectReduceScatterThreshold, "DIRECT_REDUCE_SCATTER_THRESHOLD", 2097152);
```
KawtharShafie, line 44: If this parameter is editable by the user, then we need to make sure we add an option and a check for TEMP_BUFF_SIZE also. This would make sure that TEMP_BUFF_SIZE is always equal to or larger than this value. If not, then this should at least be clear in a WARN that the user value is ignored and the DIRECT_REDUCE_SCATTER_THRESHOLD won't be changed beyond 2MB.

src/device/reduce_scatter_direct.h
```diff
@@ -0,0 +1,53 @@
...
+      // Sync threads to ensure all srcPtrs are set before reduction
+      __syncthreads();
+
+      T* recvbuff = (T*)work->recvbuff;
+      // Array for destination pointer to recvbuff
+      void* dstPtrs[1];
+      dstPtrs[0] = (void*)(recvbuff + channelOffset);
+      if (tid < nThreads) {
+        // Call reduction across all rank offsets in tempbuff and store in recvbuff
+        // TODO: Adjust maxSrcs to nRanks
```
KawtharShafie: This TODO can be removed and a separate comment can be added saying that maxSrsc set to 64

src/rccl_wrap.cc
AbandiGa: That’s right, but in rcclUseDirectReduceScatter and in enqueue, I make sure the threshold is set to 2MB or less.

src/collectives.cc
```diff
@@ -429,7 +434,45 @@ ncclResult_t ncclReduceScatter_impl(const void* sendbuff, void* recvbuff, size_t
...
+    // Use Direct Reduce Scatter Algorithm
+    comm->enableDirectReduceScatter = 1;
+    
+    if (recvcount == 0) return ncclSuccess;
+    
+    // Calculate offset into buffers
+    size_t offset = recvcount * ncclTypeSize(datatype);
+    
+    // Copy Current ranks data to tempbuff
+    NCCLCHECK(ncclCudaMemcpy((char*)tempbuff + comm->rank * offset, (char*)sendbuff + comm->rank * offset, recvcount * ncclTypeSize(datatype)));
```
nusislam, line 428: This copy can be done inside the kernel or we can use send/recv. On MI300, hipMemcpy() is slower than CUs for small sizes. We need to do some evaluation on MI350.

src/rccl_wrap.cc
nusislam: In case the user tries to set a threshold higher than 2MB, we should add a message in the debug log saying the direct RS threshold would be up to 2MB.

src/enqueue.cc
```diff
@@ -749,10 +767,12 @@ static ncclResult_t scheduleCollTasksToPlan(
         proxyOp.incWorkCounter = true;
         addWorkBatchToPlan(comm, plan, c, workNode->workType, task->devFuncId, plan->workBytes);
         // Set pattern to profiler to add a proxy profiler for kernel events
-        if (task->func != ncclFuncAllToAllGda) {
+        // for Direct Reduce Scatter (DRS), we don't need to add proxy op
+        bool isDRS = (task->func == ncclFuncReduceScatterDirect && comm->enableDirectReduceScatter);
```
nusislam: Checking just the `task->func` should be OK. You will use func `ReduceScatterDirect` only when `comm->enableDirectReduceScatter` is true.

src/rccl_wrap.cc
```diff
@@ -431,6 +432,41 @@ bool rcclUseAllGatherDirect(struct ncclComm* comm, size_t& msgSize) {
...
+  size_t threshold = rcclParamDirectReduceScatterThreshold();
+  if (threshold > -1) { 
+    // Set threshold to 2MiB hard limit
+    // NOTE: If the DirectReduceScatterThreshold / hard-limit is increased, ensure TEMP_BUFF_SIZE (init.cc)
+    // is increased accordingly -> TEMP_BUFF_SIZE >= 2 * (max enabled msgSize) for headroom.
+    threshold = std::min(threshold, (size_t)2097152);
+  } else {
+    threshold = 2097152;
+  }
+  INFO(NCCL_INIT, "RCCL DIRECT REDUCE-SCATTER threshold set to: %zu", threshold);
```
nusislam, line 467: I see you have the log about threshold here, This is good.

nusislam:
@AbandiGa - The PR needs a rebase on top of latest develop.

src/collectives.cc
nusislam: Actually, blocking memcpy call cannot be used here, use `hipMemcpyAsync()`.

src/device/reduce_scatter_direct.h
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#include "device.h"
+#include "collectives.h"
+#include "primitives.h"
+
+template<typename T, typename RedOp>
+struct RunWorkColl<ncclFuncReduceScatterDirect, T, RedOp, NCCL_ALGO_RING, NCCL_PROTO_SIMPLE> {
```
nusislam, line 13: Need templated versions for different protocols.

ammallya:
Imported to ROCm/rocm-systems

PR #2104, opened 2025-12-12
Add fix for WarpSpeed auto mode

mustafabar:

**What were the changes?**  
* For WarpSpeed auto mode, adjusted the logic to set `info->useWarpSpeed` to `true` only when conditions are met (e.g., correct function type and message size) and to `false` otherwise

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/rccl_wrap.cc
+++ b/src/rccl_wrap.cc
@@ -484,7 +484,7 @@ void rcclSetWarpSpeedAuto(struct ncclComm* comm, struct ncclTaskColl* info, size
     if(info->func == ncclFuncAllReduce || info->func == ncclFuncAllGather) minBytes = RCCL_WARP_SPEED_MIN_BYTES;
     else if (info->func == ncclFuncReduceScatter) minBytes = RCCL_WARP_SPEED_MIN_BYTES << 2; // ReduceScatter requires higher message size to benefit from WarpSpeed
     if(comm->nNodes == 1) {
-      if(nBytes >= minBytes) {
+      if(nBytes >= minBytes && minBytes > 0) {
         comm->unroll = NCCL_UNROLL_2;
         info->nWarps = 4;
         info->useWarpSpeed = true;
--- a/src/rccl_wrap.cc
+++ b/src/rccl_wrap.cc
@@ -477,23 +477,26 @@ void rcclSetWarpSpeedAuto(struct ncclComm* comm, struct ncclTaskColl* info, size
   if(!comm->topo->warpSpeedEnabled) {
     return;
   }
- info->useWarpSpeed = (info->algorithm == NCCL_ALGO_RING); // Enabled by default for any RING algorithm when platform supports it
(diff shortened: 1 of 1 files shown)


src/rccl_wrap.cc
```diff
@@ -477,23 +477,26 @@ void rcclSetWarpSpeedAuto(struct ncclComm* comm, struct ncclTaskColl* info, size
   if(!comm->topo->warpSpeedEnabled) {
     return;
   }
- info->useWarpSpeed = (info->algorithm == NCCL_ALGO_RING); // Enabled by default for any RING algorithm when platform supports it
-  if(rcclParamWarpSpeedAutoMode() != 0 && IsArchMatch(comm->archName, "gfx950")) { // Auto mode only available for gfx950 currently
+  info->useWarpSpeed = (info->algorithm == NCCL_ALGO_RING); // Enabled by default for any RING algorithm when platform supports it
+  if(rcclParamWarpSpeedAutoMode() != 0) {
```
wenkaidu: if rcclParamWarpSpeedAutoMode() is zero,   info->useWarpSpeed will be always true with ring? There is no else handling

mustafabar: I added a small refactor to clarify logic. Can you check now? 
Previous logic worked as at that point it assumed that the user force-enabled it and topo allowed it.

corey-derochie-amd (approved):
lgtm

src/rccl_wrap.cc
```diff
@@ -474,26 +474,25 @@ void rcclSetWarpSpeedSupportAndFinalCuCount(struct ncclComm* comm, struct ncclKe
...
+    }
     size_t minBytes = 0;
+    commSetUnrollFactor(comm);  // TODO: reset unroll factor per task rather than per comm
     if(info->func == ncclFuncAllReduce || info->func == ncclFuncAllGather) minBytes = RCCL_WARP_SPEED_MIN_BYTES;
     else if (info->func == ncclFuncReduceScatter) minBytes = RCCL_WARP_SPEED_MIN_BYTES << 2; // ReduceScatter requires higher message size to benefit from WarpSpeed
     if(comm->nNodes == 1) {
       if(nBytes >= minBytes && minBytes > 0) {
         comm->unroll = NCCL_UNROLL_2;
         info->nWarps = 4;
+        info->useWarpSpeed = true;
```
corey-derochie-amd, line 490: There is no check of `comm->topo->warpSpeedEnabled` up to this point, is that intentional?

mustafabar: > There is no check of `comm->topo->warpSpeedEnabled` up to this point, is that intentional?

warpSpeedEnabled is essentially true when either force-enable or auto-enable is true (based on its initial value in init.cc), so should be ok without the check at this point. The original intention for it was to use it for topo detection. Topo detection mode was dropped to support lower subscriptions (less than 8 gpus used per MI3xx node).

wenkaidu (approved):

PR #2110, opened 2025-12-17
Virtual device enablement ( Minimal changes )

PJAvinash:
## Details

**Work item:** Internal
**What were the changes?**  
- virtual device enablement through change in function input
- Added a warning if NICs across NUMA domains / PCIe root complexes are merged.

- Customer ask

- minimal code changes

**Additional Documentation:**  
- Using Virtual device feature is changing the default ALGO,and NCHANNEL selection in the execution of collective
- Forcing the execution to use default ALGO and NCHANNEL in virtual device mode doesn't revive the performance

## Approval Checklist

### DIFF
--- a/src/init.cc
+++ b/src/init.cc
@@ -1246,9 +1246,10 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   if (dumpXmlFile) {
     NCCLCHECKGOTO(ncclTopoGetSystem(comm, NULL, dumpXmlFile), ret, fail);
   }
-
   // Topo detection / System graph creation
   NCCLCHECKGOTO(ncclTopoGetSystem(comm, &comm->topo), ret, fail);
+  comm->topo->tuning = rcclGetTuningIndexForArch(comm->archName);
+  INFO(NCCL_INIT, "Tuning index set to: %d",  comm->topo->tuning);
   // save nRanks to ncclTopoSystem as indicator of multi-node
   comm->topo->nRanks = comm->nRanks;
   // init netGdrLevel
--- a/src/transport/net_ib.cc
+++ b/src/transport/net_ib.cc
@@ -559,6 +559,110 @@ failure:
   return false;
 }
 
+/**
+ * Extract the PCIe root complex (domain:bus) from a Linux sysfs PCI device path.
+ *
+ * This function scans a sysfs PCI device path (e.g.
+ *   "/sys/devices/pci0000:80/0000:80:03.1/0000:81:00.0")
+ * and extracts the first occurrence of the PCI root host bridge identifier
+ * of the form "pciDDDD:BB", where:
+ *   - DDDD is the PCI domain (hex)
+ *   - BB   is the PCI bus number (hex)
+ *
+ * The extracted root is returned as a string formatted as "DDDD:BB".
+ *
+ * This helper is typically used to determine whether multiple NICs or devices
+ * reside under the same PCIe root complex, which is important for validating
+ * NIC merging and avoiding cross-root PCIe traffic.
+ *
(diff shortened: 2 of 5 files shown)


wenkaidu (approved):

src/transport/net_ib.cc
```diff
@@ -561,7 +561,7 @@ static bool ncclMlx5dvDmaBufCapable(ibv_context *context){
 
 ncclResult_t ncclIbMakeVDeviceInternal(int* d, ncclNetVDeviceProps_t* props) {
   if (ncclParamIbMergeNics() == 0 && props->ndevs > 1) {
-    INFO(NCCL_NET, "NET/IB : Skipping makeVDevice, NCCL_IB_MERGE_NICS=0");
+    WARN("NET/IB : Skipping makeVDevice, Please set NCCL_IB_MERGE_NICS=1");
```
nusislam, line 668: @PJAvinash - Is it asking to enable merge when num NICs per GPU is > 1?

PJAvinash: props->ndevs holds number of physical NIC devices visible to the process.

mustafabar:
>Using Virtual device feature is changing the default ALGO, and NCHANNEL selection in the execution of collective

@PJAvinash can you add exactly the env variable name that will interfere with algo/proto channel selection. Do you know why this is happening? is it due to changing the topo tuning model because that's all we need to select those unless the index returned from your new function is not accurate or some other bug. Performance can be impacted in a way or another, but algo/protocol/channel selection should not be impacted

src/init.cc
```diff
@@ -1246,9 +1246,10 @@ static ncclResult_t initTransportsRank(struct ncclComm* comm, struct ncclComm* p
   if (dumpXmlFile) {
     NCCLCHECKGOTO(ncclTopoGetSystem(comm, NULL, dumpXmlFile), ret, fail);
   }
-
   // Topo detection / System graph creation
   NCCLCHECKGOTO(ncclTopoGetSystem(comm, &comm->topo), ret, fail);
+  comm->topo->tuning = rcclGetTuningIndexForArch(comm->archName);
```
mustafabar, line 1251: can you comment on why you had to introduce this function?

PJAvinash: This only sets the default topo tuning model number ( which would be 0 otherwise). When the NIC merge is selected, and merge happens, the topo matching is false for all the models, and hence comm->topo->tuning remains 0 after initTransportsRank() completion.

mustafabar: If Rome model matching is failing, this will definitely regress, because there's a lot that is done in those models besides just setting the tuning_model. @PJAvinash

mustafabar:
@PJAvinash  - I am going to approve this, but we need to track why Rome model matching is failing, as failure in this detection will degrade performance irrespective of how you set the tuning_model

Please merge if possible as we are including this PR in a drop and need perf traceability.

mustafabar (approved):

PJAvinash:
Unable reproduce performance regression due to this branch. Both RCCL version : 2.27.7-develop:74690ea and RCCL version : 2.27.7-VIRTUAL_DEVICE_NEW:c7fcc0e show similar performance on fp8_e5m2 and fp8_e4m3 at 512K to 4MB message sizes on 8 GPU MI300 system.

mustafabar:
> Unable reproduce performance regression due to this branch. Both RCCL version : 2.27.7-develop:74690ea and RCCL version : 2.27.7-VIRTUAL_DEVICE_NEW:c7fcc0e show similar performance on fp8_e5m2 and fp8_e4m3 at 512K to 4MB message sizes on 8 GPU MI300 system.

I think this can be ignored. We have seen the same issue reported by this CI multiple times with fp8

PR #2125, opened 2026-01-05
WarpSpeed: Improve handling for auto and manual modes

mustafabar:
## Details

**Work item:** _Internal._

**What were the changes?**  

- For WarpSpeed Auto mode:
  -  If algorithm is not ring, avoid changing unroll factor, warp and thread count and disable WarpSpeed.
-  For WarpSpeed Manual mode (enabled either through setting `RCCL_WARP_SPEED_CU_COUNT` and `RCCL_WARP_SPEED_ENABLE`):

   - Force ring algorithm to activate the WarpSpeed path

Achieve more deterministic functionality.

**Additional Documentation:**  

## Approval Checklist

### DIFF
--- a/src/rccl_wrap.cc
+++ b/src/rccl_wrap.cc
@@ -533,7 +533,7 @@ void rcclSetWarpSpeedAuto(struct ncclComm* comm, struct ncclTaskColl* info, size
     // to allow unroll factor to be reverted to default.
     // This can be changed once per-task unroll factor setting is implemented.
     if(info->algorithm != NCCL_ALGO_RING) {
-      return;
+      return; // If Ring is not selected, assume it is suboptimal and return
     }
     if(info->func == ncclFuncAllReduce || info->func == ncclFuncAllGather) minBytes = RCCL_WARP_SPEED_MIN_BYTES;
     else if (info->func == ncclFuncReduceScatter) minBytes = RCCL_WARP_SPEED_MIN_BYTES << 2; // ReduceScatter requires higher message size to benefit from WarpSpeed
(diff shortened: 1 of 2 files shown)


PedramAlizadeh (approved):

src/rccl_wrap.cc
```diff
@@ -529,7 +544,10 @@ void rcclSetWarpSpeedAuto(struct ncclComm* comm, struct ncclTaskColl* info, size
         info->useWarpSpeed = true;
       }
     }
-  } else if (comm->topo->warpSpeedEnabled && info->algorithm == NCCL_ALGO_RING) {
+  } else if (comm->topo->warpSpeedEnabled) {
+    if(info->algorithm != NCCL_ALGO_RING) {
+      info->algorithm = NCCL_ALGO_RING; // Force Ring when WarpSpeed is enabled in manual mode as it only supports Ring
```
atulkulk, line 550: Should we let the user know by INFO message that the RING algorithm is force enabled in manual mode if it is set with a non-supported one?

atulkulk (approved):
LGTM w/ minor suggestion.

src/rccl_wrap.cc
mustafabar: We could do that. Any subsystem suggestion? I am just worried it is a lot of logs since it is called for each coll call. We can put it in a specialized subsystem like INFO(NCCL_TUNING, )

atulkulk: Yes, putting it in `INFO(NCCL_TUNING, )` makes sense.

PR #2133, opened 2026-01-08
8 bytes mem leak fix

PJAvinash:
## Details

**Work item:** 

**What were the changes?**  
malloc + missing free is addressed
Memory leak because of missing free()

**Additional Documentation:**  
This only fixes memory leaks from RCCL
There are some leaks appearing from amd comgr , which cannot be fixed at RCCL layer.
the free() doesn't update the pointer to NULL, this is a common practice in rccl codebase. We plan to address this in a new PR at once.

## Approval Checklist


src/init.cc
```diff
@@ -2004,7 +2004,7 @@ static ncclResult_t ncclCommInitRankFunc(struct ncclAsyncJob* job_) {
   double sum_timers = 0;
   uint64_t timers[TIMERS_INIT_COUNT] = {0};
   unsigned long long commIdHash;
-  char* archName;
+  static __thread char archName[GCN_ARCH_NAME_LEN];
```
atulkulk: I would suggest to use just `char archName[GCN_ARCH_NAME_LEN];` instead of adding `static __thread` to avoid shared buffer concern if there are multiple comms defined within the same thread. Also, in very rare case, what if the multiple comm are defined with different archs? It will overwrite the same buffer resulting into undefined behavior.

corey-derochie-amd (changes requested):
This solution is not valid. `archName` is assigned to the `comm` structure later in the function, so neither a stack-local nor a static thread-local buffer will work.

The proper solution is to `free` the memory when destroying `comm`. There may be more work required to free on early `return` within the function.

PJAvinash:
> This solution is not valid. `archName` is assigned to the `comm` structure later in the function, so neither a stack-local nor a static thread-local buffer will work.
> 
> The proper solution is to `free` the memory when destroying `comm`. There may be more work required to free on early `return` within the function.

Resolved

src/include/comm.h
```diff
@@ -747,7 +747,7 @@ struct ncclComm {
   int unroll;
   // custom collective [RCCL]
   bool enableCustColl;
-  // gfx name from hipDeviceProp_t [RCCL]
+  // gfx name from hipDeviceProp_t [RCCL] , Memory resource owned by comm allocated in ncclCommInitRankFunc
```
corey-derochie-amd, line 750: good idea

corey-derochie-amd (approved):

atulkulk (approved):

ammallya:
Imported to ROCm/rocm-systems

PR #2137, opened 2026-01-11
fix potential segfaults due to use after malloc fails

prasanna-amd:
Details

What were the changes?
fix potential segfaults due to use after malloc fails

Why were the changes made?
Improves rccl stability, especially in low memory conditions it is not desirable to have segmentation faults, it is better to gracefully handle out of memory conditions

How was the outcome achieved?
check for return value from malloc, and take failure path instead of continue if malloc failed

### DIFF
--- a/src/proxy.cc
+++ b/src/proxy.cc
@@ -96,6 +96,10 @@ static ncclResult_t expectedProxyResponseEnqueue(struct ncclProxyState* state, v
 
   // Pre-alloc response buffer
   ex->respBuff = malloc(respSize);
+  if (ex->respBuff == NULL && respSize > 0) {
+    free(ex);
+    return ncclSystemError;
+  }
   ex->respSize = respSize;
   ex->res      = ncclInternalError;
   ex->done     = false;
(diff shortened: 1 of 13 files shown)


wenkaidu (approved):

ext-profiler/example/plugin.c
```diff
@@ -95,6 +95,7 @@ __hidden ncclResult_t exampleProfilerInit(void** context, int* eActivationMask,
 
   // pre-allocate memory for event object pools in dedicated profiler context
   struct context* ctx = (struct context *)calloc(1, sizeof(*ctx));
+  if (ctx == NULL) return ncclSystemError;
```
alex-breslow-amd: Can you bulk replace the use of NULL with nullptr?

src/proxy.cc
```diff
@@ -94,6 +94,10 @@ static ncclResult_t expectedProxyResponseEnqueue(struct ncclProxyState* state, v
 
   // Pre-alloc response buffer
   ex->respBuff = malloc(respSize);
+  if (ex->respBuff == NULL && respSize > 0) {
+    free(ex);
```
alex-breslow-amd, line 100: Is there any risk of a double free here with code changes?

alex-breslow-amd (approved):
LGTM

ext-profiler/example/plugin.c
prasanna-amd: fixed.

PR #2139, opened 2026-01-12
fix bug in reduce kernel bfloat16 for ROCm >= 6.0

prasanna-amd:
As part of an earlier commit, bfloat16 handling in reduce kernel for FuncMinMax fell into generic/default template when there is no SPECIALIZE_REDUCE for a particular type, this generic template does a bitwise integer comparison and it broke bfloat16 ops. 
change the else-if statement to else statement, that way it covers both ROCm version < 6.0 and >= 6.0 (with ROCm > 6.0, device.h already typedefs __hip_bfloat16 to hip_bfloat16, so no special case is needed here).


src/device/reduce_kernel.h
```diff
@@ -414,7 +414,7 @@ SPECIALIZE_REDUCE(FuncMinMax, half, 1, half, fn.isMinNotMax ? __hmin(x, y) : __h
   SPECIALIZE_REDUCE(FuncMinMax, __nv_bfloat16, 1, __nv_bfloat16, fn.isMinNotMax ? __hmin(x, y) : __hmax(x, y))
   // coverity[copy_constructor_call]
   SPECIALIZE_REDUCE(FuncMinMax, __nv_bfloat16, 2, __nv_bfloat162, fn.isMinNotMax ? __hmin2(x, y) : __hmax2(x, y))
-#elif ROCM_VERSION < 60000
```
corey-derochie-amd, line 417: @PedramAlizadeh @mberenjk Do we have a sense of why the original guard was there? This would have disabled the reduce specialization on all ROCm versions after 6.0, right? Is there a ROCm version where the BF16 implementation changes?

corey-derochie-amd (approved):

wenkaidu (approved):

src/device/reduce_kernel.h
prasanna-amd: When the specialize function is missing for a template in c++ it falls back to the generic implementation:
that means for FuncMinMax, the implementation it uses is:

template<typename T>
struct Apply_Reduce<FuncMinMax<T>, /*EltPerPack=*/1> {
__device__ __forceinline__ static BytePack<sizeof(T)> reduce(FuncMinMax<T> fn, ByteP
ack<sizeof(T)> a, BytePack<sizeof(T)> b) {
return (a.native ^ fn.xormask.native) < (b.native ^ fn.xormask.native) ? a : b;
}
};

this generic implementation compares the raw bytes as unsigned integers, it was designed
for integer types where xormask flips sign bits to get correct signed comparisons.

generic FuncSum/FuncProd implementations in reduce_kernel.h are:

template<typename T>
struct Apply_Reduce<FuncSum<T>, /*EltPerPack=*/1> {
__device__ __forceinline__ static BytePack<sizeof(T)> reduce(...) {
return toPack<T>(fromPack<T>(a) + fromPack<T>(b)); // Uses operator+
}
};

template<typename T>
struct Apply_Reduce<FuncProd<T>, /*EltPerPack=*/1> {
__device__ __forceinline__ static BytePack<sizeof(T)> reduce(...) {
return toPack<T>(fromPack<T>(a) * fromPack<T>(b)); // Uses operator*
}
};

generics use the actual + and * operators on the type T, and they're implemented for these types
Sum/Prod still work because hip_bfloat16 and __hip_bfloat16 has operator+ and operator* defined, so the
generic implementations produce correct results

but, MinMax fails because the generic does a raw bitwise unsigned integer comparison on a .native,
which doesn't give correct min/max for floating-point bit patterns (negative floats have high bit set)

minmax was broken since the earlier commit, will add a unit-test in another PR to catch this issue in the future

PedramAlizadeh (approved):

PR #2143, opened 2026-01-13
[CUMEM] Initial support for cuMem APIs

BertanDogancay:
## Details

AICOMRCCL-110

**What were the changes?**  
Add support for when NCCL_CUMEM_ENABLE=1

A lot of the new features are relying on cumem support and this has been enabled in NCCL by default.

**Additional Documentation:**  
This feature is disabled by default until issues regarding performance are resolved. Currently we only mark it as functional not performant.

Some of the other work for the enablement of these APIs were made in 2.27 sync. 

## Approval Checklist


BertanDogancay:
@mustafabar can you plz review this PR for warpspeed changes? After #2073 there were segfaults seen when cumem memory was being used so I came up with a fix. Let me know if you are okay with these changes.

src/include/alloc.h
corey-derochie-amd, line 318: This is not in NCCL, do we know why this is necessary?

src/include/alloc.h
```diff
@@ -303,7 +303,8 @@ static inline ncclResult_t ncclCuMemFreeAddr(void *ptr) {
   if (ptr == NULL) return ncclSuccess;
   ncclResult_t result = ncclSuccess;
   size_t size = 0;
-  CUCHECK(cuMemGetAddressRange(NULL, &size, (CUdeviceptr)ptr));
+  void* base = nullptr;
+  CUCHECK(cuMemGetAddressRange(&base, &size, (CUdeviceptr)ptr));
```
corey-derochie-amd: Curious why this is needed?

src/include/alloc.h
```diff
@@ -367,11 +368,12 @@ static inline ncclResult_t ncclCuMemFree(void *ptr) {
...
   CUCHECK(cuMemRetainAllocationHandle(&handle, ptr));
-  CUCHECK(cuMemRelease(handle));
-  CUCHECK(cuMemGetAddressRange(NULL, &size, (CUdeviceptr)ptr));
+  // CUCHECK(cuMemRelease(handle));
+  void* base = nullptr;
+  CUCHECK(cuMemGetAddressRange(&base, &size, (CUdeviceptr)ptr));
   TRACE(NCCL_ALLOC, "CuMem Free Size %zu pointer %p handle 0x%llx", size, ptr, handle);
   CUCHECK(cuMemUnmap((CUdeviceptr)ptr, size));
-  CUCHECK(cuMemRelease(handle));
+  // CUCHECK(cuMemRelease(handle));
```
corey-derochie-amd, line 376: NCCL makes these release calls, why are they removed?

src/include/alloc.h
```diff
@@ -442,17 +444,23 @@ ncclResult_t ncclCudaCallocDebug(const char *filefunc, int line, T** ptr, size_t
...
-    CUDACHECK(cudaStreamCreateWithFlags(&stream, cudaStreamNonBlocking));
-  CUDACHECKGOTO(hipExtMallocWithFlags((void**)ptr, nelem*ncclSizeOfT<T>(), flags), result, finish);
-  CUDACHECKGOTO(cudaMemsetAsync(*ptr, 0, nelem*ncclSizeOfT<T>(), stream), result, finish);
-  CUDACHECKGOTO(cudaStreamSynchronize(stream), result, finish);
-  if (sidestream == nullptr)
-    CUDACHECKGOTO(cudaStreamDestroy(stream), result, finish);
+  if (nelem > 0) {
+    // Need a side stream so as not to interfere with graph capture.
+    cudaStream_t stream, sidestream;
+    NCCLCHECK(getSideStream(&sidestream));
```
corey-derochie-amd, line 450: Why is `getSideStream` added here?

src/include/alloc.h
BertanDogancay: I remember this was added by Nusrat some time ago to track memory, so yes this is not from NCCL.

src/include/alloc.h
BertanDogancay: `getSideStream()` was already there, we now call it after checking for `num elements > 0`

src/include/alloc.h
BertanDogancay: HIP does not accept NULL for this API and it segfaults. This is a workaround.

src/rccl_wrap.cc
```diff
@@ -507,12 +507,11 @@ void rcclSetWarpSpeedCUs(struct ncclComm* comm, int algo, int threadsPerBlock, i
...
 }
 
-void rcclSetWarpSpeedSupportAndFinalCuCount(struct ncclComm* comm, struct ncclKernelPlan* plan, int nChannels, int& support, int &cuCount) {
-  if(!comm->topo->warpSpeedEnabled) {
-    support = 0;
-    cuCount = nChannels;
-    return;
+bool rcclWarpSpeedSupported(struct ncclComm* comm, struct ncclKernelPlan* plan) {
+  if (!comm->topo->warpSpeedEnabled) {
+    return 0;
```
mustafabar: `return false` for sanity sake :)

src/include/alloc.h
```diff
@@ -367,11 +368,12 @@ static inline ncclResult_t ncclCuMemFree(void *ptr) {
   CUmemGenericAllocationHandle handle;
   size_t size = 0;
   CUCHECK(cuMemRetainAllocationHandle(&handle, ptr));
-  CUCHECK(cuMemRelease(handle));
-  CUCHECK(cuMemGetAddressRange(NULL, &size, (CUdeviceptr)ptr));
+  // CUCHECK(cuMemRelease(handle));
+  void* base = nullptr;
```
mustafabar: let's use `CUDevicePtr` for sanity and determinism. 
`void*` will probably warn or err.

src/include/alloc.h
```diff
@@ -367,11 +368,12 @@ static inline ncclResult_t ncclCuMemFree(void *ptr) {
...
+  CUCHECK(cuMemGetAddressRange(&base, &size, (CUdeviceptr)ptr));
   TRACE(NCCL_ALLOC, "CuMem Free Size %zu pointer %p handle 0x%llx", size, ptr, handle);
   CUCHECK(cuMemUnmap((CUdeviceptr)ptr, size));
```
mustafabar, line 375: This is suspicious to me. What will happen if `base!=ptr`, do we take into account memory leaks?

src/include/alloc.h
```diff
@@ -367,11 +368,12 @@ static inline ncclResult_t ncclCuMemFree(void *ptr) {
   CUmemGenericAllocationHandle handle;
   size_t size = 0;
   CUCHECK(cuMemRetainAllocationHandle(&handle, ptr));
-  CUCHECK(cuMemRelease(handle));
-  CUCHECK(cuMemGetAddressRange(NULL, &size, (CUdeviceptr)ptr));
+  // CUCHECK(cuMemRelease(handle));
```
mustafabar, line 371: why comment this out?

src/include/transport.h
```diff
@@ -40,13 +40,9 @@ struct ncclComm;
 #define CHANNEL_MASK_OFFSET(nranks, connIndex) (nranks * (connIndex == NCCL_CONN_IDX_P2P_NET ? NCCL_CONN_IDX_P2P_NET : 0))
 
 #define CONNECT_SIZE 256
-#if defined(__HIP_PLATFORM_AMD__) || defined(__HIPCC__)
```
mustafabar, line 43: I will trust there's a good reason for removing AMD specific page size limits.

src/rccl_wrap.cc
```diff
@@ -526,9 +525,7 @@ void rcclSetWarpSpeedSupportAndFinalCuCount(struct ncclComm* comm, struct ncclKe
     }
     task = task->next;
   }
-  int warpsPerBlock = plan->threadPerBlock / comm->WarpSize;
-  support = (hasP2p || hasNonRing) ? 0 : 1;
-  cuCount = (support == 0)? nChannels : nChannels / warpsPerBlock + ((nChannels % warpsPerBlock) != 0 ? 1 : 0); // each CU can handle warpsPerBlock
+  return (hasP2p || hasNonRing) ? 0 : 1;
```
mustafabar:
```suggestion
  return (!hasP2p && !hasNonRing);
```

src/include/transport.h
BertanDogancay: I added this in 2.27 sync. At the time, HIP was only returning the minimum page size (4K), so we needed to adjust these numbers. This is no longer needed since it's now fixed by the runtime and we're matching CUDA.

src/include/alloc.h
wenkaidu: We need to file JIRA to runtime if these release calls caused issues. These APIs need to match CUDA behavior.

BertanDogancay: @corey-derochie-amd @wenkaidu @mustafabar  These calls cause segfaults during freeing the memory. I did some debugging and found out that it only segfaults when freeing `p2pBuff->directPtr` in [#L776](https://github.com/ROCm/rccl/blob/develop/src/transport/p2p.cc#L776) and [#L759](https://github.com/ROCm/rccl/blob/develop/src/transport/p2p.cc#L759) which is not in the VA space. My guess is Cuda is more forgiving with this behavior, but HIP is more strict.

src/include/alloc.h
BertanDogancay: Discussion in above comment

src/include/alloc.h
wenkaidu: Current workaroud is OK. But please file a JIRA to HIP on this issue then remove workaround later

src/include/alloc.h
wenkaidu: @BertanDogancay From CUDA documentation, cuMemRelease should be releasing the handle of the memory. It didn't say the memory must be still mapped. I think we need to discuss with HIP team

ammallya:
Imported to ROCm/rocm-systems

PR #2144, opened 2026-01-15
[gfx950] Reduce single node all-reduce's default CU usage

mustafabar:
## Details

**Work item:** _AICOMRCCL-361_

**What were the changes?**  
- Reduce all-reduce CU usage to 56 CUs by default on a single node
- Modify `RCCL_WARP_SPEED_ENABLE` to `RCCL_WARP_SPEED_FORCE_ENABLE` for debugging
- Set chunksteps to 1 for single node all-reduce which performs better with unroll factor of 2 needed by WarpSpeed
- Return `RING*` to rccl-tests to highlight warp-specialized ring usage
- Ensure semantic equivalence to NCCL_MAX_NCHANNELS to reflect CU count (not logical channels) when WarpSpeed is enabled or force-enabled. 

- End-to-end workload performance is impacted when compute is overlapped with communication. `all_reduce` is heavily used in such workloads
- WarpSpeed performance could not be observed out-of-the-box for a single node all-reduce
- Implementation's peak BW for all-reduce can now be realized without special environment variables using 56 CUs (down from 112)

- Changes to channel upscaling and downscaling logic
- For single node, use single slice and chunk for all-reduce to get an uplift for message sizes that do not use WarpSpeed with lower CU count of 56. WarpSpeed does require unroll factor of 2, so it is applied to all message sizes.

**Additional Documentation:**  

## Approval Checklist


src/collectives.cc
```diff
@@ -158,9 +158,15 @@ ncclResult_t ncclAllReduce_impl(const void* sendbuff, void* recvbuff, size_t cou
     NVTX3_PAYLOAD(comm ? comm->commHash : 0, count * ncclTypeSize(datatype), op, datatype));
 
   // RCCL update slice steps for AllReduce if single node
+  const bool isGfx950 = IsArchMatch(comm->archName, "gfx950");
+  int chunkSteps = (isGfx950 && comm->rcclUseOneSlice)? 1 : ALLREDUCE_CHUNKSTEPS;
```
alex-breslow-amd: What does updating the chunk steps do here?  I think slice steps is enough.  I'm confused here.  We're already using a single slice as is in the code for gfx950 unless this got deleted somehow, and I missed it.

mustafabar: We are currently setting [`StepPerSlice`](https://github.com/search?q=repo%3AROCm%2Frccl%20StepPerSlice&type=code) to [`NCCL_STEPS/2` for all-reduce ](https://github.com/ROCm/rccl/blob/239d62f54565657055d9659fc017586fc2ba5c51/src/include/collectives.h#L24) . This is not the same as setting it to 1.

alex-breslow-amd: This is true, but there is only one slice per chunk for single node workloads.  Are you seeing a performance uplift from fewer slice steps on single node?  I see that AllToAll sets slice and chunk steps both to 1.  When I did initial testing for single node, I don't recall the steps per slice having an appreciable performance impact, but let me know if I missed something.

mustafabar: This to recover the performance decrease from having to set the unroll factor to 2 for WarpSpeed to get competitive performance at 56 CUs. 
When keeping previous single node's unroll factor of 1 for MI350, the previous chunk and slice steps were good enough as is. The change of chunk steps is to recover 3%-9% of perf for message sizes that don't use WarpSpeed so that we don't have to introduce unroll factor per-message size tunable. I will share some performance data in the chat later.

nileshnegi (approved):

src/include/device.h
```diff
@@ -137,7 +137,7 @@ union ncclLLFifoLine {
 #endif
 
 #ifdef ENABLE_WARP_SPEED
-#define MAXCHANNELS 512
+#define MAXCHANNELS 256
```
alex-breslow-amd, line 140: A question on this.  Are you dropping the max channels back to 256?  Ah, never mind.  This makes sense.

alex-breslow-amd (approved):
Thank you for making this the default.

src/graph/connect.cc
```diff
@@ -830,10 +838,37 @@ ncclResult_t ncclTopoPostset(struct ncclComm* comm, int* firstRanks, int* treePa
...
+  if(wsEnabled) {
+    // If user didn't override, use requested channels; otherwise keep capped max.
+    if (!userUpdatedMaxChannels) {
+      maxNchannels = nc * comm->nChannels * channelMultiplier;
+      nc = singleNode? maxNchannels : std::min(maxNchannels, maxChannels);
+    } else {
+      nc = maxNchannels = std::min(adjustedMaxNchannels * channelMultiplier, MAXCHANNELS);
+    }
+
+    if (!userUpdatedMaxChannels && isGfx950 && singleNode && wsEnabled && comm->nRanks == 8) {
```
corey-derochie-amd:
```suggestion
    if (!userUpdatedMaxChannels && isGfx950 && singleNode && comm->nRanks == 8) {
```

mustafabar: Done

ammallya:
Imported to ROCm/rocm-systems

PR #2148, opened 2026-01-20
Stream Pytest Summary from Slurm Jobs into GitHub Actions Logs

araravik-psd:
PR Description

This PR improves the GitHub Actions output for Slurm-based CVS test runs by surfacing only the relevant pytest results instead of the full Slurm logs.

**Changes:**

- Log surfacing and readability
- Capture and print the Slurm job ID returned by sbatch --wait.
- Extract and print pytest output starting from “short test summary info” to the end of the Slurm stdout log.
- Add a fallback to tail the last lines of the log when the pytest summary is not present.
- Always print the Slurm job ID and paths to the full stdout/stderr log files for easy debugging.

**Correct CI failure semantics:**

- Query the final Slurm job state after completion using sacct (with scontrol fallback).
- Treat jobs that are CANCELLED, FAILED, PREEMPTED, TIMEOUT, or never started as CI failures.
- Allow CI to pass only when the Slurm job reaches the COMPLETED state, ensuring the test actually ran.
- Prevent false-positive CI passes when jobs are cancelled before execution and no logs are produced.

**Impact**

- Keeps CI logs concise and readable.
- Preserves full Slurm logs for offline analysis.
- Makes it easier to correlate GitHub Actions runs with Slurm jobs.
- Passing run printing the relevant logs:

https://github.com/ROCm/rccl/actions/runs/21156987948/job/60846454749#step:4:29

### DIFF
--- a/.github/workflows/therock-test-packages-multi-node.yml
+++ b/.github/workflows/therock-test-packages-multi-node.yml
@@ -69,6 +69,8 @@ jobs:
           JOB_ID=$(SETUP_NODES=1 sbatch --wait --parsable -N4 \
             /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch)
 
+          echo "Slurm job ID: ${JOB_ID}"
+
           LOG_FILE="/apps/cvs_tests/logs/TheRock-Multinode-RCCL-CI-Test-${JOB_ID}.out"
 
           if grep -q "short test summary info" "$LOG_FILE"; then
@@ -77,6 +79,7 @@ jobs:
             echo "[WARNING] Pytest summary not found; showing results tailing logs"
             tail -n 22 "$LOG_FILE"
           fi
+          echo "Full logs can be found at $LOG_FILE in the runner"
 
 
       - name: Configure AWS Credentials for non-forked repos
--- a/.github/workflows/therock-test-packages-multi-node.yml
+++ b/.github/workflows/therock-test-packages-multi-node.yml
@@ -66,7 +66,18 @@ jobs:
       - name: Test gfx950
         if: ${{ inputs.amdgpu_families == 'gfx950-dcgpu' }}
         run: |
-          SETUP_NODES=1 sbatch --wait -N4 /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch
+          JOB_ID=$(SETUP_NODES=1 sbatch --wait --parsable -N4 \
+            /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch)
+
+          LOG_FILE="/apps/cvs_tests/logs/TheRock-Multinode-RCCL-CI-Test-${JOB_ID}.out"
+
+          if grep -q "short test summary info" "$LOG_FILE"; then
+            sed -n '/short test summary info/,$p' "$LOG_FILE"
+          else
+            echo "[WARNING] Pytest summary not found; showing results tailing logs"
+            tail -n 22 "$LOG_FILE"
+          fi
+
 
       - name: Configure AWS Credentials for non-forked repos
         if: ${{ always() && !github.event.pull_request.head.repo.fork }}
--- a/.github/workflows/therock-test-packages-multi-node.yml
+++ b/.github/workflows/therock-test-packages-multi-node.yml
@@ -66,20 +66,8 @@ jobs:
       - name: Test gfx950
         if: ${{ inputs.amdgpu_families == 'gfx950-dcgpu' }}
         run: |
-          JOB_ID=$(SETUP_NODES=1 sbatch --wait --parsable -N4 \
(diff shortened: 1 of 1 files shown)


.github/workflows/therock-test-packages-multi-node.yml
```diff
@@ -66,7 +66,21 @@ jobs:
       - name: Test gfx950
         if: ${{ inputs.amdgpu_families == 'gfx950-dcgpu' }}
         run: |
-          SETUP_NODES=1 sbatch --wait -N4 /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch
+          JOB_ID=$(SETUP_NODES=1 sbatch --wait --parsable -N4 \
```
speriaswamy-amd: I `scanceled` the job while it was still running to simulate a job failure, these were the results : 
```
JOB_ID=$(SETUP_NODES=1 sbatch --wait --parsable -N4 \
    /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch)
  
  echo "Slurm job ID: ${JOB_ID}"
  
  LOG_FILE="/apps/cvs_tests/logs/TheRock-Multinode-RCCL-CI-Test-${JOB_ID}.out"
  
  if grep -q "short test summary info" "$LOG_FILE"; then
    sed -n '/short test summary info/,$p' "$LOG_FILE"
  else
  Warning: echo "[WARNING] Pytest summary not found; showing results tailing logs"
    tail -n 22 "$LOG_FILE"
  fi
  echo "Full logs can be found at $LOG_FILE in the runner"
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    VENV_DIR: /actions-runner/_work/rccl/rccl/.venv
    ARTIFACT_RUN_ID: 21177683813
    OUTPUT_ARTIFACTS_DIR: /apps/cvs_tests/dist_new/dist/rocm
    THEROCK_BIN_DIR: ./build/bin
    AWS_SHARED_CREDENTIALS_FILE: /apps/cvs_tests/awsconfig/credentials.ini
    pythonLocation: /actions-runner/_work/_tool/Python/3.12.12/x64
    PKG_CONFIG_PATH: /actions-runner/_work/_tool/Python/3.12.12/x64/lib/pkgconfig
    Python_ROOT_DIR: /actions-runner/_work/_tool/Python/3.12.12/x64
    Python2_ROOT_DIR: /actions-runner/_work/_tool/Python/3.12.12/x64
    Python3_ROOT_DIR: /actions-runner/_work/_tool/Python/3.12.12/x64
    LD_LIBRARY_PATH: /actions-runner/_work/_tool/Python/3.12.12/x64/lib
    VIRTUAL_ENV: /actions-runner/_work/rccl/rccl/.venv 
```   
In the failure case, `sbatch --wait` returns non‑zero when the job is canceled, and because your step is running with bash -e -o pipefail, the script exits immediately at the sbatch line.
That means everything after it (printing JOB_ID, tailing the log) never ran, so GitHub only showed `exit code 1`, which is not really useful for us to debug

speriaswamy-amd: I feel like we should do something like: 

```bash
set +e
JOB_ID_RAW=$(SETUP_NODES=1 sbatch --wait --parsable -N4 /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch)
SBATCH_RC=$?
set -e
```
and regardless of the exit code of the sbatch script we tail the out and err file if it exits
something like `if [[ -f $OUT_FILE  ]]; then tail it ` similarly for err fail, then finaly we should still return the error code of the sbatch script since the job failed and the CI should fail as well; `exit "$SBATCH_RC"`

.github/workflows/therock-test-packages-multi-node.yml
```diff
@@ -66,7 +66,21 @@ jobs:
...
+            /apps/cvs_tests/cvs-sbatch/sbatch/default.sbatch)
```
speriaswamy-amd: We should ideally throw all of this into a bash script - `ci-entrpoint.sh` and call it from the CI, to reduce the amount of inline bash in this yaml file

speriaswamy-amd (changes requested):
Ideally we should be able to go through the debug logs even when the sbatch submission fails

araravik-psd:
Thanks @speriaswamy-amd, trying out changes for failed cases of sbatch submission. Will revert once I have it running

araravik-psd:
Logs for the below Cases:

Logs when the multinode script Passes:

https://github.com/ROCm/rccl/actions/runs/21190422648/job/60970374328#step:4:74

Logs when the multinode script is cancelled with Scancel:

https://github.com/ROCm/rccl/actions/runs/21190422648/job/61028269238#step:4:133

ammallya:
Imported to ROCm/rocm-systems

PR #2159, opened 2026-02-05
Don't use `--offload-compress` at compile time

shiltian:
Since `-fgpu-rdc` is enabled, there is no benefit of doing compression at compile time, because the compressed device "object file" (an IR file in binary format) will be decompressed immediately at link time, so the compression at compile time is completely a waste of time and resource.


shiltian:
* #2159 👈
* `develop`

<h2></h2>

This stack of pull requests is managed by <a href="https://github.com/shiltian/sgh"><b>sgh</b></a>.

shiltian:
Didn't notice this repo was deprecated.

PR #2161, opened 2026-02-11
ROCM-1896 fix shutdown ordering race condition

prasanna-amd:
## Details
race condition at shutdown causes segmentation fault, instead have an atexit handler to set global shutdown state and avoid use after free bugs.

ROCM-1896

**What were the changes?**  
Added atexit handler to set shutdown flag and prevent other threads from running into segmentation faults

Without these changes, an error condition during rccl (like no open files, or running out of hip memory), could result in a segmentation fault, instead of gracefully exiting with an error reason.

Handle exit conditions and prevent other threads from running into segfault

**Additional Documentation:**  
cherry pick from develop

### DIFF
--- a/src/include/alloc.h
+++ b/src/include/alloc.h
@@ -18,11 +18,32 @@
 #include <stdlib.h>
 #include <string.h>
 #include "rccl_vars.h"
+#include <atomic>
+#include <mutex>
 
 #if CUDART_VERSION >= 11030
 #include <cuda.h>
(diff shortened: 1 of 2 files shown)


corey-derochie-amd (approved):

JeniferC99:
manual psdb triggered: http://rocm-ci.amd.com/job/compute-psdb-rel-7.2/346

PR #2176, opened 2026-04-27
Add NCCL_PAT_PRECONNECT to gate eager PAT preconnections

t2h-fb:
## Details
[PAT](https://developer.nvidia.com/blog/new-scaling-algorithm-and-initialization-with-nvidia-collective-communications-library-2-23/#pat_logarithmic_scaling_for_reducescatter_and_allgather%C2%A0) (parallel aggregated trees) algorithm is added in NCCL, which requires much more QPs (queue pairs) than ring or tree algorithms. PAT is never activated in RCCL but their preconnections are established regardless. It might be an issue to some NICs with tight QP resource constraints. Adding an environment variable to disable PAT preconnections.

## Approval Checklist


wenkaidu (approved):

alex-breslow-amd (approved):
LGTM

